Issue #185, opened 2014-02-19
Plugin Architecture


dboskovic:
@JedWatson @bladey @JossMackison

I'm absolutely loving working with Keystone. I'd be very interested in helping to develop out a plugin API and directory. This would open up Keystone to user-contributed admin functionality.

I noticed a transition to modularizing the different field's UI files... Is this something you're planning? If so, are you interested in my contribution to help expedite it? If not, let's discuss it here further to decide on a pattern and I can work on it and then issue a PR.

JohnnyEstilles:
Continuing discussion on issue #912 ...

PR #218, opened 2014-03-10
WIP: Allow custom field types to be added, ref #185

thedersen:
Hi,
I just noticed this project, and I think it looks great and have lots of potential. However, I need it to be more pluggable for me to be able to use it, so I thought I might be able to help with that.

This is a rather large pull request, and I understand if you need some time to review it, but here is the gist of it:

* All field types are moved into a new folder, plugins/fields/[fieldType]/index.js
* All templates are moved into a new folder, plugins/fields/[fieldType]/templates/form.jade
* The code and templates are in the same directory to make it easier to distribute plugins
* I have added an `extend` method to field.js to make it easier to create custom fields (see example below)
* All field types are loaded dynamically from plugins folder inside keystone
* If you want to add a custom field type, put it inside your plugins/fields/[fieldType] directory of your application, and it will be available to use
* Everything should be backwards compatible

This is still work in progress, but I thought I would get your input before spending too much time on it.

What's remaining?

* Loading plugins from npm. This should be trivial at this stage, but we need to agree on some conventions on how to package and naming the plugins etc.
* We need to look at how processFilters and getSearchFilters are done (this is currently inside list.js, but should be moved into the field types)
* Make the plugin directory configurable
* Testing... (I have not tested all the field types after this change)
* Probably something more

Example: 

``` js
// Field and keystone are injected into the
// factory method, so that you do not need to add a
// reference to keystone when publishing your custom
// field types on for instance npm.
module.exports = function(FieldBase, keystone) {
  return FieldBase.extend({
    constructor: function(list, path, options) {
      // Do something
      // Then call base
      FieldBase.apply(this, arguments);
    },

    someMethod: function(item, data) {
    }
  });
};
```
Hope you appreciate this, and that we can discuss further development on this issue.

Thomas


jamlen:
Any progress with this? I've got a couple of custom fields I want to start looking at adding.

metasansana:
@thedersen are you still on this?

JedWatson:
@jamlen @metasansana this has been superseded by our work with React, custom field types will be supported shortly after `0.3.0` is released.

@thedersen please let us know if you're still interested in working on this, I'll add you to our Slack project!

thedersen:
@JedWatson Sorry, but atm I don't have the time to contribute.

kadosh1000:
Hi I have published my npm package: 
https://github.com/kadosh1000/keystone-custom-fieldtypes

It copies your custom fieldTypes from a selected dir into the keystone's fields dir and registers it

PR #230, opened 2014-03-12
New Azure Storage Blob File field.

jbalde:
Azure Storage is a cloud storage like S3 in Azure Cloud. This field is based in S3 field with Azure client.

Configure with

```
keystone.set('azurefile config', { account: 'storage account' /* not user account */, key: 'secret api key' });
```

Or with environment

```
process.env.AZURE_STORAGE_ACCOUNT and process.env.AZURE_STORAGE_ACCESS_KEY
```

The default container (like bucket in S3 is a keystone).

Model only need defined the type

```
simpleModelProperty: { type: Types.AzureFile }
```

but cant defiend filenameFormatter and containerFormatter to format the name of the file and the container name is are diferent of default 'keystone' container.

```
simpleModelProperty: { type: Types.AzureFile }
//the next property defined the upload filename as itemID with extension format a name like 5315f3a519e382ee32b80472.pdf
//with container formatter can define a custom container name with item and filename. For example can create a custom container with only images or with a property of model.
customModelProperty: { type: Types.AzureFile, filenameFormatter: function(item, filename) { return item._id + require('path').extname(filename); }, containerFormatter: function(item, filename) { return (item.cdnName != "") ? item.cdnName : 'mycdn'; } },
```

If configure the variables with test values the resultant urls are like:

```
keystone.set('azurefile config', { account: 'myaccount' /* not user account */, key: 'secret api key' });

//TEST SIMPLE: http://myaccount.blob.core.windows.net/keystone/my-file-name.pdf
//TEST CUSTOM: http://myaccount.blob.core.windows.net/mycdn/5315f3a519e382ee32b80472.pdf
```


lib/fieldTypes/azurefile.js
```diff
@@ -0,0 +1,358 @@
...
+			'See http://keystonejs.com/docs/configuration#azurefile for more information.\n');
+	}
+
+	process.env.AZURE_STORAGE_ACCOUNT = this.azurefileconfig.account;
+	process.env.AZURE_STORAGE_ACCESS_KEY = this.azurefileconfig.key;
+
+	self.azurefileconfig.container = self.azurefileconfig.container || 'keystone';
+
+	options.filenameFormatter = options.filenameFormatter || function(item, filename) { return filename };
+	options.containerFormatter = options.containerFormatter || function(item, filename) { return self.azurefileconfig.container };
```
spiritinlife: options.filenameFormatter and options.containerFormatter are never assigned to self.
Am i missing something ?

PR #264, opened 2014-03-27
Feature 137 - mongoose history

britztopher:
Added ability to use mongoose history (https://github.com/rossan/mongoose-history).  There is an option named saveHistory that defaults to false meaning no history will be saved on the list.  There are a few concerns i have and they are as follows:

1. I added a conditional statement in list.js within the constructor to check if option saveHistory === true then use mongoose-history plugin for that list, however, there probably is a more elegant way to accomplish this same task, but I could not find where you enforce options on the list.

I also updated the docs with the new option and information needed to use the history option

### DIFF
--- a/docs/content/pages/docs/database.jade
+++ b/docs/content/pages/docs/database.jade
@@ -127,7 +127,7 @@ block content
 					td Prevents deletion of items from the list through the Keystone Admin UI
                 tr
                     td <code>saveHistory</code> <code class="data-type">Boolean</code>
-                    td Determines whether or not the item's state is history is saved. ie - saves item's history for updates, removes, and inserts into mongo.
+                    td Determines whether or not the item's state history is saved. ie - saves item's history for updates, removes, and inserts into mongo.
 
 			
 			a(name='lists-plugins')
--- a/docs/content/pages/docs/database.jade
+++ b/docs/content/pages/docs/database.jade
@@ -125,6 +125,10 @@ block content
 				tr
 					td <code>nodelete</code> <code class="data-type">Boolean</code>
 					td Prevents deletion of items from the list through the Keystone Admin UI
+                tr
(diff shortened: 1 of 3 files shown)


tivac:
Given the age of this I doubt it'd merge any more, but I'd really, *really* love to see this functionality in keystone.

JedWatson:
Sorry all for leaving this behind. I wrote up quite a large response when it was current but it doesn't look like I posted it :(

I'm keen to get this feature but I'd prefer to put the functionality directly into Keystone (rather than pulling in mongoose-history) so we can have greater control over the way it's integrated.

Specifically, information about version history should be surfaced with the meta about the item (i.e. tracking info) and available in the Admin UI; we can also do a good job of integrating it with our current tracking functionality. I think this will be easier if we implement it ourselves based on mongoose-history.

I'd also like to limit our implementation to storing _just the difference_ when a document is saved; we can use `isModified` on each field path to retrieve the changed fields (I'm thinking whole fields, not modified paths) and store that. We can also store an array of the fields changed for each history item. As well as resulting in smaller history sizes, this will let us easily show / report on changes in the UI.

The reason I didn't merge this as-is was that it would create breaking changes when we moved to our own implementation, which can be really tricky with data migrations to take into account; and it's possible to integrate mongoose-history with Keystone in a project (without requiring it to be built into the core).

JedWatson:
@britztopher @tivac just landed our own version of this. Let me know what you think. see https://github.com/keystonejs/keystone/blob/master/lib/schemaPlugins/history.js for implementation details.

kadosh1000:
@JedWatson  Is there a way in the Admin UI to go back to earlier version?

VinayaSathyanarayana:
What is the status of this? Is this available in the latest beta?

Issue #127, opened 2014-01-17
WYSIWYG Editor - default toolbar over-ride


mbayfield:
The current wysiwyg editor implements only a subset of the toolbar features that TinyMCE supports. It would be good to be able to set an over-ride in the keystone config in web.js for tool bar options to be supported.

PR #285, opened 2014-04-10
Add support for adding additional toolbar buttons to tinymce

theskabeater:
Issue #127

Example (in web.js):
keystone.set('wysiwyg additional buttons', 'styleselect, blockquote');

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,12 +17,12 @@ var templateCache = {};
 
 /**
  * Keystone Class
- * 
+ *
  * @api public
  */
 
 var Keystone = function() {
-	
+
 	this.lists = {};
 	this.paths = {};
 	this._options = {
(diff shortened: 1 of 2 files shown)


dallasbpeters:
anywhere to get a list of what can be added to the editor?

theskabeater:
It’s been awhile but I believe Keystone uses TinyMCE as it’s WYSIWYG.

http://www.tinymce.com/wiki.php/TinyMCE3x:Buttons/controls <http://www.tinymce.com/wiki.php/TinyMCE3x:Buttons/controls>

Those controls should work. Let me know.

- Eric

dallasbpeters:
Yep! and I found this late last night as well: http://keystonejs.com/docs/configuration/#options-ui which explains a bit more as well.

PR #299, opened 2014-04-24
Added descriptions to list items and updated docs.

RichardOtvos:
Hey, 

I had a usecase where I wanted to give some general info about the model in its item view, and I found the field notes a bit clunky for that. So I added the option to give a model a description and it will show up in the Admin list item view of the model.
I also updated your docs with the description of this (see database page).

Some screens of what I'm talking about:
without description (same as now):
without_desc

with description:
with_desc

EDIT: I'm really not sure why you can't just automerge this, should be just something like 4-6 new lines altogether

### DIFF
--- a/docs/content/pages/docs/database.jade
+++ b/docs/content/pages/docs/database.jade
@@ -7,47 +7,47 @@ block intro
 		.container
 			h1 Database
 			p.lead Setting up and using Data Models
-	
+
 block content
 	.container: .row
-		
+
 		.col-sm-3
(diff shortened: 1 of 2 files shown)


creynders:
:+1: for the idea, but it needs a full rewrite I think.

mxstbr:
This is over two years old and hasn't been updated, so I'm closing it – thanks very much though for working on it, and I'll definitely merge an implementation of this on `master`!

Issue #153, opened 2014-02-03
Sub-collections


azproduction:
Hello!

I have complex model with nested collections. I'm trying to define such model

``` js
{
    title: { type: String, required: true },
    images: {
        type: [{
            name: String,
            image: Types.Url,
            thumb: Types.Url,
            description: String
        }]
    }
}
```

And I get an error `Error: Fields must be specified with a type function` @ `keystone/lib/list.js:298:9`.

Is it possible to define that collection in keystone? (it is possible in Mongoose)

In admin UI this field should look like list of items with Add, Edit and "Trash" buttons. Add and Edit buttons should open modal dialog.

I am new to keystone, but I want to contribute to this project. And it will be great if you guide me a bit.

wangpingsx:
Really expect to have this feature!!!!!

And what i want is a Dynamic array field, which like tags for a blog post.

I have a temporary solution, I would like to share:

Design.add( {
	design: { type: String , initial: true, required: true, index: true },
	priceGroup: { type: String },
	price: { type: String},
	image: { type: Types.CloudinaryImage },
	},
	"Colours", { 
		1: { 
			type: String,collapse:true
		},
		2: { 
			type: String,collapse:true
		},
		3: { 
			type: String,collapse:true
		},
		4: { 
			type: String,collapse:true
		},
		5: { 
			type: String,collapse:true
		},
		6: { 
			type: String,collapse:true
		},
		7: { 
			type: String,collapse:true
		},
		8: { 
			type: String,collapse:true
		},
		9: { 
			type: String,collapse:true
		},
		10: { 
			type: String,collapse:true
		},
		11: { 
			type: String,collapse:true
		},
		12: { 
			type: String,collapse:true
		},
		13: { 
			type: String,collapse:true
		},
		14: { 
			type: String,collapse:true
		},
		15: { 
			type: String,collapse:true
		},
		16: { 
			type: String,collapse:true
		}
});

I know I am terrible smart!!!!!   :-)

All colours fields are collapsed when you open the create view of admin UI. 

if you need only one colour open the first colour field.

All looks great!!, But it will always have 16 sub items in mongodb like this:

{
    "_id" : ObjectId("ddddadfasdfasdfasdfasdfasdf"),
    "design" : "aaa",
    "__v" : 0,
    "1" : "red",
    "2" : "",
    "3" : "",
    "4" : "",
    "5" : "",
    "6" : "",
    "7" : "",
    "8" : "",
    "9" : "",
[...]

PR #301, opened 2014-04-26
Array field

subhog:
This is a very basic solution for issue #153. It allows to create array fields. Only objects can be stored in the array (no array of strings, sorry). API and drawbacks are described in the issue.


terenzeyuen:
Is this safe to use? Need something like this badly.

subhog:
I don't have anything in common with Keystone since May 2014, but if you're asking this question here it probably means nothing like this is there. This particular pull request was never merged nor safe to use.

Issue #291, opened 2014-04-14
Use named routes to generate links


subhog:
Let's get a sample route:

    app.get('/blog/article/:title/:id', routes.views.blog.article);

To create a link to that route inside a template, I need to manually enter the path:

    a(href='/blog/article/' + article.title + '/' + article.id) Article

I can write my own helper to faciliatate this, but if said helper was to work with all the routes I've got in my app, I would need to copy all the routing logic inside the helper.

What I'd like to have instead is the option to name certain routes and use those names to generate the links automatically. This is the same concept used in Rails or in Iron Router for Meteor.

For example, I could name that example route:

    app.get('/blog/article/:title/:id', routes.views.blog.article, 'blog/article');

And then Keystone could provide a global helper for my templates:

    a(href=pathFor('blog/article', article)) Article

fridays:
I would also like to use named routes. Has this been implemented?

andrewlinfoot:
We're closing all feature requests to keep the issue tracker unpolluted. From now on submit feature requests on [productpains.com](https://productpains.com/product/keystonejs)!

https://productpains.com/post/keystonejs/use-named-routes-to-generate-links

PR #305, opened 2014-04-28
`PathFor` helper #291

subhog:
This allows to name your routes in the same place where they're defined and use these names to generate links.

The API is described in issue #291.

I've tried to make as few changes to core as possible. The API cannot be easily made simpler as the route definitions go straight to Express and so you cannot simply pass the name as an argument to `app.get`.


JedWatson:
Closing because this is too old and hasn't been required since. Thanks for the original PR @subhog

PR #404, opened 2014-06-04
Restish

davidbanham:
I think this is now at the point where it warrants discussion. The point of this PR is to get an out-of-the-box generic list API from Keystone that works with [AngularJS' resource service](https://docs.angularjs.org/api/ngResource/service/$resource) and is generally rest-y.

The main things I've done are:

  1. add a new getAll method to the generic list api
  1. add a new endpoint at /keystone/restish_api

The new getAll method does what it says on the tin. Returns all members of a list. The obvious threat here is that maybe that list is gigantic. If so, I think the onus is on the app developer to add their own API routes. I think the happy path is to have it Just Work out of the box, and once you've got enough data for that to be a problem you can deal with it in whatever way makes sense for your app.

It would have been super-nice to be able to overload the /keystone/api endpoint so that it was compatible with angular, but I couldn't see a way to do it without also intercepting the old-style API calls to that endpoint. In order to not break backwards compatibility, I've namespaced these changes.

I think it's worth asking the question about backwards compatibility with this API, though. To my mind, the closer we get to the common conception of RESTfulness, the better. Even if you're not using Angular, this should work with most things that expect a rails-style REST API. Anyone that's used that kind of API in the past should be able to get up to speed pretty quickly without needing to read the docs.

### DIFF
--- a/routes/api/list.js
+++ b/routes/api/list.js
@@ -102,7 +102,15 @@ exports = module.exports = function(req, res) {
 				}
 			}
 
-			req.list.model.find(null, null, opts).exec(function(err, item) {
+			if (req.query.filter) {
+				var filter = {};
+				req.query.filter.split(',').map(function(param) {
+					var elems = param.split(':');
+					filter[elems[0]] = elems[1];
+				});
+			};
+
+			req.list.model.find(filter || null, null, opts).exec(function(err, item) {
 
 				if(err) return sendError('database error', err);
 				if (!item) return sendResponse({ name: req.query.id, id: req.query.id });
--- a/routes/api/list.js
+++ b/routes/api/list.js
@@ -89,6 +89,27 @@ exports = module.exports = function(req, res) {
 
 
 		break;
+		
+		case 'getAll':
+
+			req.list.model.find().exec(function(err, item) {
+
+				if (err) return sendError('database error', err);
+				if (!item) return sendResponse({ name: req.query.id, id: req.query.id });
+
+				switch (req.query.dataset) {
+					case 'simple':
+						return sendResponse({
+							name: req.list.getDocumentName(item, true),
+							id: item.id
+						});
+					break;
+					default:
+						return sendResponse(item);
+				}
+			});
+
+		break;
 
 		case 'get':
 
--- a/routes/api/list.js
+++ b/routes/api/list.js
@@ -99,7 +99,7 @@ exports = module.exports = function(req, res) {
 				opts = {
 					skip: per_page * (page - 1),
 					limit: per_page * (page - 1)
-				}
+				};
 			}
 
 			if (req.query.filter) {
@@ -108,7 +108,7 @@ exports = module.exports = function(req, res) {
 					var elems = param.split(':');
 					filter[elems[0]] = elems[1];
 				});
-			};
+			}
 
 			req.list.model.find(filter || null, null, opts).exec(function(err, item) {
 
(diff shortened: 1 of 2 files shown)


cameronjroe:
I agree with @Globegitter that the namespace should be a little more conventional then `/restish_api` and would agree that `/api/v2` could be a good choice. Also, it seems like there's high coupling happening in both the new PR routes added and the switch statement under the current `/api`. Would it be beneficial to create a RouteController class that handles the api routes in a RESTful way along with any other non-RESTful handlers that need to be used in the switch statement - all of which contributing to an `/api/v2`?

morenoh149:
Reading up on this. Could we achieve the desired effect by adding https://github.com/orbitjs/orbit.js to Keystone?

From http://jsonapi.org/implementations/

christroutner:
After reading [Jed's gist](https://gist.github.com/JedWatson/9741171#file-routes-index-js-L24) I opened up the API on my own installation and wrote up a [5 part tutorial on how I did it](http://christroutner.com/blog/post/front-end-widgets-part-1-creating-the-db-model). Now I'm starting to implement Backbone.js to give my app a little more structure. After reading through the mailing list and Issues, I see a lot of talk but nothing that resembles a tutorial or code snippets. Has anyone successfully implemented the Backbone.js REST API in KeystoneJS? If so, please chime in.

I'm going to push forward with this and as usual, I'll write up my findings on [my blog](http://christroutner.com/blog). I don't want to reinvent the wheel though, so please stop me if this problem has been solved.

routes/api/list.js
```diff
@@ -89,6 +89,45 @@ exports = module.exports = function(req, res) {
 
 
 		break;
+		
+		case 'getAll':
```
ericelliott: I think a dedicated router at `/api/v2/lists/` would be better than a `getAll` case here.

You may also want to check out [Chapter 8. Building RESTful APIs from Programming JavaScript Applications](http://chimera.labs.oreilly.com/books/1234000000262/ch08.html) for ideas.

ericelliott:
I definitely think improving our REST endpoints should be a high priority. See my [priority wishlist](https://github.com/keystonejs/keystone/issues/1897).

Related: [restful-keystone](https://github.com/d-pac/restful-keystone)

christroutner:
I got [a thread going on the mailing list](https://groups.google.com/forum/#!topic/keystonejs/juWVUxTJjeo) to continue discussion of this pull request. Ultimately, the effort to implement a standardized RESTful API is too much effort for my needs at the moment, so I have to abandon this. I sure hope someone else picks this up.

For reference though, here is the REST format that Backbone.js (and Angular.js and Ember.js too, I think) expects:

The default sync handler maps CRUD to REST like so:

create → POST   /collection
read → GET   /collection[/id]
update → PUT   /collection/id
patch → PATCH   /collection/id
delete → DELETE   /collection/id

This was taken from [the Backbone.js documentation](http://backbonejs.org/#Sync).

ericelliott:
Here's the restful-keystone mapping:

* list -->  GET /api/<collection>: a list operation; returns all of the resources
* create --> POST /api/<collection>: a create operation; creates a new resource in the collection
* read --> GET /api/<collection>/:id: a retrieve operation; retrieves a single resource from the collection
* patch --> PATCH /api/<collection>/:id: an update operation; updates the state of the resource with the values from the payload
* delete --> DELETE /api/<collection>/:id: a remove operation; removes the resource from the collection

Looks like all it's missing for Backbone / Angular / Ember mapping is an `UPDATE` mapping.

Maybe a good place to start is to extend restful-keystone?

JedWatson:
Closing as the work here has been superseded. An official API for generating REST endpoints will be back on the radar after the 0.4 release. Thanks for the original work @davidbanham and everyone else for the input!

PR #422, opened 2014-06-15
add the option to set file limit

izifortune:
You can set `file limit` parameter in keystone to set the file limit for express


christroutner:
Any documentation on how to use this change? Do I set it in the keystonel.js file? If so, does it look something like `keystone.set('file limit', '50MB')`?

PR #496, opened 2014-07-28
localfiles formating template fixture

a2nt:


JedWatson:
Thanks @a2nt and sorry this got left so long! Closing because we've implemented this differently since it was submitted.

PR #509, opened 2014-07-31
display image as img tag when displayed in a column

jsanchezpando:


JedWatson:
Implemented in 0.4, thanks for the original PR @jsanchezpando

Issue #458, opened 2014-07-09
Access logged in User object in model?


alsoicode:
Is it possible to get access to the currently logged in User object in a model?

I have two relationship fields to 'User' that I need for auditing purposes. I don't want to expose these fields in Keystone admin. I would be super if I could set them as part of a 'pre' or 'post' schema operation on my List.

How can I access the User object from a model?

PR #514, opened 2014-08-06
Adding current user available to pre/post save hooks

estilles:
This PR makes the current user available to `pre` and `post` save hooks.

I do not alter the model per se. I simply add a `_user` property to the `item` object just before it is saved in `updateHandler`, but since `mongoose` is unaware of it's existence, it does not save it to the database. The property is temporary in nature and does not survive the current transaction.

Here's a usage example.

```
List.schema.pre('save', function(next, done) {
    console.log('current user =>', this._user);
    next();
});
```

The idea of having the current user available to `pre` and `post` save hooks was originally raised in #458 in the context of using it for audit purposes (which I addressed in #490). However, having the current user available in `pre` and `post` save hooks is also quite useful in use cases other than this, which is why I decided to submit this as a separate pull request.

### DIFF
--- a/lib/updateHandler.js
+++ b/lib/updateHandler.js
@@ -166,6 +166,9 @@ UpdateHandler.prototype.process = function(data, options, callback) {
 	}.bind(this);
 
 	var saveItem = function() {
+		// Make current user available to pre/post save events
+		this.item._user = this.user;
+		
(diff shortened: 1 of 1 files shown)


alisajidcs:
This feature should be in pre/post remove too

PR #577, opened 2014-08-26
Add hook before and after delete a file

SamyLegal:
Add possibility, for the localfile field, of specify a pre and/or post hook when we delete a file.

When i upload an image on my website with the localfile field, i generate multiple images of differents size. For that, i use post move hook of localfile. I have add a pre/post delete hook to delete images i have generate, when i delete the original file.


morenoh149:
this could be useful

creynders:
Hi @SamyLegal sorry for the _very_ long wait. Just a quick update: I've rewritten the entire hooking mechanism for keystone, which will land in v0.4. This means that your PR would need a number of changes to be up-to-date. We're starting assembly of v0.4 changes today or tomorrow, so I'm going to take this PR along partially and modify it to use the new hooking system.

JedWatson:
Closing in line with @creynders note - this has been superseded. Thanks for the original PR @SamyLegal!

Issue #320, opened 2014-05-05
View item button in admin UI


jossmac:
It'd be nice to be able to view the changes you've just made to an item from the editing interface. I suggest a button right aligned in the toolbar [ View Post ] or whatever your item name.

Perhaps in the model we could store a `previewPath` or something similar that would look like "/blog/post/:key".

Yinear:
Why this is not been merged?

morenoh149:
@Yinear this is not a pull request. If mean #608 then the author said himself that it doesn't solve the most complex usecase. More discussion is needed. @Yinear have your reviewed the pull request? what do you think is missing?

Yinear:
@morenoh149 I need just a simple button like he does.  Or a text label to show the link of the Post.
Maybe I can do it by adding a note to the post title

JedWatson:
Wow, this one is old. Sorry about that. I'll take a look.

JedWatson:
Unfortunately the PR isn't production-ready, this issue among others:

`return this.options.preview.replace(':key', item[this.autokey.path]);`

We need to include proper template parsing so all fields (configurable auto key path, id, etc) are supported.

I'll see if we can get a solution that works in soon based on the work that's been done.

albertogasparin:
I have improved my previews PR with a better, new one. This time the code will handle also more complex scenarios:

```
// simple config and path as a string
var Post = new Keystone.List({
    preview: '/blog/post/:key'
});

// simple config and path as a method
var Post = new Keystone.List({
    preview: function () {
        var date = new Date(this.publishedDate);
	return '/blog/'+ date.getUTCFullYear() +"/"+ this.key
    }
});

// complex config and path as a method
var Post = new Keystone.List({
    preview: {
        label: 'View Post',
        available: function() { return (this.state == 'published'); },
        path: function() { return '/blog/post/' + this.key }
    }
});
```

However, I would rise a question around the naming: `preview` sounds misleading because you are not allowed to view unsaved changes. What about something like `publicPath`?

Moreover, if you want to view items with a "draft" status you have to explicitly add an exception in your front-end views code, otherwise you will get a 404:

```
var q = keystone.list('Post').model.findOne().where('slug', locals.filters.post);

if(!locals.user) { // filter unpublished posts if not logged in
    q.where('state', 'published');
}
...
```

This should be documented somewhere.

JedWatson:
[...]

PR #608, opened 2014-09-03
Add "view item" button in Admin UI if "preview" is defined

albertogasparin:
Solves issue #320, however my code does not cover the most complex usage.


morenoh149:
could you make this production ready?

JedWatson:
Thanks for the PR @albertogasparin, we'll be introducing a similar feature in 0.4 but implemented differently!

Issue #13, opened 2013-08-07
New field type: array


JedWatson:
Add an `array` field type that stores a simple list of `String`, `Number` or `Date` values.

Pylipala:
Can I use array of custom inline object?
For example, I have an order model, which contains many order details. I want the order header and its details to be in the same mongodb document. Is it possible in keystonejs?

QualityWorksCG:
Would love the ability to have an array of Type.Select

morenoh149:
@Pylipala @QualityWorksCG open up a new issue if you think its a good new feature

VinayaSathyanarayana:
Where can one find documentation of TextArray and NumberArray?

morenoh149:
@JedWatson https://github.com/keystonejs/keystone/issues/1021 do you need to cut a new site update?

dsmurl:
Would be nice to have for adding varying number of images to an article, recipe, or other list type item.

That is to say, an array of type Image or something.

PR #656, opened 2014-09-27
Multi-value number & text fields

bardzusny:
True support for "multi: true;" property for Text and Number fields. Issue #13 .

I'm pretty sure it's hackish (to say the least). But seems to work fine, and looks like a pretty fundamental feature. Any suggestions appreciated.

Catches:
- Got an array with 10 000 values? Tough luck, that's how many input fields your browser will end up rendering (trying to render?).
- No fancy drag'n'drop re-ordering of values or anything even remotely similar.


quitequinn:
This is awesome — but not included in the docs. Is there a spot where new...ish... field types exist?

PR #680, opened 2014-10-23
Adding Support for sub directories

jtymann:
I have added two new options that allow for keystone to be better integrated with existing sites. The options are "webroot" and "cmsdir".

Webroot allows you to change the base location of where keystone is hosted. Previously you always needed to host the public files at your web root. This makes for sloppy integration with existing sites. This also makes it so you do not need to use the mount command to achieve this setup. 

Cmsdir allows you to change the name of the keystone directory. This allows you to host your admin panel in a custom directory, which can allow for easier to remember urls or allow for additional security by making the location of the admin panel not easily predictable. 

If you want to host the site your site where "http://www.mysite.com/blog" is the cms frontend and the backend is hosted at "http://www.mysite.com/blog/admin" you would do the following in your init call:

keystone.init({
'name': 'Keystone',
'brand': 'Keystone Co.',

'less': 'public',
'static': 'public',
'favicon': 'public/favicon.ico',
'views': 'templates/views',
'view engine': 'jade',

'emails': 'templates/emails',

'auto update': true,
'session': true,
'auth': true,
'user model': 'User',
'cookie secret': 'lasdjhfalsdkjfhalsdjkfhasldkjfhasldjfahsdlfasjkdhf',

'webroot': '/blog/',
'cmsdir': 'admin',
});

### DIFF
--- a/index.js
+++ b/index.js
@@ -43,7 +43,7 @@ var Keystone = function() {
 		'auto update': false,
 		'model prefix': null,
 		'webroot': '/',
-		'cmsdirectory': 'keystone',
+		'cmsdir': 'keystone',
 	};
 	this._pre = {
 		routes: [],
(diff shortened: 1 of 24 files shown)


JedWatson:
Closing as we've implemented dynamic admin paths in 0.4. Thanks for the original PR @jtymann!

PR #738, opened 2014-11-24
Add option to specify multiple values for dependsOn 

snowkeeper:
Allows you to specify a dependsOn option with an Array to match one of multiple entered values


VinayaSathyanarayana:
Can dependsOn be a function that provides an Array of values?

PR #761, opened 2014-12-04
Adds appversion rendering to footer

creynders:
I almost always add a version number display to my apps for various reasons: people can refer to it when encountering bugs, it's an easy confirmation the application was rebooted after a release, etc.

This PR introduces a new setting `appversion` which is rendered (together with `brand`) in the footer.

Usage examples:

``` js
//file: keystone.js
keystone.init( {
  'name'  : 'd-pac',
  'brand' : 'd-pac',
  'appversion' : require("./package.json").version
} );
```

or

``` js
keystone.set('appversion', "0.5.4");
```

Which will get rendered as (on a retina resolution):

footer

Sidenote: TBH I think it would be better to store an `app` object, sth like:

``` js
'app' : {
  'name'  : 'd-pac',
  'brand' : 'd-pac',
  'version' : require("./package.json").version   
}
```

But I'm a bit hesitant to make the necessary changes, since it's a bit hard to track down all places where `name` is used for instance ( a search will yield 1000+ results).

### DIFF
--- a/lib/core/render.js
+++ b/lib/core/render.js
@@ -50,6 +50,7 @@ function render(req, res, view, ext) {
 		numeral: numeral,
 		env: keystone.get('env'),
 		brand: keystone.get('brand'),
+		appversion : keystone.get('appversion'),
 		nav: keystone.nav,
 		messages: _.any(flashMessages, function(msgs) { return msgs.length; }) ? flashMessages : false,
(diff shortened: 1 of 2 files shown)


DaTebe:
Is this feature removed in 4.0.0?

PR #774, opened 2014-12-14
updated item.jade

brianjd:
seemed a bit redundant to have the objectid referenced twice on the same page


JedWatson:
Moved to React.

PR #788, opened 2014-12-17
Add CA option to SSL configuration

snowkeeper:


justinsullard:
@snowkeeper Unfortunately the current implementation only appears to be setup for a single ca bundle path, but https.createServer option spec (https://nodejs.org/api/https.html#https_https_createserver_options_requestlistener) allows for an array of them. If a bundle file is used (GoDaddy ca_bundle for instance) https only serves up the first cert in the file, not all contained, so the array must be implemented.

JedWatson:
@justinsullard would you mind opening a new Issue for that? it's hard to track updates needed for merged PRs

justinsullard:
@JedWatson Opened a new issue: https://github.com/keystonejs/keystone/issues/1247

PR #848, opened 2015-01-02
Use jsxhint instead of jshint

### LINKED ISSUE #847: Linter doesn't understand JSX
@morenoh149 @sebmck 

New tests implementation looks great.

I've noticed that the linter is now throwing a lot of issues because it doesn't understand JSX though, is there a plugin or alternative package we can use that is JSX-aware?

Alternatively perhaps we should exclude the JSX files until we find a workaround, I think it's better to lint fewer files than a get bunch of false-positives in the report.

sebmck:
Closes #847

Also removes `quotmark` jshint option since the react transform uses double quotes instead of single.

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -14,7 +14,6 @@
   "nonbsp": true,
   "nonew": true,
   "plusplus": false, /* custom */
-  "quotmark": true,
   "undef": true,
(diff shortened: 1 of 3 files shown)


JedWatson:
Fantastic, thanks @sebmck

PR #850, opened 2015-01-03
v0.3.0 add back fixed toolbar

snowkeeper:
I added the previous toolbar code into the react component.  I played around with using bootstrap affix, but the current code works the best.

Are we adding the version number for specific branch commits as well or is selecting the correct branch enough for you Jed?


JedWatson:
@snowkeeper this looks good for now :)

Adding the version number pre- 0.3.0 works, helps add clarity.

Cheers!

PR #852, opened 2015-01-03
add Adding cloudinary secure option to zh docs, fixes #373

### LINKED ISSUE #373: Unclear how to force Cloudinary requests over HTTPS
I configured my NGINX server to use SSL directly without going through the Keystone SSL server. Everything works fine, except that all Cloudinary requests are over HTTP, which is raining on my parade. I don't see any straightforward way to let Keystone know that I am indeed attempting to use SSL and that it should use https:// prefixes for Cloudinary URLs, and I don't see anything in the documentation about this. Please advise?

weekwood:


JedWatson:
Thanks @weekwood!

PR #859, opened 2015-01-05
session store options example

snowkeeper:
Updated the session store options examples to include keystone.set. I blindly copied it earlier and even used the object key as my option name.

Fixed the view path in docs.js so server would start.

@weekwood Can you check my copy paste?

### DIFF
--- a/docs/content/en/pages/docs/configuration.jade
+++ b/docs/content/en/pages/docs/configuration.jade
@@ -253,7 +253,8 @@ block content
 						.code-header
 							h4 Example for connect-mongostore
 						pre: code.language-javascript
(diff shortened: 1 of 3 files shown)


weekwood:
@snowkeeper looks good!

JedWatson:
@snowkeeper looks like you've got changes in here that are for Express 4 from the master branch, but this is now 0.2.x which is still on Express 3.

Can you revert your changes to `docs/docs.js`? Then I can merge it.

JedWatson:
You need to `rm -rf node_modules && npm install` when switching branches or things to badly ;)

PR #867, opened 2015-01-05
Throw error if express app doesn't have necessary middlewares

morenoh149:
ran into this error


JedWatson:
These are good changes, thanks @morenoh149.

I'll try and give you some more clarity on custom Express app setup shortly.

Issue #868, opened 2015-01-05
Shorthand syntax of createItems does not log errors


webteckie:
@JedWatson this has been an issue forever.  I sent a PR once that logged the errors right in createItems and you tweaked the PR and logged the errors upstream in update.js, which makes sense if it would just work :-)  However, that has never worked for me with the shorthand syntax and any mongoose validation or any other errors as those are not propagated up to update.js since for some strange reason that I can't figure out the callback into createItems logs as undefined.  I've looked at it a couple of times and could never come up with a reason as to why it isn't working.  I will just tell you that I added some logs (prefixed with ***) at the following points with results.  Hopefully that will give you some clue as to what may be going on :-)

   update.js:

	// auto-wrap create scripts for a friendlier shorthand syntax
	if (_.isObject(update.create)) {
		var items = update.create,
			ops = update.options || {};
		update = function(done) {
			keystone.createItems(items, ops, function(err, stats) {
				console.log("***Returned from createItems");
				stats && console.log(stats.message);
				done(err);
			});
		};
	}

   createItems.js (L11):
    function createItems(data, ops, callback) {
	
        console.log("***callback at entry: " + JSON.stringify(callback) );
		
	var keystone = this;

   createItems.js (L294):
	], function(err) {

		console.log("***callback at exit: " + JSON.stringify(callback) );
		
		if (err) {
			return callback && callback(err);
		}

Results:
***callback at entry: undefined
***callback at exit: undefined
***Returned from createItems

Successfully created:
...

webteckie:
Actually the logging I added for the callback above was bogus.  I fixed the issue in updates.js.  See PR for #870.

PR #870, opened 2015-01-07
Log mongoose validation errors generated during createItems

webteckie:
Addresses #868 

Without the changes and if there is a mongoose validation error I get the following:

    ------------------------------------------------
    An error occurred applying updates, bailing on Keystone init.

    Error details:

With the changes and if there is a mongoose validation error I get something like the following, which is a huge help in troubleshooting:

    { [ValidationError: Validation failed]
      message: 'Validation failed',
      name: 'ValidationError',
      errors:
       { startTime:
          { [ValidatorError: Path `startTime` is required.]
            message: 'Path `startTime` is required.',
            name: 'ValidatorError',
            path: 'startTime',
            type: 'required',
            value: undefined } },
      model: 'Event',
      data:
       { name: 'Test Booking 1',
         status: 'pending',
         endTime:
          { _isAMomentObject: true,
            _isUTC: false,
            _pf: [Object],
            _locale: [Object],
            _d: Tue Jan 06 2015 19:09:37 GMT-0500 (Eastern Standard Time) },
         __ref: 'Event_1',
         __doc:
          { endTime: Tue Jan 06 2015 19:09:37 GMT-0500 (Eastern Standard Time),
            name: 'Test Booking 1',
            type: 'none',
            repeat: 'none',
            status: 'pending',
            guests: [],
            allDay: false,
            listName: 'Event',
            _id: 54ac7943a2751ad02b12ed57 } } }

### DIFF
--- a/lib/updates.js
+++ b/lib/updates.js
@@ -71,7 +71,10 @@ exports.apply = function(callback) {
 					update(function(err) {
 						if (err) {
 							console.log(_dashes_ + '\nUpdate ' + file + ' (background mode) failed with error:');
-							console.log(err);
+							console.error(err);
+							if ('stack' in err) {
+								console.trace(err.stack);
+							}
 						} else {
 							console.log(_dashes_ + '\nSuccessfully applied update ' + file + ' (in background).');
 							if (update.__commit__ !== false) {
@@ -84,7 +87,10 @@ exports.apply = function(callback) {
 					update(function(err) {
 						if (err) {
 							console.log(_dashes_ + '\nUpdate ' + file + ' failed with error:');
-							console.log(err);
+							console.error(err);
+							if ('stack' in err) {
+								console.trace(err.stack);
+							}
 							done(err);
 						} else {
 							updateCount++;


webteckie:
Give me a few...the "(_dashes_ + '\nUpdate ' + file + ' failed with error:\n')" line is not logging...hmmm

webteckie:
@JedWatson not sure why the dashes...are not logging.  If I do the following:

						if (err) {
							console.error(_dashes_ + '\nUpdate ' + file + ' failed with error:\n'+err);
							if ('stack' in err) {
								console.trace(err.stack);
							}
							done(err);
						} else {

I get the following, but doesn't have enough details, in which case I prefer the first commit at the expense of not logging the update file name:

    ------------------------------------------------
    Update 0.0.1-allvacationplaces_com failed with error:
    ValidationError: Path `startTime` is required.

I've tried multiple combinations of console.error, console.log, and no combination logs all 3 things (the dashes/update-filename, the error, and the stack).  Any ideas?

JedWatson:
@sebmck can you have a look at this?

sebmck:
`console.trace` doesn't accept any arguments. It just dumps the current call stack. `console.error(err.stack);` should work.

webteckie:
@sebmck according to http://nodejs.org/api/console.html console.trace does accept arguments.   Anyways,  some of the tests that I did dumped the trace correctly.   There's something going on with the logging that happens just before the trace is dumped.   I wonder if this is still a case of node exiting too quickly before the logging is dumped.  I will put a setTimeout to discard that possibility.   I will try again later tomorrow.  I will also try your suggestion in case there's something to that.

sebmck:
@webteckie Ah right, it accepts a `message` argument, still not very relevant. There's no real way the event loop could be empty so I'm skeptical that's the case. It's likely just that mongoose messes with errors and in some scenarios the call stack isn't bound to the error (so only the `message` appears in the `stack`).

JedWatson:
I think this change looks good, I've merged it so we can throw some more cases at it.

PR #876, opened 2015-01-08
Add code style checking to make lint

BerkeleyTrue:


BerkeleyTrue:
@JedWatson added AirBnB presets plus the tab spacing. Let me know if there are any settings that need to change.

PR #879, opened 2015-01-08
makes connect-redis work

gmmeyer:
Connect redis does not work, in any environment that I've tried, with the default options. There is always an error thrown within the interior code of connect-redis itself that propagates through and crashes keystone, at booting.

Here is an example of such a stack trace:
```
/home/greg/site/node_modules/keystone/lib/core/mount.js:197
                                throw e;
                                      ^
TypeError: Cannot read property 'prototype' of undefined
  at module.exports (/home/site/node_modules/connect-redis/lib/connect-redis.js:123:21)
  at [object Object].mount (/home/site/node_modules/keystone/lib/core/mount.js:172:44)
  at [object Object].start (/home/site/node_modules/keystone/lib/core/start.js:235:7)
  at Object.<anonymous> (/home/site/keystone_serve.js:3:13)
  at Module._compile (module.js:456:26)
  at Object.Module._extensions..js (module.js:474:10)
  at Module.load (module.js:356:32)
  at Function.Module._load (module.js:312:12)
  at Function.Module.runMain (module.js:497:10)
  at startup (node.js:119:16)
  at node.js:906:3
```
Tracing it back in keystone leads to this line:
https://github.com/keystonejs/keystone/blob/0.2.x/lib/core/mount.js#L172

Which, then, takes us to this line here:
https://github.com/tj/connect-redis/blob/master/lib/connect-redis.js#L128

The connect redis docs claim that you can use it with express if you install 1.4.7. I have not found this to be the case with keystone, as I get the same exact error when I use 1.4.7 as when I use 1.5.0, and 2.1.0. I do not know why this is the case, but, simply, it is. 

So, I have rewritten the relevant parts of mount.js to fix this.

### DIFF
--- a/lib/core/mount.js
+++ b/lib/core/mount.js
@@ -150,10 +150,8 @@ function mount(mountPath, parentApp, events) {
 			case 'redis':
 				// default session store for using Redis
 				sessionStore = 'connect-redis';
-				// TODO: are there default options that should be applied here?
 				break;
 			case 'connect-redis':
-				// TODO: are there default options that should be applied here?
 				break;
 
 			default:
@@ -168,8 +166,13 @@ function mount(mountPath, parentApp, events) {
 
 		// Initialize the session store
 		try {
-			
-			var SessionStore = require(sessionStore)(express);
+      var SessionStore = {};
+			if (sessionStore === 'connect-redis') {
+				var expressSession = require('express-session');
+				SessionStore = require(sessionStore)(expressSession);
+			} else {
+				SessionStore = require(sessionStore)(express);
+			}
 			
 			if (waitForSessionStore) {
 				sessionStorePromise = new P(function(resolve) {
@@ -183,9 +186,16 @@ function mount(mountPath, parentApp, events) {
 			
 			if (e.code === 'MODULE_NOT_FOUND') {
 				
-				// connect-redis must be explicitly installed @1.4.7, so we special-case it here
-				var installName = (sessionStore === 'connect-redis') ? sessionStore + '@1.4.7' : sessionStore;
-				
+				// connect-redis must be installed, as well as express-session,
+				// with keystone having access to it,
+				// so we special-case it here.
+        var installName = '';
+				if (sessionStore === 'connect-redis') {
+					installName = sessionStore + '@2.1.0 express-session@1.10.0';
+				} else {
+					installName = sessionStore;
+				}
+        
 				console.error(
 					'\nERROR: ' + sessionStore + ' not found.\n' +
 					'\nPlease install ' + sessionStore + ' from npm to use it as a `session store` option.' +
--- a/lib/core/mount.js
+++ b/lib/core/mount.js
@@ -166,7 +166,7 @@ function mount(mountPath, parentApp, events) {
 
 		// Initialize the session store
 		try {
-      var SessionStore = {};
+			var SessionStore = {};
 			if (sessionStore === 'connect-redis') {
 				var expressSession = require('express-session');
 				SessionStore = require(sessionStore)(expressSession);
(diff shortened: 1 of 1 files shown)


lib/core/mount.js
```diff
@@ -146,10 +146,11 @@ function mount(mountPath, parentApp, events) {
...
 				}
 				break;
-			
+
 			case 'redis':
 				// default session store for using Redis
-				sessionStore = 'connect-redis';
+			  sessionStore = 'connect-redis';
+
 				// TODO: are there default options that should be applied here?
```
morenoh149: could you look into whether or not we should pass default options here? if not please also remove both TODOs

lib/core/mount.js
```diff
@@ -168,9 +169,13 @@ function mount(mountPath, parentApp, events) {
 
 		// Initialize the session store
 		try {
-			
-			var SessionStore = require(sessionStore)(express);
-			
+      if(sessionStore === 'connect-redis') {
+				expressSession = require('express-session');
```
morenoh149: `express-session` is available above on https://github.com/keystonejs/keystone/blob/master/lib/core/mount.js#L57

gmmeyer: Ah, so it is. I'll modify it to change that, thank you!

gmmeyer: It's available in master, but not in the 2.x branch. Are you building future 2.x versions from the 2.x branch or from master?

morenoh149: future 2.x versions are being built in the 2.x branch I believe @JedWatson

gmmeyer: Ah, they're being built from the branch. Okay, I was editing 2.x to make it available in 2.x. I can also port my changes over to the master branch, is there an established process for doing this? Or, do I just port them there by hand?

gmmeyer: Yea, the change has already been made to master.

gmmeyer: There aren't really any defaults you'd wanna use. I guess we could put localhost and the port that redis uses by default (which I don't remember) in there. But in production you're going to have to set all of these with environment variables, regardless, so setting defaults would only affect developers on their local machine. And, that seems like the worst thing to do.

gmmeyer: I can go to the master branch and delete the todos, if you'd like.

sebmck: Noone's allowed to touch master except via pull request.

gmmeyer: Yes, I meant that I could submit a pr. Sorry if that was unclear.

morenoh149:
so connect-redis works on master and 0.2.x? if any aren't you could work on that. If you remove the todos and fixup the trailing whitespace that'd be a net win.

gmmeyer:
Well, I can't promise it works on master. I haven't written or tested the relevant code, but it works on 0.2.x. :)

I'm sorry, I don't understand what you mean when you say if any aren't, if any what?

Also, I have removed the todos.

morenoh149:
either. If master isn't working fix it. If 0.2.0 isn't working fix it.

gmmeyer:
I'll let you know host master works.

JedWatson:
@gmmeyer your editor seems to have mangled the Whitespace (see the "Files changed" tab)

Would you mind resubmitting with preserved whitespace so I can merge this?

gmmeyer:
Oops, sorry. I will fix that.

gmmeyer:
It now has the appropriate whitespace.

lib/core/mount.js
```diff
@@ -168,8 +166,13 @@ function mount(mountPath, parentApp, events) {
 
 		// Initialize the session store
 		try {
-			
-			var SessionStore = require(sessionStore)(express);
+      var SessionStore = {};
```
JedWatson: still spaces for indentation here

gmmeyer:
fixed, I must have added those after I ran tabify.

JedWatson:
Thanks @gmmeyer!

gmmeyer:
:smile:

PR #884, opened 2015-01-10
Fix many jscs warnings

morenoh149:
This is a big one. Here I reworded all `yoda conditionals`, trailing whitespace and random other things. The test suite is not complaining and I tried my best to make only changes that I think wont cause any breakage to keystone deployments. Down 627 warnings to 145.

### DIFF
--- a/lib/asyncdi.js
+++ b/lib/asyncdi.js
@@ -9,10 +9,9 @@ var _ = require('underscore'),
 // Thanks to Angular for the regex
 var FN_ARGS = /^function\s*[^\(]*\(\s*([^\)]*)\)/m;
 
-
 /**
  * Calling the module directly returns a new Wrapper instance
- * 
+ *
  * @param  {Function} fn
  * @return {Wrapper}
  */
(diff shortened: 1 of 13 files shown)


JedWatson:
Why are we rewording all the `yoda conditionals`? I like them :(

JedWatson:
My preference is also to keep whitespace on empty lines - saves re-indenting blocks for me a lot.

I'm worried I'm fighting a losing battle with that one...

(+1 for removing trailing whitespace on lines with code on them)

BerkeleyTrue:
Yoda's are not allowed with AirBnB code styling. see https://github.com/jscs-dev/node-jscs/blob/master/presets/airbnb.json#L27

So for trailing whitespace, it is. https://github.com/jscs-dev/node-jscs/blob/master/presets/airbnb.json#L25

Plus, these are super annoying.

-1 for trailing whitespace

creynders:
:+1: yoda conditionals
:-1: trailing whitespace

It's time for a keystone preset.

morenoh149:
@JedWatson you could evaluate and choose one of the many presets available here http://jscs.info/overview.html#presets We could also make a keystone-specific one if none of these fit your preferences

JedWatson:
Need to make some time to review the rules, today was a big codebase cleanup day so I'll get on it soon.

In the meantime:

:-1: _Remove Yoda conditionals_ - I find it clearer
:+1: _Remove trailing whitespace_ - if the line has characters on it, empty lines in the middle of code blocks that are just indentation are good IMO
:+1: _consistent quotes_, should be `'` but I'm OK with certain exceptions
:+1: _spaces around operators_ - always, also between arguments but not inside parentheses

What does "Var declarations should be joined" mean?

re: _All identifiers must be camelCase or UPPER_CASE_, variables should be `camelCase`, Classes should be `TitleCased` for clarity and consistency with React, and constants should be `UPPER_CASE`.

re: tabs vs. spaces, sorry but no - we have a mix of devs who prefer either 2 _or_ 4 space indentation (I am in the latter) and tabs are the only way to make everyone happy, because they're configurable. Start of line indentation should always use tabs, mid-line indentation should use spaces for consistency. This is set up with the `.editorconfig` so if you've got that plugin installed it should be pretty seamless.

morenoh149:
>.> I think waffle.io keeps opening this

PR #885, opened 2015-01-10
removes todos from redis

gmmeyer:
As we discussed in #879, I have removed the TODOS from the master branch.


JedWatson:
Cheers @gmmeyer

PR #886, opened 2015-01-10
Add pagination query docs to Database section

wuhaixing:
Add pagination query docs to Database section,and fix a path error in
docs/docs.js


JedWatson:
Great @wuhaixing, thanks!

PR #894, opened 2015-01-12
Warn if req.body is missing

morenoh149:
added warning if a post request is detected but no body is found. This happens if the multer middleware is not specified

### DIFF
--- a/routes/views/item.js
+++ b/routes/views/item.js
@@ -161,6 +161,9 @@ exports = module.exports = function(req, res) {
 			});
 			
 		};
+		if (req.method === 'POST' && req.body === undefined) {
+			console.warn('\nKeystoneJS Runtime Warning:\n\napp must have multipart form middleware installed. Try adding "multer" to your express instance.\n');
+		}
 		
 		if (req.method === 'POST' && req.body.action === 'updateItem' && !req.list.get('noedit')) {
 			


JedWatson:
Looks good but is there somewhere more high-level we can put this? We'll actually be phasing POST request out of the routes as we get more of the React UI communicating via API.

Maybe there is a technique we can use to generally validate middleware config..? (not sure how, just putting the idea out there)

morenoh149:
@JedWatson definitely. I was speaking with @BerkeleyTrue about doing this a better way. `mount.js` has checks for all of this. We need checks for when a user chooses not to call `Keystone.mount()`. Doing so would allow us to remove the 2 other warnings I proposed https://github.com/keystonejs/keystone/commit/72d8dbdb36bb5057120589656d725552d0c45c1b and here https://github.com/keystonejs/keystone/commit/935fa4bbcecfe6f66d525a1a071beca84d062a59

BerkeleyTrue:
Currently diving through Expressjs code to see if there is a way to identify middleware.  Good new, in express4 there is a way to view the what is in the middleware stack. Bad news, looks like a lot of them are registered without the name property and mutler and flash are among them.
node_inspector

morenoh149:
I don't like checking for a name actually. What if there were two middlewares that did the same thing? Keystone could have a flag for each middleware category and complain if not enough are detected. `mount()` handles a lot of this. Why exactly couldn't we figure out how to use an existing express app while calling mount? Perhaps the easier fix is to make mount accept an express instance?

BerkeleyTrue:
@morenoh149 Because mount adds a lot of behavior we don't need and it adds express middleware that our app was already using, this would have caused a duplicate middleware stack.

BerkeleyTrue:
This PR should be closed.

JedWatson:
Thanks for the bump @BerkeleyTrue

Issue #869, opened 2015-01-06
Relationship fields not validating unique when first created


snowkeeper:
As referenced from google groups the Relationship field is not validating unique on create.  Take this model for example:
```
var keystone = require('keystone'),
  Types = keystone.Field.Types;

var Country = new keystone.List('Country');

Country.add({
  name: { type: String, required: true, initial:true, index: true },
  primeMinister: { type: Types.Relationship, ref: 'User', required: true, unique:true, index: true, initial: true },
});
Country.defaultColumns = 'name, primeMinister';


Country.register();
```
Multiple **primeMinister** fields are allowed with the same `ObjectId`.

By adding a definition directly to the schema the desired results are seen and an error is received when trying to create the second item.
```
var keystone = require('keystone'),
  Types = keystone.Field.Types;

var Country = new keystone.List('Country');

Country.add({
  name: { type: String, required: true, initial:true, index: true },
  primeMinister: { type: Types.Relationship, ref: 'User', required: true, unique:true, index: true, initial: true },
});

Country.schema.pre('save',function(next) {
	this.primeMinister2 = '54997a0dc474c1ef1b126414';
	next();
});

Country.schema.add({primeMinister2: { type: keystone.mongoose.Schema.Types.ObjectId, unique:true, index: true, initial: true }});

Country.defaultColumns = 'name, primeMinister';


Country.register();
``` 

This could be associated with Issue #743

JedWatson:
Close by #901

PR #901, opened 2015-01-14
Passing relevant options from Types.Relationship to schema.path so mongo...

dgitsio:
It looks like the trouble I was having with ```Types.Relationship``` fields' required-ness and uniqueness in my Lists relates to the overridden ```addToSchema``` not passing along options specified during List creation.  By checking for those and passing them along in the call to ```schema.path```, the relevant field options seem to be recognized.  Thanks @snowkeeper for the help and filing #869 on behalf of this Issue.

### DIFF
--- a/lib/fieldTypes/relationship.js
+++ b/lib/fieldTypes/relationship.js
@@ -65,7 +65,10 @@ relationship.prototype.addToSchema = function() {
 
 	var def = {
 		type: this._nativeType,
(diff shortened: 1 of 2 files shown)


JedWatson:
Great work @dgitsio and @snowkeeper - amazing how obvious this was once I can see it in PR form, but a mystery until you went digging. Thanks.

PR #902, opened 2015-01-14
Renamed attribute

metasansana:
http://facebook.github.io/react/docs/tags-and-attributes.html#html-attributes


JedWatson:
Thanks @metasansana

PR #907, opened 2015-01-15
Mongo session store in production env fix

gooday:
When NODE_ENV=production and session store is set to mongo, there's a bug - neverending waiting for SessionStore due to never resolved sessionStorePromise

### DIFF
--- a/lib/core/mount.js
+++ b/lib/core/mount.js
@@ -190,7 +190,8 @@ function mount(mountPath, parentApp, events) {
 			
 			if (waitForSessionStore) {
 				sessionStorePromise = new P(function(resolve) {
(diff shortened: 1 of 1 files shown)


metasansana:
@gooday The api for `connect-mongo` changed. I ran into this a few days ago had to downgrade for now.

I made a note of it here:
https://github.com/keystonejs/keystone/wiki/Troubleshooting

JedWatson:
This issue has a bit of history. We have to wait for the connection or errors can occur in production; however, unfortunately the `connect-mongo` project recently deprecated the callback.

I've updated both `0.2.x` and `master` to correctly listen for the `on('connected')` event for both the mongo and reds session stores, and released a new version of Keystone with the fixes in it (0.2.41).

Hopefully now all will be well with the world :)

PR #917, opened 2015-01-17
Add support for cross-referencing relationships across update files

webteckie:
This is useful when you want to stack updates across different files.  For example, setup a base updates file and have custom updates files reference base list items.

### DIFF
--- a/lib/core/createItems.js
+++ b/lib/core/createItems.js
@@ -7,7 +7,7 @@ var _ = require('underscore'),
 	async = require('async'),
 	utils = require('keystone-utils');
 
-function createItems(data, ops, callback) {
+function createItems(data, refs, ops, callback) {
 	
 	var keystone = this;
 	
@@ -31,7 +31,6 @@ function createItems(data, ops, callback) {
 	}
 	
 	var lists = _.keys(data),
-		refs = {},
 		stats = {};
 
 	// logger function
@@ -62,7 +61,8 @@ function createItems(data, ops, callback) {
 					return doneList();
 				}
 				
-				refs[list.key] = {};
+				if (!refs[list.key]) refs[list.key] = {};
+				
 				stats[list.key] = {
 					singular: list.singular,
 					plural: list.plural,
--- a/lib/updates.js
+++ b/lib/updates.js
@@ -22,7 +22,8 @@ exports.apply = function(callback) {
 	var Update = mongoose.model('App_Update'),
 		updateCount = 0,
 		deferCount = 0,
-		skipCount = 0;
+		skipCount = 0,
+		refs = {};
 
 	var updatesPath = keystone.getPath('updates', 'updates');
 
@@ -54,7 +55,7 @@ exports.apply = function(callback) {
 					var background_mode = update.__background__ ? ' (background mode) ' : '';
 					
 					update = function(done) {
-						keystone.createItems(items, ops, function(err, stats) {
+						keystone.createItems(items, refs, ops, function(err, stats) {
 							if (!err) {
 								var statsMsg = stats ? stats.message : '';
 


JedWatson:
I'm a bit worried about this, because there's nothing explicitly linking updates together (i.e. ensuring that update 1 has been run in the same process as update 2) - and so this would lead to potentially inconsistent results in some environments. And although it "should" work most of the time, having something that could cause difficult-to-detect / debug issues in the platform is not something I'm comfortable with.

Maybe we can come up with a better way of abstracting this so updates can be stacked together, and enforce they get run together too?

I _do_ like the idea of being able to provide refs to `createItems`...

webteckie:
@JedWatson I do see your concern :-)  However, if you think deeper into it can you come up with a scenario where it may cause data integrity issues?  If you are worried about reusing a __ref in the "next" updates file in the group then that should be fine as the value of the last one should win.  If you use a __ref of the right type that will also fail validation thus won't go through.  Also the refs is scoped to the current updates group.  What else can you think of?  Would you rather have this be configurable?  Something like 'allow cross-referencing updates' and then if that's not set refs would always be initialized/reset per updates.  'allow cross-referencing updates' would default to false so it defaults to the current behavior.  Can you think of other alternatives of accomplishing the same thing.  My project now depends on this behavior :-)

webteckie:
Or are you  saying that updates may run out of order?  Yeah, that's an assumption that I made would not happen if you  semver the updates in stack increasing order.  For example:

    0.0.0-updates-base.js
    0.0.1-updates-custom.js

If you messed up and used versions that would result in out of order application then the other checks in place should catch it if cross-referencing is used.  Anyways, we want to get it right and perfect so I'm all ears :-)

morenoh149:
@webteckie haven't reviewed the pull request code in depth. But if you're taking inspiration from [Rails Active Record Migrations](http://edgeguides.rubyonrails.org/active_record_migrations.html) I'm on board. It's sounding a lot like it.

JedWatson:
@webteckie my concern is that although the scripts will always run in order, they may not run in the same process. e.g.

* `0.0.0-updates-base.js` runs successfully, then the process quits for some reason.
* `0.0.1-updates-custom.js` runs when keystone is started again, but because it's in a different process, the refs from `0.0.0-updates-base.js` are not available so it fails.

We would need to ensure that both scripts ran sequentially _in the same process_ for this enhancement to be safe.

> My project now depends on this behavior :-)

No problem, we'll work it out. I'll come back to you with some ideas shortly.

webteckie:
@JedWatson ok, I see.  But wouldn't that be the same issue now?  The only difference is that with the new stuff is that if 0.0.1-updates-custom.js tries to reference a __ref in 0.0.0-updates-base.js it would fail thus letting the user know something is wrong.  Before this change, what if the process fails after creating all items in 0.0.0-updates-base.js and before writing the file to App_Updates?  Perhaps, we need a different approach to make updates more resilient to such failures.  For example:

    1. Make __ref be part of the App_Update schema.  Also define a slug field.
    2. Before createItems is done, write the updates file to db.app_updates, if it hasn't been written before
    3. Pass the _id of the updates file in db.app_updates.<updates-file> to createItems (e.g., as fileId).  Also pass an indication as to whether this file was just created (newFile).
    4. (in case step 5 happened already for this item) this needs much refinement
         a. if the file was just created (newFile===true) continue on creating the Item; else...
         b. if a __ref is provided try to read it from db.app_updates({_id: fileId})
         c. use some query heuristics to attempt to match if this record has been created (e.g., query the list matching on all the fields)
         d. if the query in the previous step returned a record then take its slug and query the db.app_updates({_id: fileId, slug: itemSlug)})
         e. if step d return a valid item then assume the item has already been created and continue to next one; else keep creating the item.
    5. As createItems creates items it fills in __ref (if provided) in db.app_updates.__ref as well as write the slug of the created items into db.app_updates.slug

lucykav:
@webteckie I've been keeping an eye on your PR as this idea would be pretty useful for a site I'm currently building using keystone.

As you've closed it, I wondered if that was because you had an idea for an alternative approach to referencing relationships across updates? Or do you no longer think it makes sense?

webteckie:
@lucykav as you can read above @JedWatson had some valid concerns about this.  My project I guess has some specific requirements not , yet, required by the general community.  As such, I've, for now, included updates and createItems directly in my module and added this and some other stuff to meet my needs until #990 gets addressed. Not sure how, but I can always share what I have (maybe an npm module, which I've never done one for public consumption :-)

lucykav:
@webteckie thought as much, thanks - I'll watch https://github.com/keystonejs/keystone/issues/990. And yeah, let me know if you share the work you've done on updates for your own project - I definitely have a use for it!

PR #918, opened 2015-01-17
Bugfix .mimetype -> .type

aschwersenz:
I thinks that's a bug? The field is called type.


aschwersenz:
Ups, I added the stuff from my other pr as well...sorry

PR #919, opened 2015-01-17
A convenient way to add additional classes to the wysiwyg editor

aschwersenz:
I know, it is possible to achieve this with the existing options, but this is a shortcut to add new classes to the wysiwyg editor. The syntax is the same as with the current wysiwyg options. Just add

keystone.init({
	...
	'wysiwyg importcss': '/path/to/classes.css',
	....
});

All classes from the css file will be added to the Format > Formats menu entry.


aschwersenz:
Arg...sorry!

PR #920, opened 2015-01-17
Bugfix .mimetype -> .type

aschwersenz:
I thinks that's a bug? The field is called type.

### DIFF
--- a/lib/fieldTypes/s3file.js
+++ b/lib/fieldTypes/s3file.js
@@ -247,8 +247,8 @@ s3file.prototype.uploadFile = function(item, file, update, callback) {
 		prefix = field.options.datePrefix ? moment().format(field.options.datePrefix) + '-' : '',
 		name = prefix + file.name;
 
-	if (field.options.allowedTypes && !_.contains(field.options.allowedTypes, file.mimetype)){
-		return callback(new Error('Unsupported File Type: '+file.mimetype));
+	if (field.options.allowedTypes && !_.contains(field.options.allowedTypes, file.type)){
+		return callback(new Error('Unsupported File Type: '+file.type));
 	}
 
 	if ('function' === typeof update) {
@@ -263,7 +263,7 @@ s3file.prototype.uploadFile = function(item, file, update, callback) {
 		}
 
 		knox.createClient(field.s3config).putFile(file.path, path + name, {
-			'Content-Type': file.mimetype,
+			'Content-Type': file.type,
 			'x-amz-acl': 'public-read'
 		}, function(err, res) {
 
@@ -277,7 +277,7 @@ s3file.prototype.uploadFile = function(item, file, update, callback) {
 				filename: name,
 				path: path,
 				size: file.size,
-				filetype: file.mimetype,
+				filetype: file.type,
 				url: url
 			};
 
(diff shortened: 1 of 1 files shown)


morenoh149:
@aschwersenz I'm the one who changed it to `mimetype` This same issue was brought up by Tom from the keystone team. We have to investigate what's cause the discrepancy. I've dug into https://github.com/LearnBoost/knox and didn't find anything obvious. I think the `type`/`mimetype` depends on something maybe the machine. Could you look into what may be causing the discrepancy?

A simple solution would be to check both fields for existence and use the one that exists or a default. I would greenlight that.

JedWatson:
From what I've seen, you never get _both_ so perhaps the simplest fix to this is something like this at the top of the function:

```
var filetype = file.type || file.mimetype;
```

I'd love to know why it's different for you though @morenoh149. What platform are you on? I know @bladey is testing on Mac OS X 10.10 and heroku. If you're on windows, maybe that's related?

aschwersenz:
@morenoh149 But if I look into the s3file.prototype.uploadFile function, the mime check, where the type/mimetype field is used, happens before the file gets handed over to knox?

JedWatson:
Also relevant: https://github.com/keystonejs/keystone/tree/s3file-tweak

sebmck:
`req.files` is populated by `multer`, [here](https://github.com/expressjs/multer/blob/master/index.js#L86-97) is where the file object is built. Doesn't look like `type` is getting populated anywhere.

morenoh149:
if I add `console.log('file:', file);` on line https://github.com/keystonejs/keystone/blob/master/fields/types/s3file/S3FileType.js#L245

I get
```
GET /keystone/ 302 10.521 ms - 88
GET /keystone/signin 200 79.840 ms - -
POST /keystone/signin 302 283.313 ms - 74
GET /keystone/ 200 143.511 ms - -
GET /keystone/events 200 539.961 ms - -
GET /keystone/events/5488ec06f3885a763bf33eed 200 275.118 ms - -
GET /keystone/events/5488e422552e65db0535a024 200 242.621 ms - -
file: { fieldname: 'coverPhoto_upload',
  originalname: '10422484_10205195275308944_468927363624441868_n.jpg',
  name: '524e1cdba46afb8f711eb012f20850ae.jpg',
  encoding: '7bit',
  mimetype: 'image/jpeg',
  path: '/var/folders/m_/2dqg23cx2h3566ppmyxrb5zc0000gn/T/524e1cdba46afb8f711eb012f20850ae.jpg',
  extension: 'jpg',
  size: 46564,
  truncated: false,
  buffer: null }
POST /keystone/events/5488e422552e65db0535a024 302 845.870 ms - 138
GET /keystone/events/5488e422552e65db0535a024 200 163.073 ms - -
```
What else could I check to narrow down the discrepancy?

sebmck:
Looks like it's a result of 2.x using `formidable` for multipart that uses `type` and 3.x using `multer` which uses `mimetype`.

JedWatson:
@sebmck is correct and I suspect a fix for Express 4 was incorrectly merged into `master` while it was still the source for the 0.2.x (Express 3) versions. What a mess this has been.

I'm committing fixes that correct this across the board for 0.2.x (will release 0.2.42 today) then will merge into master for 0.3.0.

I'm also making it work consistently across _all_ field types, which is something that PRs to date have not done :(

PR #921, opened 2015-01-17
A convenient way to add additional classes to the wysiwyg editor

aschwersenz:
I know, it is possible to achieve this with the existing options, but this is a shortcut to add new classes to the wysiwyg editor. The syntax is the same as with the current wysiwyg options. Just add
```
keystone.init({
...
'wysiwyg importcss': '/path/to/classes.css',
....
});
```
All classes from the css file will be added to the Format > Formats menu entry.

### DIFF
--- a/lib/core/render.js
+++ b/lib/core/render.js
@@ -80,7 +80,8 @@ function render(req, res, view, ext) {
 			additionalOptions: keystone.get('wysiwyg additional options') || {},
 			overrideToolbar: keystone.get('wysiwyg override toolbar'),
 			skin: keystone.get('wysiwyg skin') || 'keystone',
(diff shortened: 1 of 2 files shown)


morenoh149:
What's the purpose of this? To style the wysiwyg editor?

JedWatson:
Looks good @aschwersenz.

@morenoh149 this adds styling options to the editor (TinyMCE reads the css and lets you apply its styles to your content from the format menu)

I'll merge into 0.2.x, @aschwersenz would you please also add the feature into the React implementation in `master`? I can't pull this directly across, and if we're adding it to `0.2.x` we'll need to bring it forward.

PR #923, opened 2015-01-18
Fixes #922 - A convenient way to add additional classes to the wysiwyg editor (0.3.0)

### LINKED ISSUE #922: v0.3.0: Implement importcss wysiwyg option in React
Added to 0.2.x by #921, we need to make sure this option is also implemented in 0.3.0

aschwersenz:


webteckie:
Should be tagged with Needs Documentation.

JedWatson:
I added the docs

PR #925, opened 2015-01-19
Use filename without suffix as default publicID for cloudinary

aschwersenz:
When I upload a picture to cloudinary using the built-in field type or via the wysiwyg editor plugin, the image gets by default the file name of the tmp file as its publicID, which is normally a very cryptic string. My suggestion would be to use the original file name of the image without the suffix as the default publicID?

### DIFF
--- a/lib/fieldTypes/cloudinaryimage.js
+++ b/lib/fieldTypes/cloudinaryimage.js
@@ -396,6 +396,8 @@ cloudinaryimage.prototype.getRequestHandler = function(item, req, paths, callbac
 				uploadOptions.tags.push(tp + 'dev');
 			}
 
+			uploadOptions.public_id = req.files[paths.upload].name.substring(0, req.files[paths.upload].name.lastIndexOf('.'));
+
 			if (field.options.publicID) {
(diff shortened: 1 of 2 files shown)


JedWatson:
This looks like a good idea. What happens to Cloudinary in the case of a conflict on `public_id` though? Is there any way this could be a breaking change?

It may be a good idea to expose this as an option, rather than force-changing the default behaviour...

aschwersenz:
I will elaborate this a bit more...

damianamallia:
Hi, is there a way to keep the original name at the moment? The idea to expose this as an option will be soon implemented?

JohnnyEstilles:
@JedWatson, I've been meaning to weigh in on this, but was waiting on @aschwersenz for an update.

This is a great idea, but I agree it should be an option and not the default. If @aschwersenz is unable to move forward with this, I can work on the issue.

Schwersenz-OGOIO:
Sorry, I was a bit occupied, I will push something in the next days!

aschwersenz:
I have created a new pull request as I changed this for 0.3. #1190

Issue #905, opened 2015-01-15
Choose coding style


morenoh149:
Re https://github.com/keystonejs/keystone/pull/884
any of these http://jscs.info/overview.html#presets or a custom one for keystone

creynders:
@morenoh149 I've just been experimenting with jshint, jscs, editorconfig and code style preference sets in Webstorm for one of my own projects. I think it's pretty doable to get everything uniform. Webstorm goes a long way in a) pointing out errors against any of the configured rules b) fixing these automatically. Not all errors are fixed automatically unfortunately (yet?)

Anyway, once a code style has been chosen I'll gladly _try_ to get the entire code base to adhere to it, unless you want to do it yourself.

creynders:
Oh, and if it works, I think it's a good idea to share the Webstorm code style preset file for others to use as well.

morenoh149:
@creynders definitely you could do it! I don't think there would be an easy way for myself to do it (I use vim)

creynders:
@morenoh149 AFAICT the jscs docs are wrong/unclear regarding the `disAllowMultipleVarDecl` rule. It says it's equal to jshint's `onevar` rule, but the example they provide as "valid" would not pass the `onevar` rule, which _does_ have to do with variable hoisting.

I have no idea whether it's WAI or a bug, but actually it's the opposite of the `onevar` rule, i.e. you can't declare multiple variables with a single var statement. Meaning, this is valid:

```
var a="foo";
doSomething();
var b="baz";
// fails `onevar`, passes `disAllowMultipleVarDecl`
```

However, this is not:

```
var a= "foo",
  b="baz";
// passess `onevar`, failes `disAllowMultipleVarDecl`
```

WAI or not; very confusing, to say the least.

morenoh149:
what's WAI?

creynders:
Works as intended

JedWatson:
I've implemented ESLint with appropriate config, and it lints our JSX as well which is great. Currently working through the warnings list with @vitalbone and @leo7568.

PR #951, opened 2015-01-25
stops server.listen() from being called more than once, so cluster works

gmmeyer:
Node cluster would error out, because app.listen() was called multiple times with the same port. The first time this happens, with the first node in the cluster, it wouldn't error out. However, the second time it happened, the server would error out.

So, I cleaned up start.js to call it only once. It also consolidated the message creation for the http server.

A small note: When I linted it, there were some errors. They were almost all, either, in parts of the code that I did not touch, or for trailing whitespace. Generally, if there were errors, I attempted to keep them consistent.

### DIFF
--- a/lib/core/start.js
+++ b/lib/core/start.js
@@ -172,10 +172,10 @@ function start(events) {
 				var httpsReadyMsg = (ssl === 'only') ? keystone.get('name') + ' (SSL) is ready on ' : 'SSL Server is ready on ';
 				
 				if (sslHost) {
-					httpsStarted(httpsReadyMsg + 'https://' + sslHost + ':' + sslPort);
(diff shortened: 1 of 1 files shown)


morenoh149:
@gmmeyer yeah don't worry about the linting errors. Those should go down after https://github.com/keystonejs/keystone/issues/905 gets closed.

The code looks good to me. Haven't used keystone with ssl yet though.

gmmeyer:
@morenoh149, I figured, I had seen #905 and #884 and figured that I should just try to do it as best as I could.

I have never used it either. I always do SSL though something else like nginx. Though I realized I made a mistake and had removed something, while debugging, that I definitely should not have. So I fixed it.

JedWatson:
Looks good @gmmeyer, thanks

Issue #943, opened 2015-01-23
moduleRoot issue


alanshaw:
The inaccessible `moduleRoot` is a little problematic when trying to test my keystone app.

I'm using tape and have an initial test that requires my app so that subsequent tests can test it.

The test is in `test/example.js` and it does a `var keystone = require('keystone')` along with other requires at the top of the file, **before** the app is required and started.

The problem is that because this is the first time keystone is required the `moduleRoot` is at `test/` and not the root of the project.

Can we talk about this? I mean, I could obviously `require('keystone')` in the tests after my app has been required, but that feels like a bit of a hack, and is going to be confusing to new developers coming to the project.

Could the moduleRoot be set automatically when `keystone.init` or `keystone.start` is called? That would probably fix the issue since `module.parent` would be my app at that point. What do you think?

morenoh149:
Related https://github.com/keystonejs/keystone/issues/652

JohnnyEstilles:
@alanshaw, check out my comment in #652.

JohnnyEstilles:
Closing ... duplicate issue. Essentially the same issue as #652. Let's continue the discussion in that thread.

Issue #652, opened 2014-09-25
keystone should not be sensitive to module.parent


indexzero:
This is a problem we encountered recently:

**bin/some-file.js**
``` js
var Schema = require('../schemas/some-schema'),
      keystone = require('keystone');

// Where there is a ../models directory
keystone.import('../models');
```

**schemas/some-schema.js**
``` js
var keystone = require('keystone');

/* Other code is irrelevant */
```

This caused the `keystone` module to have a parent that was different than our other main `index.js` script which because of [this code](https://github.com/keystonejs/keystone/blob/master/index.js#L14-L24) made many `keystone` helpers impossible to use until we reordered the requires.

This is a very subtle and very important bug because most novices will have absolutely no idea what is going on. You should not be using `module.parent`, but instead have some way to configure this that is required on setup to be  __the project root.__ Otherwise all of your requires can be borked for no reason just because of a different `require` ordering.

morenoh149:
can you propose a solution?

indexzero:
It's hard to say 100%, because I'm not everywhere that the `moduleRoot` variable it is being used as reference for is used. Basically you want the root of the module that you are being required in. 

_If you assume that keystone is always a first-level dependency then you can simply recurse up the directory tree to find it._

Honestly though I would just make a breaking change where you pass in the root into a `keystone.configure` method or something like that. Relying on any sort of internals of the `node` module system is almost always going to cause adverse side-effects.

JohnnyEstilles:
Keystone uses `moduleRoot` for a number of functions, including finding the `views` path, the `favicon` path, etc. Most Keystone developers (myself included) `require('keystone')` somewhere in the root of the app, so this isn't an issue for the majority of us.

That said, I propose we do two things:
1. Document the fact that `Keystone` expects to get required at the root of the app (which should be the standard convention)
2. Add an optional setting to Keystone (e.g. `keystone.set('moduleRoot', '../path/to/root')` for those who wish to deviate from the standard convention

If we do decide to add the `moduleRoot` setting, we also need to make sure we document well that it needs to be set with a *relative* path (that would be expanded by Keystone). Otherwise the app could run into issues when being deployed (or simply if it's moved it to a different system or folder).

[...]

PR #954, opened 2015-01-27
Adding "module root" setting (#652)

JohnnyEstilles:
As was pointed out in issue #652 by @indexzero, Keystone is sensitive to the apps module root path. This "sensitivity" is by design. Keystone utilizes the module root path to resolve the location of a number of files, the path of which is typically specified relative to the module root.

By default, Keystone uses as its root path the location of the first script that required it. While this is not an issue in most circumstances, there are instances in which it may be undesirable. Unit testing your app is one of these circumstances, as explained by @alanshaw in #943.

While we cannot eliminate this "sensitivity", we can provide a means to allow developers to specify a custom root path. That is purpose of this PR, which introduces the new `module root` setting.

The PR includes docs (so I won't bore you with those details here), as well as unit tests for the new feature. It is backwards compatible and does not introduce any breaking changes (AFAICT ... so, please let me know if you find any issues).

I typical don't like to add dependencies to Keystone in my PRs, but I felt that it was merited here. I added as a dependency a module called [caller-id](https://www.npmjs.com/package/caller-id), which I used to determine the path of the calling script, so that we could allow specifying a `module root` relative to the caller.

@indexzero and @alanshaw, please let me know if this works for you.

As always, comments, suggestions and constructive criticism is more than welcome.

### DIFF
--- a/docs/content/en/pages/docs/configuration.jade
+++ b/docs/content/en/pages/docs/configuration.jade
@@ -47,6 +47,14 @@ block content
 					td <code>brand</code> <code class="data-type">String</code>
 					td Displayed in the top left hand corner of the Admin UI
 				tr
+					td <code>module root</code> <code class="data-type">String</code>
+					td
+						p This setting tells Keystone the root path of your app. By default, <code>module root</code> points to the path of the first script that required Keystone within your app. This default may be undesireable at times.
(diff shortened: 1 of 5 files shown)


JedWatson:
Looks good to me, @JohnnyEstilles. Nice work.

PR #956, opened 2015-01-27
add option to configure the base url for keystone

hyun007:
instead of /keystone you can configure it to /admin by adding a 'base url': 'admin' to the keystone options


morenoh149:
looks good

hyun007:
never mind. there are so many other places where the /keystone path is hard coded.

Issue #958, opened 2015-01-28
Support unix sockets in Keystone.start()


adamscybot:
Currently doing this by using my own express instance. It would be nice to be able to specify this in the keystone config. Express supports this out of the box by passing unix://socket_file_example to app.listen().

High load applications fall over with high volume traffic when they use internal TCP connections with a reverse proxy (nginx, apache), as eventually the internal TCP connection limit is reached. Unix sockets are a common fix for this.

snowkeeper:
Added via #961

PR #961, opened 2015-01-28
Add support for starting app with unix socket

snowkeeper:
Adds option `keystone.set('unix socket','/path/app.socket')` refrenced in #958 

When set, http and https are ignored and the app is started listening to the socket. 

Will add docs if approved.

Should this be 0.3.0 only or ported to 0.2.x?


jjhesk:
i think it should keep it back in 0.3.0. maybe some more minor support and configurations should be add along with this 'unix socket' thing..

JedWatson:
Agreed, let's keep it on 0.3.0

PR #962, opened 2015-01-28
expose express session

snowkeeper:
Adds 
`keystone.set('express session', session(sessionOptions))` 
and updates
`app.use(session(sessionOptions));`
to
`app.use(this.get('express session'));`

Allows use of sessions for sockets, etc.
```
var sharedsession = require("express-socket.io-session");
var sock = require('socket.io')();

var io = sock.attach(keystone.httpServer);

/* set up session middleware */
io.use(sharedsession(keystone.get('express session'), keystone.get('session options').cookieParser));

```

### DIFF
--- a/lib/core/mount.js
+++ b/lib/core/mount.js
@@ -198,9 +198,11 @@ function mount(mountPath, parentApp, events) {
 		}
 	}
 
-	// expose initialised session options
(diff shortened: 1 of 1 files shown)


JedWatson:
:+1: good change

PR #971, opened 2015-01-30
set admin page title to project name

snowkeeper:
A small change until we move to a complete SPA.  Sets the Admin UI page title to the project name if available.


JedWatson:
Good call.

PR #974, opened 2015-01-31
Headings docs

snowkeeper:
I went back and forth on the placement and chose above Fields.  If you think it should be within or under Fields I can move it.  Here is the text if anyone has suggestions for improvement.

keystonejs setting up and using data models - google chrome_010


snowkeeper:
I think I used the wrong branch again.  Are you still wanting docs in 0.2.x @JedWatson ?

JedWatson:
@snowkeeper no, I realised there was nothing stopping me from building the docs from master so don't worry about the 0.2.x branch anymore.

I think your placement is good.

PR #976, opened 2015-01-31
Update ui-cloudinaryimage.js

michaelcho:
The missing "+" to join the strings on this line was giving me errors in Chrome in production and prevented the rest of the script from running (and therefore could not upload to Cloudinary). Adding the "+" fixes it....


morenoh149:
:+1:

PR #986, opened 2015-02-03
allow dependsOn to check for numeric value

snowkeeper:
`{ dependsOn: { sortBy: 2 } }` fails since the value is always a string.

This checks if `this.props.numeric` is set and saves the prop as a Number on `this.props.onChange`

### DIFF
--- a/fields/types/select/SelectField.js
+++ b/fields/types/select/SelectField.js
@@ -14,7 +14,7 @@ module.exports = Field.create({
 	valueChanged: function(newValue) {
 		this.props.onChange({
 			path: this.props.path,
-			value: newValue
+			value: (this.props.numeric) ? Number(newValue) : newValue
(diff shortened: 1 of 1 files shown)


JedWatson:
Good call

JedWatson:
Actually should we check the incoming value is a number before wrapping it in `Number()`?

Otherwise we're passing up NaN when the value should be `undefined` (for no option selected)

snowkeeper:
Yeah, your right.  Is Number to restrictive for this?

JedWatson:
I've just updated the logic here, as `Number('') === 0` which is not the correct behaviour (clearing the value, which submits an empty string, is not the same as selecting an option with the value `0`)

snowkeeper:
Thanks.  I was slow today.

PR #998, opened 2015-02-06
README update

snowkeeper:
Update 0.2.38 to 0.2.42

Change path for `keystone.prototype.start` to correct path.

Put field types in alphabetical order.


JedWatson:
Thanks @snowkeeper

PR #1005, opened 2015-02-06
Add back support for filters to the relationships field - attempts to fix #940

### LINKED ISSUE #940: v0.3.0: Select field not behaving properly
@jedwatson I think this is a blocker for 0.3.0  The select seems to have various issues!  For example, as soon as I create an item and submit the initial form then on the follow-up form the select options do not show.  If I type a character then it will show.  If I tried to scroll down the options it wont let me.

sebmck:
@JedWatson Not confident that it'll work as intended. Will need to be reviewed.

### DIFF
--- a/fields/types/relationship/RelationshipField.js
+++ b/fields/types/relationship/RelationshipField.js
@@ -65,9 +65,45 @@ module.exports = Field.create({
 		});
 	},
 	
+	buildFilters: function() {
+		var filters = {};
+
+		_.each(this.props.filter, function(value, key) {
+			if(_.isString(value) && value[0] == ':') {
+				fieldName = value.slice(1);
+
+				var val = self.props.values[fieldName];
+				if (val) {
+					filters[key] = val;
+					return;
+				}
+
+				// check if filtering by id and item was already saved
+				if (fieldName === ':_id' && Keystone.item) {
+					filters[key] = Keystone.item.id;
+					return;
+				}
+			} else {
+				filters[key] = value;
+			}
+		});
+
+		var parts = [];
+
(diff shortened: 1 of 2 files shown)


JedWatson:
Looks good @sebmck, thanks!

webteckie:
@sebmck @JedWatson:  Fix didn't solve the issue :-)  Really think we got to emphasize testing core pieces like this.

JedWatson:
Actually fixed now

webteckie:
@JedWatson I just refreshed from master and the issue still persists :-)

JedWatson:
Damn, what? I've tested it locally on mine and it works fine :/

You got the updated app.js right?

webteckie:
Is this the correct one:

$ md5sum node_modules/keystone/public/build/js/app.js
c1988835d9a2e7a08973eede9923fcfe *node_modules/keystone/public/build/js/app.js

$ ls -lrt node_modules/keystone/public/build/js/app.js
-rw-r--r--    1 Carlos   Administ   381155 Feb  8 17:56 node_modules/keystone/public/build/js/app.js

JedWatson:
```
MD5 (node_modules/keystone/public/build/js/app.js) = 6d603afa4b32b6d29e6c1f2c71e97c98
```

Maybe try pulling latest from master then running `gulp build-scripts` locally.

webteckie:
@JedWatson I got latest and rebuilt locally and now it works!  Hmmm...now I don't trust the process I was using to refresh keystone, which seems to have been working ok.  The only issue I still see is a select issue where no entries show up by default.  I need to type something in the input field to get entries to show.

Issue #1007, opened 2015-02-06
sortOrder not being initialized on item creation


JohnnyEstilles:
After [this commit](https://github.com/keystonejs/keystone/commit/8137c756656aaafadcba083a645858b66e00f183) (alt. implementation of #634, sortOrder type detection fix), `sortOrder` is no longer being initialized when an item is created.

The original code (below) tested the type of the `sortOrder` value, which is `undefined` before `sortOrder` is set, and `number` after it is.

```js
if (typeof this.sortOrder === 'number') {
    return next();
}
```

The current code tests the type of the schema field type, which is always `mongoose.Schema.Types.Number` regardless of whether `sortOrder` is initialized or not. This is causing `sortOrder` not to be initialed upon creation.

```js
var sortOrderType = this.schema.path('sortOrder');
if (typeof sortOrderType !== 'undefined' && sortOrderType instanceof keystone.mongoose.Schema.Types.Number) {
    return next();
}
```

I propose we rollback this commit.

JohnnyEstilles:
@JedWatson do you intend to merge this into both `master` and `0.2.x`?

JedWatson:
It's a breaking change so just 0.3.0

PR #1008, opened 2015-02-06
Fixing sortOrder init on item create

JohnnyEstilles:
fix for #1007

### DIFF
--- a/lib/schemaPlugins.js
+++ b/lib/schemaPlugins.js
@@ -16,9 +16,8 @@ exports.sortable = function() {
 	});
 
 	this.schema.pre('save', function(next) {
-		
-		var sortOrderType = this.schema.path('sortOrder');
-		if (typeof sortOrderType !== 'undefined' && sortOrderType instanceof keystone.mongoose.Schema.Types.Number) {
+
+		if (typeof this.sortOrder === 'number') {
 			return next();
 		}
 


JedWatson:
I'm happy to go with this, @JohnnyEstilles. #634 will be a problem again, however I'm happy (for now) to live with that limitation. Maybe we can come up with a better way of fixing the original issue...

JohnnyEstilles:
@JedWatson ... The problem is that #634 "apparently" solved the issue because it didn't save the `sortOrder` at all during the `pre save hook`. Can you explain the use-case that caused the issue with #634, so that I can try to come up with a more permanent solution?

JedWatson:
Yes - #634 was bunk :(

The use-case is that if you do a selective load of the document from the database (e.g. `model.find().select('_id').exec(...)`) then `sortOrder` won't be a number, and it'll be defaulted, overwriting the actual value.

So this is an issue whenever you partially load a document that *has* a sortOrder, don't load `sortOrder` itself, update the document and write it back to the database.

Maybe we can use the `isNew` special mongoose functionality to fix this more reliably - or maybe we can detect whether a document has only been selectively retrieved and just doesn't have the `sortOrder` path populated.

JohnnyEstilles:
`isNew` seems like it would do the trick ... I'll do some testing ...

Issue #1010, opened 2015-02-06
Delete button still shows even if nodelete is set to true


adamscybot:
If nodelete is set on a list, the delete button is still available in the bottom right when you go on a single item. When clicked, thankfully, it doesnt delete the object but just redirects to the list.

However, I dont think the button should be showing at all.

PR #1020, opened 2015-02-09
delete button state

snowkeeper:
Provides fix for #1010 .

Uses existing JQuery to add confirm dialog until a better React solution.


JedWatson:
I was hoping to avoid this but it's probably a good workaround :)

snowkeeper:
@JedWatson are you speaking towards the jquery confirms?  I wasn't very keen on adding it back in, but felt it needed some kind of accidental click insurance until a real solution is done.  If you know how you want to handle that and can jott it down I might be able to do it for you.

PR #1026, opened 2015-02-10
fixed typical example of a connection string

nicdutil:
In the docs, the typical example of a connection string for the mongo option will crash the app 
if used as written when deploying to a platform such as heroku.

the localhost connection string should always be the last value assigned to the mongo option (as done
in the mount function when the mongo option is not set).


JedWatson:
Correct, thanks @nicdutil!

PR #1039, opened 2015-02-11
ignore less of the codebase from jscs. add style task to makefile

morenoh149:
89 style violations

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,5 +1,8 @@
 {
-  "excludeFiles": ["coverage", "fields", "admin", "docs", "test", "node_modules", "public/build", "public/js/lib", "public/js/build"],
+  "excludeFiles": [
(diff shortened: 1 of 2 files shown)


sebmck:
Did you even test this? The `admin` and `fields` directories were specifically excluded because they contain JSX and will make jscs vomit.

creynders:
Wouldn't it be easier to rename all jsx files to `.jsx` and ignore that pattern?

morenoh149:
ah didn't notice they were jsx violations. And the complaints from the test suite are from re-declaring the `req` object many times. These files could have an `ignore jscs` declaration at the top. I know @JedWatson doesnt' want to jump on the jsx bandwagon. I guess the `fields` directory could be ignored altogether.

I'll factor out the `style` make task into another PR.

Issue #1036, opened 2015-02-10
connect.js - why is express "app" detection dependent on arguments[i].name === "app"


mrappweareratio:
I'm working with keystone and followed the website instructions for BYO express and mongoose, where it suggests to use:
```
app = express()
keystone.connect(mongoose, app);
```
However, during runtime (at least in a Mohca test), the app argument name and constructor.name are both blank "".

The following code doesn't detect.
```
if (arguments[i].name === 'app') {
	debug('detected express app');
	this.app = arguments[i];
}
```

Can this please be fixed?

I would suggest to simply update the documentation to either set keystone.app directly, or update the connect functionality to take mongoose and express separately.

morenoh149:
ah I came across this too when I was trying to BYO express instance. @mrappweareratio I proposed refactoring the logic to trust the developer and perform no introspection on the incoming express instance. Introspection would work if every module had a name field. But from @BerkeleyTrue research not all modules provide the `name` field.

@mrappweareratio want to write up the pull request so there are no checks on `name` and the software just assumes the object implements the express api?

BerkeleyTrue:
The connect method is deprecated, or at least it was going to be and fell through the cracks for 0.3.0 as express 4 and on no longer has the name field and this is broken. You can already keystone.set('mongoose', mongoose), but I don't think you can use the same method for the express app. 

@JedWatson Is it possible to use keystone.set('express app', app) yet?

see Jeds comment here: https://github.com/keystonejs/keystone/issues/864

JohnnyEstilles:
I agree with @morenoh149. The introspection in `keystone.connect()` is completely unnecessary.

Right now we can set both the Express and the Mongoose instance directly on the Keystone object. (Which is what I've been using to avoid the `.connect()` issues.)

```js
keystone.app = require('express')();
keystone.mongoose = require('mongoose');
```

Interestingly enough, `keystone.init()` uses the `.connect()` method to set the default Mongoose instance, but not the Express instance (see [here](https://github.com/keystonejs/keystone/blob/d51fcf58181e3d62f71a66c93d121c5922044a25/lib/core/init.js#L22-L28)).

AFAICT ... [this](https://github.com/keystonejs/keystone/blob/d51fcf58181e3d62f71a66c93d121c5922044a25/lib/core/init.js#L27) is the only place within Keystone that the `.connect()` method is used. And, personally, I think its convenience has outlived its usefulness. 

I vote for deprecating it in favor of Keystone-like options.

```js
[...]

PR #1041, opened 2015-02-11
Adding app/mongoose project options and deprecating connect()

JohnnyEstilles:
Fixing issue #1036

### DIFF
--- a/lib/core/connect.js
+++ b/lib/core/connect.js
@@ -15,7 +15,8 @@ var debug = require('debug')('keystone:core:connect');
 
 function connect() {
 	var warningMsg = 'keystone.connect() is now deprecated and will not be available in future versions of KeystoneJS\n' +
-		'It has been replaced with the Keystone "app" and "mongoose" options.\n\n' +
+		'It has been replaced with the Keystone "app" and "mongoose" options.\n' +
+		'Due to changes in Express 4, "keystone.connect()" no longer works as expected.\n\n'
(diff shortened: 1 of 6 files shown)


lib/core/connect.js
```diff
@@ -14,6 +14,12 @@
 var debug = require('debug')('keystone:core:connect');
 
 function connect() {
+	var warningMsg = 'keystone.connect() is now deprecated and will not be available in future versions of KeystoneJS\n' +
```
JedWatson: I think we should actually throw here, since the `connect` function won't work at all in 0.3

JedWatson:
For consistency we should also make sure that `init()` uses `set` too, otherwise `get('express')` won't always work.

```
	if (!this.app) {
		this.set('express'), require('express')());
	}
	
	if (!this.mongoose) {
		this.set('mongoose', require('mongoose'));
	}
```

lib/core/connect.js
JohnnyEstilles: Agreed!

JohnnyEstilles:
Agreed! Will push commits in a few ...

lib/core/connect.js
JohnnyEstilles: I'll updated the deprecation warning in docs as well.

JohnnyEstilles:
Done!

JedWatson:
:+1:

PR #1043, opened 2015-02-11
Fixing updateItem() in TextArrayType

JohnnyEstilles:
Used `.getValueFromData()` to get the current data (works well with nested paths)

### DIFF
--- a/fields/types/textarray/TextArrayType.js
+++ b/fields/types/textarray/TextArrayType.js
@@ -31,10 +31,12 @@ util.inherits(textarray, super_);
  */
 
 textarray.prototype.updateItem = function(item, data) {
-	if ( data[this.path] === undefined ) {
-		item.set(this.path, []);
+	var value = this.getValueFromData(data);
+	
+	if (value !== undefined && value != item.get(this.path)) { // jshint ignore:line
+		item.set(this.path, value);
 	} else {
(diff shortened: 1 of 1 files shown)


JedWatson:
Eek! This is going to destroy data when `value === undefined` (i.e. if no value was submitted to be updated)

Also value will always `!= item.get(this.path)` because it's an array.

JohnnyEstilles:
> Eek! This is going to destroy data when value === undefined (i.e. if no value was submitted to be updated)

Understood! So we should simply not save anything ... ???

> Also value will always != item.get(this.path) because it's an array.

`item.get()` wouldn't return an array?

JohnnyEstilles:
What if I just update it ONLY when it's undefined?

```js
textarray.prototype.updateItem = function(item, data) {
	var value = this.getValueFromData(data);
	
	if (value !== undefined) {
		item.set(this.path, value);
	}
};
```

JedWatson:
> item.get() wouldn't return an array?

It would, but it wouldn't be the *same* array, so the simple comparison would fail.

I've just added some more tests for TextArray (including another that fails currently) that will hopefully clarify the required behaviour.

JohnnyEstilles:
Rescinding PR ... will give it another shot.

PR #1047, opened 2015-02-11
Fixing TextArrayType validateInput() and updateItem()

JohnnyEstilles:
Attempt number two! How about this @JedWatson?

### DIFF
--- a/fields/types/textarray/TextArrayType.js
+++ b/fields/types/textarray/TextArrayType.js
@@ -34,7 +34,7 @@ textarray.prototype.validateInput = function(data, required, item) {
 	var value = this.getValueFromData(data);
 
 	if (required) {
-		if (value === undefined && item && item.get(this.path)) {
+		if (value === undefined && item && item.get(this.path) && item.get(this.path).length) {
 			return true;
 		}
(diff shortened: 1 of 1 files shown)


fields/types/textarray/TextArrayType.js
```diff
@@ -25,16 +25,50 @@ function textarray(list, path, options) {
...
+ * Validates that a value for this field has been provided in a data object
+ *
+ * @api public
+ */
+
+textarray.prototype.validateInput = function(data, required, item) {
+	var value = this.getValueFromData(data);
+
+	if (required) {
+		if (value === undefined && item && item.get(this.path)) {
```
JedWatson: This will return a false positive because `item.get(this.path)` should always return an array even when empty, which would pass the truthy check. Need to validate length of the existing array.

this should fix it: `value === undefined && item && item.get(this.path) && item.get(this.path).length`

JohnnyEstilles:
Corrected.

fields/types/textarray/TextArrayType.js
JohnnyEstilles: Corrected!

JedWatson:
Looks good to me. Nice work!

PR #1052, opened 2015-02-12
throw error to user if Amazon doesn't save a file correctly

morenoh149:
This will show the user an error if the file failed for whatever reason.
screenshot 2015-02-11 19 04 53


JedWatson:
Good call @morenoh149

PR #1053, opened 2015-02-12
Correcting numberArrayType validateInput() and updateItem()

JohnnyEstilles:
Updated numberArrayType to conform with test-spec


JedWatson:
Nice work @JohnnyEstilles!

PR #1054, opened 2015-02-12
Correcting getDefaultProps() on DateField and DatetimeField

JohnnyEstilles:
Thanks to @sebmck for pointing this out


JedWatson:
Thanks for fixing it

PR #1055, opened 2015-02-12
Fixing relationship field dup key test

JohnnyEstilles:
Change to check `err.code` (which is always `11000` for dup key errors) instead of `err.err`, because the actual message sometimes differs from platform to platform.


JedWatson:
Good call

PR #1058, opened 2015-02-12
Type Url: a link to the URL after the input

ignlg:
Added a link to the URL field value after the input (like the one for Relationship).


JedWatson:
@ignlg any chance you'll move to 0.3 soon? There's a fair bit of overhead adding features to 0.2.x at this point. I basically need to immediately land support for any new features in 0.3 to keep them in sync.

JedWatson:
React would also make this much nicer to implement, with live updating :wink:

ignlg:
_tl;dr_ I'll try to keep 0.2.x features/fixes to the minimum and maybe 0.3.x will enter a more stable stage when we need them. But it depends on the clients side in the end.

In fact we're now developing a huge app made with Backbone & React but the backend is in Play (Java) so I'm stuck until a new Keystone-able project arrives. As soon as it does I'll be back with features for 0.3.x.

But I perfectly understand your point of view and I'm trying to figure out a solution to keep our clients support with their Keystone 0.2.x and contribute to the framework at the same time, other than keeping our own fork.

0.3.x is not ready for production AFAIK and, even if I have the time to _reactify_ everything, it's a double effort that will impact us, in time and money, more than I can allow. On the other side, if we go too far away with a fork, it'd be impossible to upgrade it in the future. It's hard to manage.

mxstbr:
This is no longer relevant now that we've almost finished the move to React for 0.4 – thanks so much for taking the time to submit this feature and very sorry for not landing it sooner!

PR #1048, opened 2015-02-11
Add View Item button based on preview list option

albertogasparin:
Issue #320

PR #1059, opened 2015-02-12
List autokey link

ignlg:
List autokey `link` property

`String or Function`

Shows the key as a quick access link in the info toolbar of the item view (top right). If a `String` is provided, the key is concatenated at the end. Otherwise the `Function` receives the item as `this` context. Defaults to `null` and the key is shown as plain text.


JohnnyEstilles:
@ignlg, not sure if you were made aware, but branch 0.2.x is frozen (i.e. feature locked). AFAIK, we will accept PRs for bug fixes on 0.2.x, but not for new features. Maybe you can submit a PR for the feature on the master branch (0.3.x).

ignlg:
@JohnnyEstilles tl;dr I could commit this to the 0.3.x (master) branch but I'll need extra time to test it.

I added this for a project but it's in production and I cannot update Keystone to 0.3.x until it shows it's stable enough and fits well with our customisations. I supposed it could add some value to 0.2.x users meanwhile.

JohnnyEstilles:
@ignlg ... I understand perfectly. Ultimately it's @JedWatson's call.

JedWatson:
I'm happy to add this to 0.2.x if it helps you, I haven't had a chance to review it yet but there's an open issue about connect-redis in 0.2.42 so it looks like there'll be another release anyway.

This seems to overlap with #1048 / #320 in terms of functionality - I like the simplicity of this solution but am not sure about having two options that do the same thing, and having the view option not bound to the `autokey` option seems like a good thing.

What do you think @ignlg?

ignlg:
@JedWatson It looks good but it's a React component. So, what I'm thinking about is to rewrite this feature for 0.2.x to follow the #320 `preview` rules, adding a new link to the bottom toolbar.

But it's not a _preview_ if we stick to the meaning so I'm a little hesitate to use that term. You're not previewing the changes before Save, or have I missed something?

What about `viewLink` or just `link`? At the end it's a custom link to somewhere related to that item.

ignlg:
I close this PR because the branch's name has lost its meaning. I'll do a new one.

Issue #1038, opened 2015-02-11
Click `Today` button in AdminUI enters infinite loop


morenoh149:
I've been testing out the 0.3.0 branch on a third project. Any one else can confirm clicking the `Today` button in the date picker enters an infinite loop?
![screenshot 2015-02-10 21 45 14](https://cloud.githubusercontent.com/assets/587438/6142620/98737cb2-b16e-11e4-9b00-77620e6b513c.png)
See the lower right corner above. The UI is frozen in the screenshot and crashes chrome.

perryao:
Ran into the exact same issue a moment ago. Don't have a fix for it, but figured I'd let everyone know you weren't alone in experiencing this.

morenoh149:
is there a way to set a debug flag that will have keystone deliver an unminified `app.js` file? it's just saying `app.js:1` in the console.

JohnnyEstilles:
@morenoh149 I've been testing all the fields in the last couple of days and I haven't encountered that problem at all. Do you have any other details?

![image](https://cloud.githubusercontent.com/assets/5514878/6144020/6ee9b5f2-b20c-11e4-8c35-6cbc5dc1c6e1.png)

morenoh149:
@JohnnyEstilles you can clone it and run it locally https://github.com/r3dm/keystone-demo trying to update it to 0.3.0. Unless the infinite loop is caused by something I did to the environment.

JohnnyEstilles:
Just cloned and tested your demo, and I'm sorry to say I can't replicate your problem. Sorry.

Tried on Windows and OS X ... and in multiple browsers (Safari, Chrome, Firefox and IE) ... and it still works for me. Screenshot below is obviously Chrome on OS X.

![screenshot 2015-02-12 08 59 33](https://cloud.githubusercontent.com/assets/5514878/6159812/d3256e56-b295-11e4-9129-335e3b9cc4ca.png)

morenoh149:
maybe it's my environment. Firefox too 
![screenshot 2015-02-11 17 13 21](https://cloud.githubusercontent.com/assets/587438/6159934/54824c6a-b211-11e4-8904-6ac236e8a02c.png)

webteckie:
fyi, I cloned @morenoh149 repo and, yeah, there's something wrong with the Today click.  I don't see the logging but the thing does go into an infinite loop.

JohnnyEstilles:
[...]

PR #1061, opened 2015-02-12
fixed app hanging when date is selected

lojack:
fix for #1038


morenoh149:
@lojack thanks for fixing up the code!

Don't we need the `isValid`,`setDate`, etc methods for other things though?

JohnnyEstilles:
I'm still trying to replicate @morenoh149's issue on #1038, but unfortunately haven't been able to (still trying though). I would love to troubleshoot this, but I haven't encountered the issue at all, even under the circumstances described by @gregoj.

While I really appreciate @lojack taking a crack at this, I'd be careful merging this PR.

First, it doesn't really explain how the issue was reproduced and what problem was found. Second, it removes code that is required for specific use-cases and that have no apparent connection with the problem at hand.

Have any of the folks that are encountering this issue (@morenoh149, @webteckie, @gregoj) tested this fix?

webteckie:
@JohnnyEstilles I have not verified the fix by @lojack  Does it correctly handle user-entered invalid dates?

JohnnyEstilles:
@webteckie ... It does (to a certain extent). But what worries me more is that the current code does two things that this fix removes: (1) it makes sure that date is converted to/stored in ISO format (required by the server and allows us to handle timezone difference issues), and (2) it makes sure to state changes are pushed up to the `<EditForm />` component, which is serving as a sort of view-controller/single source of truth for that view's state. I assume this state will eventually be used to update the server, so keeping it consistently updated is important, regardless of how we currently send the data back to the server.

I tested the fix and I found several of issues. The first is that depending on the timezone difference between the client and the server, the date store may be displayed as "one day before" the date that is actually saved. There are other issues, but too many to enumerate here right now. There is no infinite loop problem in the fix, but bear in mind that I've never actually encountered the problem myself.

I understand the current code has a bug causing the infinite loop that is being reported. I just think we need to do a better job of troubleshooting and identifying the source of the bug.

morenoh149:
@JohnnyEstilles indeed. Is there anyway to make keystone deliver an unminified app.js file? Currently I can' get very far debugging since the file is so hard to read. Chrome Prettifying the code doesn't stick around when I run the debugger as well. It'd be great to enable a project-wide `debug` mode.

JohnnyEstilles:
@morenoh149, it looks to me like the build process automatically uglifies `app.js`. If you're using the gulp `watch-scripts` task, I suppose you can temporarily comment out lines that uglify `app.js` in the `rebundle()` function.

JohnnyEstilles:
@morenoh149 ... I also added some background/comments in #1038 that might be of help to anyone troubleshooting this.

morenoh149:
clicking okay from the wysiwyg source code editor modal also freezes the browser 
screenshot 2015-02-12 21 22 21

lojack:
Sorry, was a little trigger happy with the PR. I'll take another look at this today, with the background given by @JohnnyEstilles in mind.

Does it make sense to look into implementing a Flux style pattern so we can avoid these recursive cascading updates in the future?

JohnnyEstilles:
@lojack, you may want to hold off for a bit ... @JedWatson and I were discussing the possibility of replacing `pickaday` within the `Date` field component with a React date picker. I'm testing one right now with the [React Widgets Date & Time Picker](http://jquense.github.io/react-widgets/docs/#/datetime-picker).

@JedWatson, I'm going to submit a PR just so that you guys can test and tell me what you think. If you guys like it, the I will convert the `DateTime` field as will, and submit it along with unit tests (speaking of which, we need to discuss unit testing of React components).

JohnnyEstilles:
The `Date` field using the React Widget Date & Time Picker is on the `feature/datefield-wip` branch on [my Keystone fork](https://github.com/JohnnyEstilles/keystone/tree/feature/datefield-wip). Please check it out and tell me what you think.

morenoh149:
@JohnnyEstilles where in the codebase is the build process for app.js? I'm not using the generator of the gulp watch task. I have a custom gulp for my project.

JohnnyEstilles:
@morenoh149, yes! You can use either the `build-scripts` or `watch-scripts` after you install the dependencies with `npm install`.

morenoh149:
@JohnnyEstilles I just built app.js locally and both bugs (saving with the wysiwyg editor in source code mode, and clicking today field on your experimental branch) are solved. I think the solution is to run the build step on the end-user's machine but that's annoying as browserify is supposed to build a script that works anywhere. 

Diffing is near impossible.
screenshot 2015-02-13 20 46 51
A good next step would be to disable uglify https://github.com/keystonejs/keystone/blob/master/gulpfile.js#L32 so we could debug more easily. Afterwards decide if we want to have a `postinstall` script added https://docs.npmjs.com/misc/scripts

JohnnyEstilles:
@morenoh149, thanks for the help troubleshooting this. Regarding the points you brought up:

- changing the build step to `postinstall` is something we would needs to discuss with @JedWatson 
- removing `uglify` is something we could probably consider for the `watch-scripts` gulp task to help with troubleshooting
- keeping the existing `Date` field versus moving to a pure React component is also something that needs to be discussed more with @JedWatson. I don't want to speak for @JedWatson, but in a recent conversation he seemed more in favor of using pure React components (with which I agree). This is why I opened issue #1065, proposing the use of the date time picker included with React Widgets in lieu of `pickaday`.

morenoh149:
@JohnnyEstilles Re `postinstall` https://github.com/substack/ploy#start *`postinstall` is a good place to put bundle steps like browserify.*

JohnnyEstilles:
@morenoh149, Re: moving build to `postinstall`, I do not disagree, I just think we should consult with @JedWatson to make sure we're not overlooking anything.

morenoh149:
oh yeah of course. I was just sharing what I found regarding it.

webteckie:
fyi, the following sequence spins -- onInputChange/setDate happens repeatedly, which I think is what @lojack saw as well and attempted to fix.

image

JohnnyEstilles:
@webteckie ... thanks!. Although I can't really see the loop on the image you sent. I'll take your word for it. So, were you able to identify to identify the set of circumstances that lead to the loop? I ask, because as much as I've tried I haven't been able to reproduce the problem even once.

webteckie:
@JohnnyEstilles the image wasn't meant to show the loop but rather an instance of it while stepping through it after capturing the initial event. In slow motion clicking the button I can see that the date is set in the input field but then I guess the onChange events kicks in and the loop starts. The onInputChange/setDate sequence need to be stopped  somehow once it is done once.

JohnnyEstilles:
@webteckie ... thanks for the explanation. Unfortunately, I don't see/experience that loop when I use the `Date` field component. So, like I've mentioned before. We need to identify the set of circumstances that lead to the loop.

JedWatson:
We can't merge this as-is, there's a lot of required functionality missing.

`format` is used by `renderUI` when `this.shouldRenderField()` returns false, and other issues.

We also need to push the change event back up via `props.onChange` or the rest of the form won't know when the date changes (breaking `dependsOn` and future ajax data submission)

Re `postinstall` discussion, the real answer is to build scripts dynamically when keystone starts based on the _fields you need_.

johngroves:
Awesome, thanks all.

joshlasdin:
I'm running into this problem, does anyone have a workaround in the meantime?

JedWatson:
Has been fixed, see #1038

PR #1063, opened 2015-02-13
fixed missing dot in regex for keystone urls

nicdutil:
The current regex: /^\/keystone($|\/*)/

Without the missing dot in the regex, paths like /keystoneaaaa will match the regex in addition 
to all /keystone/* urls.


JedWatson:
Thanks @nicdutil

Issue #472, opened 2014-07-15
Cannot delete s3 files


kumo:
I created a project using yo and modified the Post model replacing the CloudinaryImage type with S3File. I can upload files to S3 and access them but I am unable to delete them. Should I have done something else?

morenoh149:
can anyone confirm if this is still the case on 0.3.0?

kumo:
@morenoh149 what is the easiest way of checking with 0.3.0?

morenoh149:
take an existing keystone project. change the package.json dependency to something like `"keystone": "git://github.com/keystonejs/keystone.git#master"` try to upload an s3file and share a screenshot if it doesn't work.

kumo:
I can confirm that it still doesn't work and also if I replace a file the old one remains. There are no errors, but the files remain in the S3 Management Console.

Additionally, I was only able to upload a file to an S3 bucket in the standard US region but that seems related to #532

morenoh149:
solved by https://github.com/keystonejs/keystone/pull/1064

PR #1064, opened 2015-02-13
[s3FileType] delete file on remove, change

montmanu:
'Remove File' and 'Change File' actions should delete previous file from S3.


montmanu:
Fixes #472

montmanu:
Think we could open this back up since it was removed from #1167 ?

eggytronixx:
How can I get this update.

PR #1067, opened 2015-02-14
fixed undefined keystone variable

nicdutil:
Throws an error when starting the app since keystone is not defined in connect.js.


JohnnyEstilles:
@nicdutil ... great catch! My PR was missing this.

JedWatson:
Thanks @nicdutil

PR #1073, opened 2015-02-17
relationship field should update when order, but not contents, has changed

lojack:
Prior to this change the relationship field wasn't being updated when only the order was changed. The only potential gotcha that I'm aware of is that this no longer updates the array in-place. I don't see why this would be necessary, especially since everything is moving to React, but I could be mistaken.


JedWatson:
Good point, @lojack.

PR #1076, opened 2015-02-18
added note. Select FieldType is very similar to Enum in other frameworks...

morenoh149:
....


morenoh149:
Just want to share this too http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html

PR #1079, opened 2015-02-19
Remove code mirror dependency from libs and use npm version

theolampert:
This removes the codemirror dependency that was still being loaded into the admin ui from the libs folder in favour of the npm version. I had to change the browserify shim to get this working as well as modify some core css. Codemirror's css is inlined in the head from it's node_modules folder, I'm not sure this is the cleanest approach?.


JedWatson:
We removed this specifically so as to not blow out the size of the build file by always including CodeMirror even when Code fields aren't in use.

We'll be able to add it back in when the build is dynamically generated based on which fields are included.

What's the reason for making this change *now*?

theolampert:
Hey Jed,

Codemirror as well as a few others are still loaded into the admin ui regardless of whether or not the field is present in your models. I figured we could just start moving all of these to the browserify build since these libs are all being loaded anyway. This will need to happen eventually I guess but maybe jumping the gun a bit ?

JedWatson:
Including it anyway was a bug I hadn't noticed - I'll fix it up in the next version! Also planning to pull the major dependencies (React / etc) out of the Browserify build and load from CDN in production to improve speed of build and delivery.

Issue #957, opened 2015-01-28
On-start building of Admin JS app.js


JedWatson:
We need to change the build process for `app.js` so that it is built when Keystone is started, including _only_ the fields that are actually used by models.

This will both remove the need for a built file to be included in the npm package (and github project, see #955) _and_ let us dramatically reduce the file size of app.js in most cases (it is currently 1.6mb)

i.e. if you don't need CodeMirror because you don't have Code fields, we _really_ shouldn't be adding that 300kb to the file size.

My current idea is to dynamically write the `/admin/src/fields.js` file (or equivalent) when Keystone is started, then use browserify to build `/public/build/js/app.js` (or equivalent) on `start()` after the models have been defined. `fields.js` would be built to only require the field types that have been used by models (this feature is already in place).

In the future it would make room to dynamically include custom (non-core) field types in the build tool, which we'll need.

The actual method may be different, e.g. we may build a separate `fields.js` bundle that dynamically requires and exposes just the field types in use, leaving the build of the main `app.js` separate. This would be good to experiment with during implementation.

A similar process should replace the less middleware that builds the stylesheet too, but that's another issue.

sononix:
To some extent shouldnt this be taken care by including the relevant task in the gulp/grunt files - Ive only just download the repo so I still feeling my way around the architecture - kicking the wheels to some extent

does that mean that app.js needs to be 'restarted' everytime a change is made to app.js - similar to raising a sails app each time -- what is best practice in development then

morenoh149:
@JedWatson could we get an eta for when this would get done?
If the eta is > 1 week could we perhaps work on getting a `postinstall` step working. I believe we could get the `postinstall` done in < 1 week. @lojack @JohnnyEstilles 

As of right now the master branch of this project is not working for a significant number of devs.

JedWatson:
@morenoh149 happy for you to set it up as a postinstall and see if that solves the problem. I haven't had a similar experience with any of my other react packages (all are built before submitting to npm) so I don't know why this is a special case. However, I can't reproduce the issue at all (while many others can) so anything we can do to narrow down the cause or even provide a band-aid at this point would be good.

JedWatson:
The first part of this is complete, we've now got a proper Browserify build running when Keystone starts.

The fields being bundled still need to be dynamically built based on the field types in use, see https://github.com/keystonejs/keystone/blob/master/admin/app/static.js#L18

[...]

Issue #955, opened 2015-01-27
Can we remove the build files from the repo?


danielmahon:
Shouldn't building be handled by an npm hook? I have been working on a few feature branches that make it hard to create a clean rebase/history when Im expected to checkin the built files (less, app.js).

morenoh149:
so you're working on parts of the code that would trigger a rebuild? Which file exactly?

danielmahon:
creating new field types, which includes changes to react and less, both of which get "compiled" and commited

morenoh149:
ah and this file https://github.com/keystonejs/keystone/blob/master/public/build/js/app.js has be rebuilt? any others?

danielmahon:
It makes sense to me to include the `gulp build-scripts` to the `prepublish` script in package.json.

`/public/styles/keystone.min.css` is committed also which can be taken out since it will be compiled on first run anyway

JedWatson:
@danielmahon I was planning to remove them once I got the on-init build process in place (see #957) but would be happy to do this first since it's causing you issues.

If you could submit a PR with the changes needed to `package.json` and `.npmignore` (is anything else needed?) I'd be glad to merge it, otherwise I'll get to it as soon as I can.

danielmahon:
@JedWatson Ill work on a PR today, one thing ive been struggling with is the "proper way" to build/deploy this on heroku once I take out app.js. If we use the `prepublish` script to build the npm module it's fine because the build file will be included in the published module. But for my current setup, my current app is pointing to a forked branch of keystone. This means I need to build on heroku, but it's tricky because my app is isnt building... keystone as a dependency needs to. Im just not sure how to go about doing it without adding all of keystone's build requirements as my app's dependencies, and then running `gulp --cwd ./node_modules/keystone build-scripts` in the `postinstall` script. Seems kinda hacky and I also dont think Im going about it correctly. Suggestions?

dcousens:
@jedwatson I think is worth looking at once we have extracted the admin UI from keystone core.

JedWatson:
Actually it's been done, I missed this issue when I was committing the fix. Thanks for the bump!

PR #1080, opened 2015-02-19
added gulp build-scripts task to npm postinstall

lojack:
solution to #955. Related to #957.


morenoh149:
to make this work the `chalk` module must be moved from `devdependencies` to regular `dependencies`

morenoh149:
also need module `del` moved. Almost there! :)

lojack:
I merged upstream and fixed the dependencies. I also think I figured out why I wasn't running into any of these issues with the devDependencies needing to be under dependencies in the first place.

morenoh149:
@lojack I've abandoned the postinstall strategy. I built app.js locally again. And this time the bug didn't go away. I'm now focusing my efforts on narrowing the source of the bug.

Do you still think building app.js on each client is the solution? I don't think so anymore. Thanks for the effort though :) See latest comment on https://github.com/keystonejs/keystone/issues/1038 where I shared a link to a live keystone app that experiences the bug. Also app.js in not uglified so we can debug it more easily.

lojack:
@morenoh149 I don't think this will solve #1038, but I do think it's probably still a good idea to build on postinstall so the minified app.js isn't included in the repo. It would help make commits cleaner, and prevent a lot of merge conflicts.

creynders:
:+1: this really needs to be done. 

1. The git history is polluted with "Updating build" commits
1. When working on any feature which requires rebuilding the scripts you have to be very careful not to commit the scripts as well or massive merge conflict mayhem ensues
1. Adding this to postinstall will enable a very crude sanity check when travis rebuilds the scripts after each merge
1. Having build results in git repos is simply not maintainable

morenoh149:
@creynders I agree the build itself shouldn't be under version control. @JedWatson has expressed the same.

However this pull request does not work in its current state. And the preferred strategy would be to rebuild the bundle based upon the fieldtypes the developer selects for his project. That's here https://github.com/keystonejs/keystone/issues/957 Closing this. Though thanks @lojack for the attempt :)

PR #1095, opened 2015-02-22
Add frame guard to zh doc

weekwood:


JedWatson:
:+1: Thanks @weekwood!

PR #1100, opened 2015-02-22
Removing unnecessary debug

JohnnyEstilles:
I thought we had introduced the use of `debug()` on `0.2.x` ... I was wrong. ... Sorry about that! :-(


JedWatson:
No problem!

PR #1113, opened 2015-02-25
change from yandex to airbnb

weekwood:
keystone follow airbnb code style, right?

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,6 +1,6 @@
 {
   "excludeFiles": ["coverage", "fields", "admin", "docs", "test", "node_modules", "public/build", "public/js/lib", "public/js/build"],
-	"preset": "yandex",
+	"preset": "airbnb",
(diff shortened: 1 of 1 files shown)


sebmck:
> keystone follow airbnb code style, right?

Not in the slightest.

JedWatson:
@weekwood it's not that simple, there was a lot of prior conversation behind this and a lot of (invalid) errors are shown when airbnb is selected because we don't follow it entirely. The config needs to be changed quite a bit from the preset to match our standards.

We had a linter that worked until changes to jshint and the introduction of jsx, now it's a mess. I'm currently thinking our best option is to switch to babel's linter (which supports JSX and ES6) with the appropriate config set up for Keystone.

JedWatson:
(nudge @sebmck? :)

weekwood:
ha @sebmck counting on you :)

PR #1121, opened 2015-02-26
allow ability for lists to inherit from each other

lojack:
This is just a proof of concept to add the ability to create lists that inherit from other lists. This doesn't currently allow field inheritance, but if this is something worth pursuing I would like to build this in. My branch currently is only useful for querying content. To accomplish this, you just need to do the following:

```javascript
var keystone = require('keystone');

var BasePage = new keystone.List('BasePage', {
    map: { name: 'title' },
    autokey: { path: 'slug', from: 'title', unique: true },
});
BasePage.add(
    {
        title: { type: String, required: true },
        slug: { type: String, readonly: true },
    }
);
BasePage.register();

var ChildPage = new keystone.List('ChildPage', {
    map: { name: 'title' },
    autokey: { path: 'slug', from: 'title', unique: true },
});
ChildPage.add(
    {
        title: { type: String, required: true },
        slug: { type: String, readonly: true },
        child_content: { type: String, readonly: true }.
    }
);
ChildPage.register(BasePage);
```

Then you could query the lists by doing:
```javascript
keystone.list('BasePage').model.find()
```
to find all models, or:
```javascript
keystone.list('ChildPage').model.find()
```
to find just the child models.

Ideally, It would be nice to be able to allow inheritance of fields and schema like the following:

```javascript
var keystone = require('keystone');

var BasePage = new keystone.List('BasePage', {
    map: { name: 'title' },
    autokey: { path: 'slug', from: 'title', unique: true },
});
BasePage.add(
    {
        title: { type: String, required: true },
        slug: { type: String, readonly: true },
    }
);
BasePage.register();

var ChildPage = new keystone.List('ChildPage', { inherits: 'BasePage' });
ChildPage.add({ child_content: { type: String, readonly: true } });
ChildPage.register();
```


JedWatson:
Really interesting idea. I'll have to have a think about how it would play out holistically but it looks like a neat feature to me. Will have to look further into what `baseList.model.discriminator(this.key, this.schema)` does...

I have a significant preference for your second proposed syntax:

`var ChildPage = new keystone.List('ChildPage', { inherits: 'BasePage' });`

... much more in-line with how inheritance is usually handled syntactically. `ChildPage.register(BasePage);` is weird and clunky for me.

lojack:
Yeah, the first syntax is primarily as a proof of concept to show that it will actually work as expected. I'm also 100% in agreement with the second proposed syntax.

I *believe* that model.descriminator() creates a field in Mongo which tells Mongoose what type of Schema it's dealing with. I can dig in to verify this. You can view this [in their documentation](http://mongoosejs.com/docs/3.7.x/docs/api.html#model_Model.discriminator). Their docs also describes how to go about having the actual Schemas inherit from each other. I don't believe this will be possible since Keystone wraps the schema types in its own fields. 

As you can see, being able to query multiple list types using this inheritance is a pretty simple task. Having the actual fields inherit from their parent appears to be a more significant chunk of work. Off the top of my head one way to accomplish this would be with a flag marking the base class as abstract, like so:

```javascript
var BasePage = new keystone.List('BasePage', {
    map: { name: 'title' },
    autokey: { path: 'slug', from: 'title', unique: true },
    abstract: true,
});
```

Then in `List.prototype.add()` we could store the fields in an object named something like `abstract_fields`. In the constructor for Child content we would then call List.prototype.add() on the Parent's `abstract_fields`. Not sure on the implications of this and whether or not its a good idea, but its one way of accomplishing field inheritance.

JohnnyEstilles:
I think this is an awesome idea. I've used Mongoose discriminators on other projects and it works great. I hope we can use this to automatically inherit Keystone fields (along with all the associated meta data). @lojack / @JedWatson ... let me if you would like some assistance with this.

webteckie:
+1 for this!

lojack:
So, I was digging through how Mongoose handles discriminators, and it appears that they don't allow nested inheritance. This makes sense as it simplifies things since the discriminator key is stored in a single field. This is a limitation we'll need to keep in consideration regardless of the implementation. 

With this known, does it make sense to have an API closer to the following:

```javascript
var keystone = require('keystone');

var BasePage = new keystone.AbstractList('BasePage', {
    map: { name: 'title' },
    autokey: { path: 'slug', from: 'title', unique: true },
});
BasePage.add(
    {
        title: { type: String, required: true },
        slug: { type: String, readonly: true },
    }
);
BasePage.register();

var ChildPage = new keystone.List('ChildPage', { inherits: 'BasePage' });
ChildPage.add({ child_content: { type: String, readonly: true } });
ChildPage.register();
```

Key being that the BasePage is now of type `AbstractList`. The idea behind this would be that we want the ability to query AbstractList's, but don't necessarily want to treat it the same as a List. First, we probably don't want to see them in the Admin UI. Second, they are the only Object that can be inherited from.

JedWatson:
@lojack I am all for this if you can get that syntax up and running.

lojack:
Just pushed a change that allows fields and options to be inherited from parents. One caveat I've found is that we cannot depend on the list registry, since child elements depend on their parents to be defined. That means that parent lists must be exported and required by their children. The API is as follows:

```javascript

var keystone = require('keystone');

var BasePage = new keystone.List('BasePage', {
    map: { name: 'title' },
    autokey: { path: 'slug', from: 'title', unique: true },
});
BasePage.add(
    {
        title: { type: String, required: true },
        slug: { type: String, readonly: true },
    }
);
BasePage.register();

var ChildPage = new keystone.List('ChildPage', { inherits: BasePage });
ChildPage.add({ child_content: { type: String, readonly: true } });
ChildPage.register();

```

I'm open to feedback on this approach. I also changed the behavior of the autokey option. Prior to this change, `schemaPlugins.autokey()` was modifying the value of `List.options.autokey` since objects are pass-by-reference. I suspect that this wasn't supposed to be occurring.

I haven't tested any of this on relationships or underscore methods.

JedWatson:
@lojack this looks good but I'm going to wait until the next release to merge it - getting very close to 0.3.3

webteckie:
@lojack does the ChildPage inherit any hooks defined in the BasePage as well?

lojack:
@webteckie These are defined on the mongoose schema, and I haven't overwritten any of that behavior. Based on what I know about how discriminators work in mongoose, they should be inherited. That said, I haven't personally tested this behavior.

creynders:
@JedWatson your last comment was to include this in 0.3.3, what's your take on this now, since it apparently didn't make the cut?

JedWatson:
Sorry for the confusion, I meant that 0.3.3 was close and I didn't want to merge it before launching that for stability reasons. Going to merge it now. @lojack any chance we can get docs added?

lojack:
Yeah, I'll put together some docs.

JedWatson:
:+1: thanks!

SharadKumar:
Not sure if tracking is addressed, I get:
```
throw new Error('Invalid List "track" option for ' + list.key + '\n' +
			      ^
Error: Invalid List "track" option for ChildPage
"createdAt" is already defined in the Schema.
    at /Users/sharadk/Documents/auctionomy-consign/node_modules/keystone/lib/schemaPlugins/track.js:62:10
```
when:
```
var BasePage = new keystone.List('BasePage', {
	map: {
		name: 'title'
	},
	autokey: {
		path: 'slug',
		from: 'title',
		unique: true
	},
	track: true
});
```

Issue #1118, opened 2015-02-26
upgrade to 0.3.1 Openshift 


tonila:
After upgrading, app does not have permission to execute. 

I'm getting the following:
    
```
==> app-root/logs/nodejs.log <==
Error: listen EACCES
    at errnoException (net.js:901:11)
    at Server._listen2 (net.js:1020:19)
    at listen (net.js:1061:10)
    at Server.listen (net.js:1135:5)
    at EventEmitter.app.listen (/var/lib/openshift/54ee09d0e0b8cdb22b000148/app-root/runtime/repo/node_modules/keystone/node_modules/express/lib/application.js:559:24)
    at Object.events.onMount (/var/lib/openshift/54ee09d0e0b8cdb22b000148/app-root/runtime/repo/node_modules/keystone/lib/core/start.js:128:30)
    at mounted (/var/lib/openshift/54ee09d0e0b8cdb22b000148/app-root/runtime/repo/node_modules/keystone/lib/core/mount.js:607:29)
    at /var/lib/openshift/54ee09d0e0b8cdb22b000148/app-root/runtime/repo/node_modules/keystone/lib/updates.js:180:15
    at /var/lib/openshift/54ee09d0e0b8cdb22b000148/app-root/runtime/repo/node_modules/async/lib/async.js:157:25
    at Promise.<anonymous> (/var/lib/openshift/54ee09d0e0b8cdb22b000148/app-root/runtime/repo/node_modules/keystone/lib/updates.js:120:5)
    at Promise.<anonymous> (/var/lib/openshift/54ee09d0e0b8cdb22b000148/app-root/runtime/repo/node_modules/keystone/node_modules/mongoose/node_modules/mpromise/lib/promise.js:177:8)
    at Promise.EventEmitter.emit (events.js:95:17)
    at Promise.emit (/var/lib/openshift/54ee09d0e0b8cdb22b000148/app-root/runtime/repo/node_modules/keystone/node_modules/mongoose/node_modules/mpromise/lib/promise.js:84:38)
    at Promise.fulfill (/var/lib/openshift/54ee09d0e0b8cdb22b000148/app-root/runtime/repo/node_modules/keystone/node_modules/mongoose/node_modules/mpromise/lib/promise.js:97:20)
    at /var/lib/openshift/54ee09d0e0b8cdb22b000148/app-root/runtime/repo/node_modules/keystone/node_modules/mongoose/lib/query.js:1400:13
[...]

PR #1127, opened 2015-02-28
Fixing issue when deploying Keystone app to OpenShift

JohnnyEstilles:
Fixing issue #1118.

OpenShift requires that the `host/ip` argument be explicitly passed when calling `app.listen()`. When available, Keystone uses `process.env.OPENSHIFT_NODEJS_IP` to set the `host` option. This PR passes the value of the `host` option to `app.listen()` ONLY when it is set, effectively correcting the issue when deploying to OpenShift, while maintaining backwards compatibility.

### DIFF
--- a/lib/core/start.js
+++ b/lib/core/start.js
@@ -125,7 +125,11 @@ function start(events) {
 				};
 			};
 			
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @JohnnyEstilles!

Issue #1128, opened 2015-02-28
noedit true on a Types.Relationship field shows as not set in admin ui


AndersonTaylor:
```
foo: { type: Types.Relationship, ref: 'Foo', noedit: true }
```
	
is set in a pre save but shows in admin UI as 

![screen shot 2015-02-28 at 12 07 08 pm](https://cloud.githubusercontent.com/assets/1709939/6427306/5d1966f4-bf42-11e4-89b5-3b25d5430ab9.png)

Removing noedit:true  verifies that it is in fact set.
It used to show up as a nice link:

![screen shot 2015-02-28 at 12 11 51 pm](https://cloud.githubusercontent.com/assets/1709939/6427334/0fa7a7cc-bf43-11e4-8550-6093c5f109b6.png)

PR #1134, opened 2015-03-02
Temporary fix for linking non-editable relationship fields with multiple...

creynders:
... values to their respective resources.

ATM if you set a relationship field to `{noedit:true, many:true}` it is rendered as a sequence of `<span>`s, i.e. they're all shown inline, which is just horrible. Also, no links to the respective resources are created.

As a temporary fix I uncommented the relevant lines starting from https://github.com/keystonejs/keystone/blob/master/fields/types/relationship/RelationshipField.js#L151 (and also fixed some errors), but saw that this would revert commit e5905fbcf2277f29406c53dd5e2fa5c4142c3c57. @jossmac do you remember what the error was, and why you chose for this solution?? TBH I think the only thing wrong with the previous code (i.e. the one you commented) was that it returned an undeclared `value` variable instead of `body` which it should have.

Just to be clear: this PR **DEFINITELY** needs to be reviewed, I really don't know the first thing about how the fields are rendered, my changes are a result of trial-and-error in an existing project, but I have NO IDEA of its impact on the relationship field at large.

### DIFF
--- a/fields/types/relationship/RelationshipField.js
+++ b/fields/types/relationship/RelationshipField.js
@@ -143,19 +143,15 @@ module.exports = Field.create({
 		if (!this.state.ready) {
 			return this.renderLoadingUI();
 		}
-		// TODO expand IDs
-		if (this.props.many) {
-			// a(href='/keystone/' + refList.path + '/' + item.get(field.path), data-ref-path=refList.path).ui-related-item= item.get(field.path)
-			return <div className='field-value'>{this.props.value}</div>;
-		} else if (this.props.many && this.props.value.length) {
-			// var body = [];
-			// 
-			// _.each(this.props.value, function (value) {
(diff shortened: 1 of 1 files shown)


creynders:
Just saw #1128 and realized I could fix it at once as well, hence the added commit.

JedWatson:
@creynders good PR, all works as intended. I tweaked the styling a little too, so the min-width is gone now.

gonzalodiaz:
Exactly what I needed! Is there a planned release soon for this fix?

morenoh149:
@gonzalodiaz should be in the next release since it's merged. I'd wager within a week for 0.3.3

gonzalodiaz:
@morenoh149 0.3.3 contains your fix and it is working great. Thanks!

PR #1141, opened 2015-03-03
Localfile enhancements

creynders:
A few minor tweaks to the localfile field type. With this the original filename is automatically saved to the DB as well. This is necessary to allow more advanced custom collision resolutions. Also, the full file object is passed to the `options.filename` function, in symmetry with `options.format` and to allow more complex file name generators. I updated the docs at once as well.
If anything else needs to be done, let me know.


JedWatson:
Looks good to me.

@sebmck is leading an effort to consolidate our file fields in the `file-fields` branch, should these changed be integrated into that also?

PR #1143, opened 2015-03-03
Added DateArray field

liamwooding:
DateArray works much the same as NumberArray and TextArray, but with a Pikaday datepicker being attached to each field in the array.


morenoh149:
Good pull request. But I wonder - wouldn't it be better to remove the dependency on pikaday and use https://jquense.github.io/react-widgets/docs/#/datetime-picker I think it'd be wise to make this move now before adding support for an Array of them.

liamwooding:
Ah good idea, I'm fairly new to React and I hadn't clocked that it had its own datepicker. It feels like a matter for another pull request to me though - as it is, we now have the backend logic and tests which lay the groundwork for when someone wants to make that change.

JedWatson:
:+1+ I agree, @liamwooding.

I've also been looking at [material-ui](http://material-ui.com/) which is a beautiful framework - I love their date picker implementation.

I can't see us actually using it in Keystone because IMO the style of material design is very specific (Google's design language) and I want Keystone to remain more independent than that - we should have our own style that's simple, beautiful but not opinionated or strongly aligned with any one platform or existing design language. Also I suspect Material will become a bit like Bootstrap with its ubiquity.

However I'm interested in seeing what we can borrow from them. One for the roadmap.

PR #1147, opened 2015-03-04
Allow asynchronous callbacks for `value` methods in fields

creynders:
If I declare a field to watch a relationship field it only receives the `id` of the selected relationship, however in most cases you'd want to use the `name` field or another field value. Unfortunately since the  result of the `value` option is used synchronously you can't use the passed in id to look up the relevant document either.

E.g.:

```js
//file: models/representations.js
var Representation = new keystone.List( "Representation", {
  track : true
} );
Representation.add({
  name : {
    type     : String,
    watch    : "assessment",
    value    : function(){
        //I'd like to do this, but this.assessment is an id
        return this.assessment.name + " - " + Date.now();
    }
  },

  assessment : {
    type     : Types.Relationship,
    ref      : "Assessment"
  }
});
```

Not sure what the best solution would be. One possibility would be to modify [this line](https://github.com/keystonejs/keystone/blob/2ffbbbfefafbd9b92b0e9e5f93ea5d3d1ea95e2f/fields/types/Type.js#L205) to an asynchronous one, e.g.

```js
//file: fields/types/Type.js
field.options.value.call(this, function(value){
  this.set(field.path, value);
}.bind(this));
```

In which case the `value` field option is responsible for calling the callback:

```js
//file: models/representations.js
    value    : function(callback){
        //retrieve the assessment document using this.assessment
        callback(results);
    }
```

### DIFF
--- a/docs/content/en/pages/docs/database.jade
+++ b/docs/content/en/pages/docs/database.jade
@@ -498,15 +498,22 @@ block content
 				tr
 					td <code>value</code> <code class="data-type">Function</code>
 					td
-						p The function to generate the field value when a watched path is changed. Must return the new value.
+						p The function to generate the field value when a watched path is changed. Must return the new value, or accept a node-style <code>callback</code> argument, which can be called to set the field value asynchronously.
 						p The <code>this</code> context of the function will be the item being saved.
-						.code-header: h4 Example
+						.code-header: h4 Example (synchronous)
 						pre: code.language-javascript
 							| function () {
 							|     return this.total<=this.totalreceived ? true:false;
 							| }
-			
-			//- TODO: Documentation for field methods
+						.code-header: h4 Example (asynchronous)
+						pre: code.language-javascript
+							| function (callback) { // BEWARE: MUST be called "callback" to allow asynchronous execution
+							|	list.model.findById(this.createdBy).exec(function(err, user){
+							|		callback(err, user.name + "-" + Date.now());
+							|	});
+							| }
+
+	//- TODO: Documentation for field methods
 			
 			a(name="fields-underscoremethods")
 			h3 Underscore Methods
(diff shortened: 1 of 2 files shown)


creynders:
Turned the issue into a PR implementing the above approach.

JedWatson:
I think this would be a perfect opportunity for our `asyncdi` functionality.

It detects whether a function is async or not based on dependency injection techniques, and provides a consistent interface to it.

see `/lib/asyncdi.js` and `/test/lib/asyncdi_test.js` for usage.

creynders:
@JedWatson ah, cool. I'll check that out and use it instead.

creynders:
@JedWatson You were right, using asyncdi proved a MUCH cleaner solution.
Tested it manually in two dimensions: synced/asynced, result/error. All 4 worked as they should.

AFAICT this is ready to land. 
What's the strategy regarding docs translations? Do I add a label here, or open a new issue once this is merged?

creynders:
Oh, BTW I applied non-fatal error handling, i.e. if for instance a developer incorrectly calls the callback with the results as a first argument it will simply output to console but won't blow up. Hope that's ok?

E.g.

```js 
    value    : function(callback){
        keystone.list("Assessment").model
          .findById(this.assessment)
          .exec(function(err,assessment){
            //loser, `err` should go first!
            callback( assessment.name + " - " + Date.now() );
          });
    }
```
Will output something like:
```sh
Error: Watch set with value method for Representation.name (text) throws error: An assessment-20150306104132
```

JedWatson:
> AFAICT this is ready to land.

Same.

re: non-fatal error handling, for now at least I agree. I feel like this sort of question should have an overarching philosophy in Keystone to be sure we explain and are consistent with it but that's something I'm happy to come back to defining down the track.

PR #1153, opened 2015-03-05
Asyncdi minor tweaks and fixes

creynders:
ATM the context parameter is not passed to Wrapper from the module function in asyncdi.
For instance: you'd expect the callback to be called inside `context`, but it isn't.

```js
var di = require('./lib/asyncdi');
di(fn_scope, scope).call(function(err, val){
	demand(val).equal(scope); //fails
	done();
});
```

Fix and tests coming in a few secs.

### DIFF
--- a/lib/asyncdi.js
+++ b/lib/asyncdi.js
@@ -16,8 +16,8 @@ var FN_ARGS = /^function\s*[^\(]*\(\s*([^\)]*)\)/m;
  * @param  {Function} fn
  * @return {Wrapper}
  */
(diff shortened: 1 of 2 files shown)


creynders:
As discussed with @JedWatson :

```js
            di(fn_sync_error).call(function(err, val){
                demand(err).equal(thrownErr)
            });
```
Now works as expected, i.e. if a synchronous function throws an error it's caught and passed as the first argument to the callback.

JedWatson:
Nice work @creynders, thanks for cleaning this up.

Issue #1140, opened 2015-03-03
Unable to upload images, tinymce not rendering


stephencoe:
There appears to be a change since 0.3.1 that is stopping tinymce working when images are enabled

tinymce is trying to load ```/keystone/js/lib/tinymce/plugins/uploadimage/plugin.min.js``` which doesn't exist in the structure (404), it looks like it can be found in ```/keystone/js/lib/tinymce/plugins/image/plugin.min.js```

Checking through the file the HtmlField type refers to [uploadimage](https://github.com/keystonejs/keystone/blob/master/fields/types/html/HtmlField.js#L79) but looks like it should be "image"

I manually changed it locally to test if it would fix but even after changing the upload dialog only shows a source field without any upload functionality

awinecki:
+1 confirmed here.

It seems the whole uploadimage/ plugin dir has been removed. https://github.com/keystonejs/keystone/commit/dbb1594eb2f63709f9f582b74fc3a69dbd304ef5

pswoodworth:
+1 

having the same experience trying to use `wysiwyg cloudinary images` for tinyMCE on 0.3.2.

Interestingly, I have an earlier local install of 0.3.2 that does have `/keystone/js/lib/tinymce/plugins/uploadimage/plugin.min.js` but it is not showing up in a clean install now – It's possible that it's leftover from a previous version.

morenoh149:
a temporary hack would be to host the images elsewhere and add that manually into tinymce via the `code` button.

pswoodworth:
I've submitted a pull request that adds the plugin back in. This resolves the issue. If you'd like to use my fork with the addition while the pull request is pending you can find it here: https://github.com/pswoodworth/keystone

JedWatson:
Cheers @pswoodworth, appreciate the PR! merged now, will be released in the next version (soon!)

pswoodworth:
@JedWatson you got it – Keystone has really been a fantastic platform for us. Thanks for all your work – I'd write the heck out of a testimonial if you ever want one!

JedWatson:
:D

ehoch:
@JedWatson Looks like this issue is back with 4.0.0. Can we add this TinyMCE plugin back to master or is there another preferred way to do image uploading in the new version?

svpe:
[...]

PR #1155, opened 2015-03-05
added uploadimage back into tinymce plugins

pswoodworth:
Resolves issue [#1140](https://github.com/keystonejs/keystone/issues/1140) by adding the `uploadimage` plugin back to tinymce plugins at `public/js/lib/tinymce/plugins/uploadimage`.


JedWatson:
Thanks @pswoodworth

PR #1164, opened 2015-03-06
Fix for validationErrors beeing overwritten

douglasf:
When a validation error is added with UpdateHandler.prototype.addValidationError it will be overwritten here https://github.com/keystonejs/keystone/blob/master/lib/updateHandler.js#L274 unless we check if it's already in validationErrors[field.path]

Hope I did not oversee anything

### DIFF
--- a/lib/updateHandler.js
+++ b/lib/updateHandler.js
@@ -136,11 +136,10 @@ UpdateHandler.prototype.process = function(data, options, callback) {
 	}, this);
 	
 	// TODO: The whole progress queue management code could be a lot neater...
-	
 	var actionQueue = [],
 		addValidationError = this.addValidationError.bind(this),
 		validationErrors = this.validationErrors;
-	
+
 	var progress = function(err) {
 		if (err) {
 			if (options.logErrors) {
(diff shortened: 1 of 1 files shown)


creynders:
AFAICT if an error was encountered already that code won't run due to `!invalidated` in the conditional, or am I missing something (both code wise and/or problem wise)? I.e. I don't think it ever gets overwritten?

douglasf:
Actually I might have misunderstood the use for addValidationError function, noticed just before leaving work. But I did something in line with 

```javascript
...
var updater = newComment.updateHandler(req);

updater.addValidationError(some_path, 'the message');

updater.process({ fields: 'comment', showErrors: true }, function() { ... });
```

Excuse the shit code, not on my workstation ;)
'the message' was however overwritten by 'Comment is required' when i submitted an empty comment. If this don't make any sense I can get more elaborate code on Monday

creynders:
Yes, only a single error is thrown for each field, so if the field-specific validator throws one, the custom one won't be shown.

JedWatson:
Good catch @douglasf. I hadn't tried using custom validators prior to the UpdateHandler and then having it update fields too so didn't notice this issue. Thanks for the fix!

PR #1167, opened 2015-03-07
[S3FileType] support custom headers

### LINKED ISSUE #1066: Support headers option on s3FileType
would be nice to enable an option on the Field for setting headers on the S3 Object. For example:
```javascript
MyList.add({
  // ...
  pdf: { 
    type: Types.S3File, 
    headers: { 
      'x-amz-meta-Cache-Control' : 'max-age=' + (60 * 15),
      'x-amz-meta-keyHeader' : valueHeader
    } 
  },
  // ...
});

// or

MyList.add({
  // ...
  pdf: { 
    type: Types.S3File, 
    headers: function (item, file){
      var headers = {};
      headers['x-amz-meta-Cache-Control'] = item.maxAge;
      headers['x-amz-meta-keyHeader'] = item.valueHeader;
      return headers;
    }
  },
  // ...
});
```

montmanu:
Adds support for an option on the Field to set headers on the S3 Object. For example:
```javascript
MyList.add({
  // ...
  pdf: { 
    type: Types.S3File, 
    headers: { 
      'x-amz-meta-Cache-Control' : 'max-age=' + (60 * 15),
      'x-amz-meta-X-Custom-Header' : 'Object Option'
    } 
  },
  // ...
});

// or

MyList.add({
  // ...
  pdf: { 
    type: Types.S3File, 
    headers: [
      { name: 'x-amz-meta-Cache-Control', value: 'max-age=' + (60 * 15) },
      { name: 'x-amz-meta-X-Custom-Header', value: 'Array Option' }
    ]
  },
  // ...
});

// or 

MyList.add({
  // ...
  pdf: { 
    type: Types.S3File, 
    headers: function (item, file){
      var headers = {};
      headers['x-amz-meta-Cache-Control'] = 'max-age=' + item.maxAge;
      headers['x-amz-meta-X-Custom-Header'] = 'Computed Option (Object)';
      return headers;
    }
  },
  // ...
});

// or 

MyList.add({
  // ...
  pdf: { 
    type: Types.S3File, 
    headers: function (item, file){
      var headers = [];
      headers.push({ name: 'x-amz-meta-Cache-Control', value: 'max-age=' + item.maxAge });
      headers.push({ name: 'x-amz-meta-X-Custom-Header', value: 'Computed Option (Array)' });
      return headers;
    }
  },
  // ...
});
```

Also adds support for setting default headers in the 's3 config' object. For example: 
```javascript
keystone.set('s3 config', _.extend({}, keystone.get('s3 config'), {
  'default headers': {
    'x-amz-meta-X-Default-Header': 'Custom Default Value (Object)'
  }
}));

// or

keystone.set('s3 config', _.extend({}, keystone.get('s3 config'), {
  'default headers': [ 
    { name: 'x-amz-meta-X-Default-Header', value: 'Custom Default Value (Array)' } 
  ]
}));
```

Closes #1066


morenoh149:
is there anyway to narrow the possible number of ways to use the feature? Or is that just the flexibility offered due to the strategy?

montmanu:
Sure, I could restrict it anyway you guys see fit. I just figured I'd make it as flexible as possible to provide values for the option. Tried to combine aspects of SelectType's options.options syntax with S3FileType's options.filename syntax. 

Would it make more sense to restrict the option to only supporting an Object or only supporting an Array?

morenoh149:
nope. It's @JedWatson 's call. In fact I think this project prefers as much flexibility as possible.

JedWatson:
Really good stuff, @montmanu! I like the level of configuration and flexibility you've provided - suits several different use-cases...

One question - why would the headers for the field be stored in the database? they're not (as far as I can tell) customisable on a per-value basis, just for the field itself, and they're passed to S3 so there's no need for them to be stored with the rest of the file metadata.

Maybe I missed something, but otherwise we should remove `headers` from the schema.

I'll mark up the changes for clarity.

Other than that - good to merge!

fields/types/s3file/S3FileType.js
```diff
@@ -97,23 +97,25 @@ s3file.prototype.addToSchema = function() {
...
-		path:			String,
-		size:			Number,
-		filetype:		String,
-		url:			String
+		filename:   String,
+		path:     String,
+		size:     Number,
+		filetype:   String,
+		url:      String,
+		headers:      Object
```
JedWatson: This change isn't necessary as far as I can tell. They headers never come from user-submitted data so they won't change, and we don't need to keep them. (?)

JedWatson:
**edit**

I can see how headers could be different on a per-item basis, given that you provide `item` to the custom headers function and it returns dynamic headers.

_however_ I'd propose that, even in this case, the state of the data is knowable and it's not necessary to store the headers that were provided to S3 in the database.

If you disagree, we can leave them in, this isn't a use-case I have so you may be more familiar with why this is important.

In the case that dynamic headers are set when item values are changed (e.g. maxAge) we may have two things happening; race conditions (we need to test for this) and stale values in S3 (e.g. change a value and files have already been uploaded with headers that are now out of date and should be updated)

We'll have to cover the second carefully in documentation; I suspect it's the sort of thing that should be addressed with custom post-save hooks.

montmanu:
@JedWatson I included them in the schema to keep in line with the way other metadata about the file is stored (filetype, size, etc). Though, I agree.. the headers probably do not need to be stored in the database. I will remove them from the schema and include that commit in this pull request. 

Should I update the docs to include this new option for the S3File or do you guys have an existing process in place for managing documentation?

morenoh149:
@montmanu please do note it in the docs if you can. You should then make new issues for the docs to be translated See https://github.com/keystonejs/keystone/wiki/Documentation-Translation-Guidelines

sericaia:
Extremelly useful @montmanu . Thanks, I hope it is going to be merged soon.

JedWatson:
@montmanu looks good, but it combines two changes - would you please remove the change to the `reset` method for now? It's a potentially breaking change as it has side effects we should discuss in a new issue. Happy to merge the `headers

JedWatson:
Perfect :)

Issue #1156, opened 2015-03-05
Add to Date and DateTime types option for parse formats


ttsirkia:
In addition to `format`option, there should be an optional option to give a list of accepted parse formats as well. Currently, it seems that for DateTime there is a predefined list of formats and for Date everything that Moment.js parses is accepted.

This is quite an important issue because formats vary a lot. For example, here in Finland I would like to accept dates in format "5.3.2015".

ttsirkia:
I was already implementing this feature but there is a problem. It is trivial to support this if we forget the Admin UI but that is not a good option.

The problem is that `Field` should know the `parseFormat` option from the `Type` or otherwise the validation will fail in the server side.

ttsirkia:
Another option would be to include always the default formats and append the user given formats to that list. Any comments?

ttsirkia:
Implemented in ee26dbf, closing this.

PR #1173, opened 2015-03-07
Implements parseFormat and parseStrict options

ttsirkia:
This pull request adds two new options for `DateType` and `DatetimeType`. The new options are `parseFormat` and `parseStrict`.

With these options, user can define a custom input format for dates. If `parseStrict`is used, only the given format will be used, otherwise the default options are also used. If `parseFormat`is given without `parseStrict`, the given format will be tried before the default formats.

The format can be a comma-separated string or an array of strings.

Because the Admin UI doesn't know anything about these new options, `parseStrict` should be not be used if the Admin UI is in use. For this reason, `parseStrict` defaults to false.

Implements #1156

### DIFF
--- a/docs/content/en/pages/docs/database.jade
+++ b/docs/content/en/pages/docs/database.jade
@@ -666,7 +666,8 @@ block content
 			.options
 				
 				h5 Options
-				p <code>format</code> <code class="data-type">string</code> - the default format pattern to use, defaults to <code class="default-value">Do MMM YYYY</code>
+				p <code>format</code> <code class="data-type">string</code> - the default format pattern to show the date, defaults to <code class="default-value">Do MMM YYYY</code>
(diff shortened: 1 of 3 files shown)


ttsirkia:
It would be highly appriciated to get this accepted. I'd like to avoid this kind of lines before using `UpdateHandler`:

```javascript
req.body.arrivalTime = moment(req.body.arrivalDate, 'D.M.YYYY H:mm');
req.body.arrivalTime = req.body.arrivalTime.isValid() ? req.body.arrivalTime.format('YYYY-MM-DD H:mm') : null;
```

ttsirkia:
@JedWatson any thoughts?

ttsirkia:
Implemented in ee26dbf, closing this.

dcousens:
Thanks @ttsirkia,  could you may resubmit your documentation fixes and tag me?  There is still some useful code here :)

PR #1181, opened 2015-03-09
User impersonation

orthes:
First of all, thanks for a great CMS. I'm using it on a project where I as an administrator need to impersonate another users in order to see what they see and act on their behalf without needing their credentials.

**How it works:**

- I added a new option to lists: ```impersonation``` (boolean)

```javascript
var User = new keystone.List('User', {
	impersonation: true
});
```

- In order to remain access to the administration even when a user without admin rights is impersonated you need to modify ```canAccessKeystone``` virtual property:

```javascript
User.schema.virtual('canAccessKeystone').get(function() {
	return this.isImpersonated || this.isAdmin;
});
```

- When impersonation is enabled, a new button is added to a user edit form:

screen shot 2015-03-09 at 15 32 48

- After clicking the "Impersonate" button the selected user is signed in and a property ```user.isImpersonated``` is set to true, so we can still access the administration.

--

What do you think about this approach? Do you think it might be useful? If yes, I can update documentation and  yo generator.


morenoh149:
this doesn't look like a common usecase to me. Perhaps if it were fleshed out with more administrative usecases.

orthes:
What I was looking for initially, was a way to configure my own custom actions. Not to have only the default CRUD ones. For example to be able to have this "impersonate" action on users, "publish" on posts, and so on.

Since I couldn't find general option like this, I programmed the solution for my specific problem - impersonation.

I can close this PR and I will submit a new one if I come up with more general approach.

Issue #1182, opened 2015-03-09
Updated from 0.3.0 to 0.3.0 and got @select-menu-max-height is undefined


sseppola:
Here is the full error message:

  LESS Name error : variable @select-menu-max-height is undefined
  LESS File : /Users/Sindre/projects/wdBlog/node_modules/keystone/node_modules/react-select/less/select-menu.less 21:13
Name Error thrown for request: /keystone/styles/keystone.min.css
{ [Error: variable @select-menu-max-height is undefined]
  type: 'Name',
  filename: '/Users/Sindre/projects/wdBlog/node_modules/keystone/node_modules/react-select/less/select-menu.less',
  index: 732,
  line: 21,
  callLine: NaN,
  callExtract: undefined,
  column: 13,
  extract:
   [ '\tmargin-top: -1px;',
     '\tmax-height: @select-menu-max-height;',
     '\tposition: absolute;' ],
  message: 'variable @select-menu-max-height is undefined',
  stack: undefined }

sseppola:
Either it is not imported with defaults from keystone, or it is a react-select issue. It worked after adding a @select-menu-max-height variable at the top of the select-menu.less file, but the same variable is set in default.less

freder:
I also wasn't sure if it was a keystone or a react-select issue, so I opened [https://github.com/JedWatson/react-select/issues/87](https://github.com/JedWatson/react-select/issues/87).

esparragito:
Have found the root of the problem I think. The @select-menu-max-height which seems to be a new addition to react-select isn't defined in keystones public/styles/react/react/less file. Add it in there and everything should be grand.

The default.less file in the react-select module never gets loaded by keystone, so any variables in there never get seen.

porschuetz:
ah, just opened very similar issue. #1188 

@esparragito thanks. fixed it.

porschuetz:
what height did you set it to? I did 300 and now it is quite interesting.

![screen shot 2015-03-10 at 11 03 45](https://cloud.githubusercontent.com/assets/1228101/6572767/3bfa179a-c715-11e4-957a-a4b652ea2ec0.png)

esparragito:
I'm afraid I'm using less, and I haven't had any problems setting it to that value. Is SASS normally happy without units being defined?

JedWatson:
This shouldn't have anything to do with SASS, as the Admin UI stylesheets are all implemented with LESS...

Fixed by #1183

[...]

PR #1183, opened 2015-03-09
Add @select-menu-max-height variable to react.less

esparragito:
This is one solution to the error in [#1182]

### DIFF
--- a/public/styles/react/react.less
+++ b/public/styles/react/react.less
@@ -11,6 +11,7 @@
 @select-arrow-width:               5px;
 
 @select-menu-zindex:               @zindex-dropdown;
+@select-menu-max-height:           200px;
 
 @select-option-color:              lighten(@select-text-color, 20%);
 @select-option-focused-color:      @select-text-color;


tekwiz:
+1 This fixed the issue for us

JedWatson:
Thanks, really should have been a major version bump on react-select given that it needed a new less variable!

esparragito:
@JedWatson On the react-select side, is there any value in trying to get a module requiring react-select to use the default.less before overriding with defined variables, just to avoid the error being thrown in the event of future updates to the less structure?

JedWatson:
Yes, that would be good as long as the variables behaved like defaults when undefined (rather than overriding)

Haven't tested how to achieve that but it would be a great thing to look into.

esparragito:
Including default.less at the beginning rather than the end of the keystone file lets the variable definitions behave as overrides, but I don't know if that messes anything up elsewhere.

esparragito:
Maybe clearer: https://gist.github.com/esparragito/bfaa94ae5e0b30bea049

PR #1187, opened 2015-03-10
Fixes line numbers not showing up in the CodeMirror editor.

### LINKED ISSUE #1175: CodeMirror line numbers not visible
For some reason, the line numbers have disappeared from CodeMirror.

It doesn't appear to be an issue with the configuration object - setting the default to `false` removes the gutter completely.

I've taken a look and there's nothing obvious about what's going on, so it'll need a more in-depth investigation.

webteckie:
Fixes #1175


JedWatson:
Thanks @webteckie... makes me wonder why that hack was in there in the first place :/

PR #1190, opened 2015-03-10
Use filename without suffix as default publicID for cloudinary

aschwersenz:
This is a continuation from #925 as I made the changes for 0.3 now.

There are two options you can set:

_ 'wysiwyg cloudinary images filenameAsPublicID': This will change the publicID to the filename if you use the cloudinary image upload within the wysiwyg editor.

_ filenameAsPublicID for a cloudinary image field: This will change the publicID to the filename if set to true for a cloudinary image field. The publicID option takes precedence for backwards compatibility.


JedWatson:
:+1: looks good. Will need docs, I'll open a separate issue so we can get this merged but if you could add them I'd appreciate it.

PR #1193, opened 2015-03-11
Adding node 0.12 to travis builds

JohnnyEstilles:
@JedWatson, I know this may seem like a ridiculously silly PR, but would it be okay if we configure Travis to build/check Keystone on Node `0.12` as well?


JedWatson:
Sure!

PR #1205, opened 2015-03-13
Refactor "doSignin()" and expose as "keystone.session.signinWithUser()"

JohnnyEstilles:
Factored `doSignin()` out of `keystone.session.signin()`. This allows for reusability (in support features like the ones mentioned in keystonejs/generator-keystone#10)  and better testability.

Renamed and exposed the new method as `keystone.session.signinWithUser()`.

Tests for <del>`keystone.session.doSignin()`</del> `keystone.session.signinWithUser()` already included. I intend to add the rest of the tests for the `keystone.session` module as time permits.

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -30,24 +30,19 @@ var hash = function(str) {
 
 exports.signinWithUser = function signinWithUser(user, req, res, onSuccess) {
 	if (arguments.length < 4) {
-		console.error('\nkeystone.sesson.signinWithUser requires user, req and res objects, and an onSuccess callback.\n');
-		process.exit(1);
+		throw new Error('keystone.sesson.signinWithUser requires user, req and res objects, and an onSuccess callback.');
 	}
 	if ('object' !== typeof user) {
-		console.error('\nkeystone.sesson.signinWithUser requires user to be an object.\n');
-		process.exit(1);
+		throw new Error('keystone.sesson.signinWithUser requires user to be an object.');
 	}
 	if ('object' !== typeof req) {
-		console.error('\nkeystone.sesson.signinWithUser requires req to be an object.\n');
-		process.exit(1);
+		throw new Error('keystone.sesson.signinWithUser requires req to be an object.');
 	}
 	if ('object' !== typeof res) {
-		console.error('\nkeystone.sesson.signinWithUser requires res to be an object.\n');
-		process.exit(1);
+		throw new Error('keystone.sesson.signinWithUser requires res to be an object.');
 	}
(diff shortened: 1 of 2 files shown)


creynders:
Though this is great ( :+1: ) I'm always extremely wary of methods called `action` and `doAction` (unless the former is a wrapper for a protected latter) I think it should be renamed.
If I start out with keystone development and I see both those methods I have no idea which one I should use and when.

JohnnyEstilles:
@creynders ... you make a lot of sense. Technically, `signin()` is a wrapper for `doSignin()`. It searches for the `User` (by e-mail or ObjectId), then calls `doSignin()` with the `User` object, which in turn sets up the session.

How about if:

- We leave `keystone.session.signin()` as is (so this doesn't become a breaking change).
- I rename `doSignin()` to something like ... `signinWithUser()`?

`keystone.session.signin()` isn't yet documented in the [Keystone API](https://github.com/keystonejs/keystone/wiki/Keystone-API) (or anywhere else AFAICT). So, as long as we document them well our consumers should know how to use each API.

JohnnyEstilles:
@creynders ... I went ahead and renamed the new API `signinWithUser()`, so as to avoid the confusion you outlined in your comment (which I really appreciate ... BTW).

@JedWatson, any comments or suggestions on your end?

lib/session.js
```diff
@@ -20,6 +20,49 @@ var hash = function(str) {
...
+ * @param {Object} user - user object
+ * @param {Object} req - express request object
+ * @param {Object} res - express response object
+ * @param {function()} onSuccess callback, is passed the User instance
+ */
+
+exports.signinWithUser = function signinWithUser(user, req, res, onSuccess) {
+	if (arguments.length < 4) {
+		console.error('\nkeystone.sesson.signinWithUser requires user, req and res objects, and an onSuccess callback.\n');
+		process.exit(1);
```
JedWatson: process.exit is really unsafe, and only something I'm comfortable using in keystone config validation - not code that would be run in the normal context of application execution.

There's no reason to kill the process here, it can just throw an error.

JohnnyEstilles: Got it ... updating ... :-)

JedWatson:
@JohnnyEstilles looks good, except for your use of `process.exit`. Replace that with a `throw` and we're good to go.

Also :+1: for adding tests!

JohnnyEstilles:
@JedWatson updated! :-)

JedWatson:
Great!

darafsheh:
Does anyone has an example implementation of this?

Issue #1152, opened 2015-03-05
Relationship fields should be cleaned up when documents are deleted


liamwooding:
If I link some Student documents to a Class using a relationship field, then delete some Students, the relationship field on the Class will still contain their IDs. My feeling is that they should be cleaned up, but this behaviour might not be to everyone's taste, so it could be a field option on Types.Relationship.

It would also be nice to be warned before deleting a document if its ID is found in any relationship fields.

ttsirkia:
Another option (not in this context but generally) would be to delete the linked students if the class is removed. This can be achieved by using pre-save but an option would be much nicer.

creynders:
:+1: for this. It would be very useful in some cases, but I'd prefer the option presented in the admin UI, i.e. you delete a student and you need to choose whether you want all references to them deleted as well.

creynders:
Just another thing, to avoid high search times, it could exclusively search the relationships if they're explicitly set using [relationship definitions](http://keystonejs.com/docs/database/#relationship-definitions).

creynders:
I think I'm going to tackle this one. But I'm going to make it a bit more flexible, i.e. to allow more use cases. My goal is to set up relationship watchers which allow you to react to the removal of their targets, something like this, suppose we got a `UserModel` and `PostModel` with an `author` field which references `UserModel`.

```js
var Post = new keystone.List('Post', {
	map: { name: 'title' },
	autokey: { path: 'slug', from: 'title', unique: true },
	watch:{ 
		author: {
			onRemoveTarget: function(target){
				this.remove();
			}
		}
	}
});
```

This would delete a post if its author is deleted. But if you'd only want to set the reference to null instead, you could do:

```js
onRemoveTarget: function(target){
	this.author=null;
	this.save();
}
```

Or if you prefer non-destructive removals:
```js
onRemoveTarget: function(target){
	this.deleted=true;
	this.save();
}
```

@liamwooding any thoughts?

liamwooding:
@creynders watchers for relationships sounds like a great idea, gives people a lot of options. Go for it!

ttsirkia:
To make this easier to use, I'd also suggest an option in fields to use this kind of default functionality.

ttsirkia:
However, this kind of watchers are great idea!

liamwooding:
Agreed, maybe there could be a field option which is a reference to a handler function:
[...]

PR #1213, opened 2015-03-14
Extract pre & post hooking functionality to separate module

creynders:
I saw the 'pre' and 'post' hook functionality was re-implemented at several places, so I extracted it to a separate module. This is related to #1152 since I'll need some kind of mechanism to hook into the application life cycle, i.e. I'll be adding some more `pre` and `post` dispatching to keystone.

Some remarks:

- I use a mixin mechanism to add the methods to an existing instance, if you prefer a different approach @JedWatson let me know
- It would be best to extract the `prepost` module to an external module IMO, as you did with `asyncdi`

### DIFF
--- a/fields/types/azurefile/AzureFileType.js
+++ b/fields/types/azurefile/AzureFileType.js
@@ -5,10 +5,10 @@
 var _ = require('underscore'),
 	moment = require('moment'),
 	keystone = require('../../../'),
-	async = require('async'),
 	util = require('util'),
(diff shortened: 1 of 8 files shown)


JedWatson:
:+1: looks good @creynders! agreed re: external module, let's stage it like this for now but that sounds like a good idea.

creynders:
Cool! I'll extract it to an external module then and add some tests and docs.

PR #1214, opened 2015-03-14
Add format option to Url type to prevent stripping http/https PR/Question

danielzurawski:
Hi,

This PR is just a question/suggestion. If you think that this is a sensible patch, I'll add the docs/integrate any other suggestions. 

It seems that the column_link mixin for Url type inside columns.jade always executes ```url.prototype.format```
method for the anchor href and preview, which strips the http & https prefix from the url, causing columns with Url type to be relative to the site. 

```
var value = col.field ? col.field.format(item) : item.get(col.path)
...
else if col.type == 'url'
    +column_link(value, value, true)
...
```

E.g. if I want to have a column with youtube urls, pressing on a record on the admin UI will point the user back to ```http://localhost:3000/keystone/img.youtube.com/vi/s2nlrpudHIM/hqdefault.jpg``` instead of ```https://img.youtube.com/vi/s2nlrpudHIM/hqdefault.jpg```

To overcome that I've added option to provide custom format for the Url, so that in your model you can do this:
```
thumb: { type: Types.Url, initial: false,
             watch: 'video',
             format: function(url) { return url; },
             value: function() { return getThumbnailUrl(this.video)} }
```

What do you think? Is there an alternative to doing this?

Thanks.

### DIFF
--- a/docs/content/en/pages/docs/database.jade
+++ b/docs/content/en/pages/docs/database.jade
@@ -957,6 +957,9 @@ block content
 			pre: code.language-javascript { type: Types.Url }
 			.options
 				
+				h5 Options
+				p <code>format</code> <code class="data-type">Function</code> - process URL before its printed in the href/text of Admin UI cell. Defaults to <code class="default-value">removeProtocolPrefix</code>
+
 				h5 Underscore methods:
 				p <code>format()</code> - formats the stored value by stripping the leading protocol (if any)
 				pre: code.language-javascript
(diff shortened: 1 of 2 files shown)


creynders:
Regardless of its usefulness (can't comment on that, since I haven't really used the url field yet) I'd move the conditional to the constructor instead of running it on each `format` call. Every nanosecond counts :grin: 
Also, since it's a new option, could you please add it to the documentation? More specifically to https://github.com/keystonejs/keystone/blob/master/docs/content/en/pages/docs/database.jade

danielzurawski:
Thanks for the feedback. I changed the function name to what it actually is (protocol, not just http) and moved the conditional to constructor. Also, added section in the docs.

JedWatson:
Looks good, @danielzurawski! Thanks.

PR #1215, opened 2015-03-14
Don't use HTML entities in JSON response

ttsirkia:
In JSON responses, there should be raw data and not HTML entities. It should be client-side responsibility to avoid any injections.

Because the data contains HTML entities, Admin UI's autocomplete feature currently shows items like `Hyvink&auml;&auml;`.


JedWatson:
Good point; this slipped in when we switched to React which now handles auto-escaping. Thanks for the PR!

Issue #1216, opened 2015-03-15
Better options for using UpdateHandler manually


ttsirkia:
`UpdateHandler`makes very nice job for updating items in the database by taking new data directly from `req.body`.

However, there are cases when only some fields should be updated. For example, user (and I mean end user, not admin) cannot never change his/her admin status or something like this. It is possible to set `noedit` (btw. inconsistent casing here) to `true` but then the value cannot be anymore updated from Admin UI.

I would like to suggest two improvements:
* create two field options: one for Admin UI and another for other updates via `UpdateHandler`
* add a new option to `UpdateHandler` for blacklisting certain fields

Of course it is easy to set the fields that should be updated but if the model contains 20 fields and only two of them should be ignored, blacklisting is easier than whitelisting.

Any ideas, comments or suggestions?

ttsirkia:
Any objections if I add new option `exclude` to UpdateHandler's options? It partially solves this issue.

mxstbr:
The updateHandler will be completely replaced with the 0.4 update to using an API for communication with the admin interface. I'll close this issue, thanks for chiming in though!

PR #1226, opened 2015-03-17
Add exclude option to UpdateHandler

ttsirkia:
The main reason for this PR is to make `UpdateHandler` more usable in own code. If there are many fields in the model, blacklisting fields may be easier than whitelisting the wanted fields. This PR adds a new option `exclude` to achieve this.

There are two ways to use the option:

```javascript
updater.process(newData, {exclude: 'user, createdAt'}, function(err) {
```

```javascript
updater.process(newData, '-user, -createdAt', function(err) {
```

If a string is passed instead of an object and the string starts with a hyphen, the string is considered as the list of fields to be excluded.

This PR partially implements #1216. This was the easier part of that and the new field-level option still requires comments and ideas.

### DIFF
--- a/lib/updateHandler.js
+++ b/lib/updateHandler.js
@@ -75,7 +75,8 @@ UpdateHandler.prototype.addValidationError = function(path, msg, type) {
  * Processes data from req.body, req.query, or any data source.
  *
  * Options:
- * - fields (comma-delimited list or array of field paths)
+ * - fields (comma-delimited list or array of field paths to include)
+ * - exclude (comme-delimited list or array of field paths to exclude)
  * - flashErrors (boolean, default false; whether to push validation errors to req.flash)
  * - ignoreNoedit (boolean, default false; whether to ignore noedit settings on fields)
  * - validationErrors (object; validation errors from previous form handling that should be included)
@@ -104,7 +105,12 @@ UpdateHandler.prototype.process = function(data, options, callback) {
 	if (!options) {
 		options = {};
 	} else if ('string' === typeof options) {
-		options = { fields: options };
+		// If the field list starts with a hyphen, exclude fields
+		if (options.trim()[0] === '-') {
+			options = { exclude: options };
+		} else {
+			options = { fields: options };
+		}
 	}
 	
 	if (!options.fields) {
@@ -114,6 +120,15 @@ UpdateHandler.prototype.process = function(data, options, callback) {
 		options.fields = options.fields.split(',').map(function(i) { return i.trim(); });
 	}
 	
+	if (!options.exclude) {
+		options.exclude = [];
+	} else if ('string' === typeof options.exclude) {
+		options.exclude = options.exclude.split(',').map(function(i) {
+			// Remove possible hyphens
+			return i.trim().replace(/^-/, '');
+		});
+	}
+
 	options.required = options.required || {};
 	options.errorMessage = options.errorMessage || this.options.errorMessage;
 	options.invalidMessages = options.invalidMessages || {};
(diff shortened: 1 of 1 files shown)


creynders:
Sorry for the wait.

Some questions:

- wouldn't it be better to implement a mongoose style `options.fields`, i.e. `id, name, -createdAt, -user`? Speaking of which: it's a real pity `fields` takes comma-delimited strings, would've been more conform to have it be space-delimited. (Nothing to do with your PR obviously @ttsirkia )
- What happens if a required field is excluded (by accident of course)? Does this blow up, or fail silently?

I'd definitely prefer a mongoose-style exclusion TBH.

ttsirkia:
The idea is like this:
* If you want to update all fields, no options is used
* If you want to update only certain fields, list them in `options.fields`
* If you want to update all fields except some, list them in `options.exclude`

In my opinion, there is no need to mix the whitelist and blacklist. Why you would like to update only some fields but in the same time prevent updating other fields?

If a required field is excluded, the original value is not modified. If you are filling a new object, the save will fail because of the missing field.

creynders:
>Why you would like to update only some fields but in the same time prevent updating other fields?

It depends, if the to-be-updated fields list is created dynamically it is useful. For instance, let's say you've got an API call which always passes 7 field/values, yet depending on some condition you might not always want to accept all field values, yet you _do_ always want to blacklist the same fields.
Anyway, writing it out like this: 
a) I confess it's a bit convoluted _and_ specific 
b) it _does_ seem better to blacklist them more explicitly.

>If a required field is excluded, the original value is not modified.

Sorry, I'm a bit thick sometimes, does this mean that if you pass it a changed value for a required field, it won't update and it won't throw? If that's the case, that will produce some pretty hard to find "why doesn't it update this value?"-bugs, no?

ttsirkia:
The only difference here is that if you exclude the field, `UpdateHandler` won't try to update that field. You should use `exclude` option in cases where you don't want user is ever able to modify the original value. If the field is required and there is no original value, save will fail in the normal fashion and complain about the missing required value.

The real use case for this is updating existing objects, not creating new ones where it does not make sense to exclude fields.

ttsirkia:
And if you use both `fields` and `exclude`, the field will be excluded although it would be present if the `fields` list so you can generate the options dynamically but still use some defaults.

creynders:
Yeah, sorry, should've clarified (doing multiple issue/pr reviews at the same time, not good at multi-tasking ;)) I meant in the case that the update tries to empty a required field value.

So, just to be clear, e.g. we've got a `title` field which is required and it already has a value in the DB.
The payload object however has `title:''`. In that case what happens? Will it be ignored, or?

ttsirkia:
If you have explicitely excluded `title`, UpdateHandler won't try to modify that field. In that case, the original value will remain and no errors are thrown because everything is OK.

creynders:
Ok, got it. So, that leads us to the following question: shouldn't a warning be thrown if the value of a field is modified, but it's blacklisted?

ttsirkia:
I'm not sure what you mean. If you blacklist a field, it won't be modified. And if a field is blacklisted, there is no point that the field exists in the request, for example.

ttsirkia:
So everything works as before, execpt you have a more user-friendly option to select the fields. If you have five fields in your model and want to update only four of them, you can use `field1, field2, field3, field4` as before or simply `-field5` by using this new feature. Nothing else has changed.

creynders:
Yeah, true. It's more of a question in general. 
For me this is ok, @JedWatson thoughts?

ttsirkia:
@JedWatson any thoughts?

ttsirkia:
I'd still love to see this in action without my own patch - makes the updating of Keystone simpler. :smile:

ttsirkia:
@creynders @JedWatson : I think I have to withdraw this because there has not been any interest for this and currently the PR cannot be anymore merged without conflicts.

morenoh149:
! jed is just pretty busy. I'm sure there's interest if this PR is useful. I would verify this myself if I was more familiar with the pain points.

mxstbr:
Since the `updateHandler` will be removed in 0.4, I'm going to close this out.

Very sorry for letting this sit stale for so long, and thanks so much for your work on this @ttsirkia, we really appreciate it. I promise we'll do our best to be more responsive in the future!

PR #1239, opened 2015-03-23
Swap `prepost` for `grappling-hook`

creynders:
Drop `prepost` for ~~pandoras-hook~~ grappling-hook. Fully backwards compatible.

Also, adds a `updates` hook.


creynders:
@JedWatson I just realised this _could_ be a breaking change, if people have added middleware accepting callbacks with any other name than `next` or `callback`.

creynders:
merged into v0.4 branch

Issue #1249, opened 2015-03-24
Trigger a 404 from within view


qwales1:
Is there a way to pass an error out of a view to be handled by the default handlers in Keystone? What I was hoping to do was pass along a 404 if a query returned no results. Passing an error to next() doesn't stop the view from rendering. 

```
var keystone = require('keystone');

exports = module.exports = function(req, res) {

var view = new keystone.View(req, res),
locals = res.locals;

// locals.section is used to set the currently selected
// item in the header navigation.
locals.section = 'serviceareas';

view.on('init', function(next){

    var q =  keystone.list('ServiceArea').model.find().sort('sortOrder');
    q.exec(function(err, result){
        if(err) next(err);
        if(result.length === 0){
            var notfound = new Error();
            notfound.status = 404;
            next(notfound); // any error I pass to next() doesn't stop the render call
        } else {
            locals.serviceareas = result;
            next();
        }

    });
})

// Render the view
view.render('index');

};
```

morenoh149:
try instead of passing the error, throwing it right then and there. I think that would produce a 500 error btw.

qwales1:
Yes throwing there causes an uncaught exception. I could use the response object directly, but then would have to implement custom 404's for each route. I was hoping to just pass it along to the default handler.

albertogasparin:
I think this is what you are looking for:

```js
if (result.length === 0) {
  return res.status(404).send(keystone.wrapHTMLError('Sorry, no page could be found at this address (404)'));
}
```

`keystone.wrapHTMLError` is a method that generates a string of the default error page. 
In addition, you can pass any number to `status()` and any string as `send` argument. I prefer a "Not found" string and then let nginx do the error page rendering.

On a final note, the `res` object is quite powerful. You can also handle dynamic redirects like:

```js
if (result.length === 0) {
  return res.status(301).redirect('/');
}
```

qwales1:
@albertogasparin Thanks for pointing out ``keystone.wrapHTMLError`` . I did not know about that and that is definitely what I'm looking for as the last step in the error handling. Sorry I think I could have done a better job naming this issue in hindsight. Passing a 404 error out of the view was what I was initially trying to do when I ran into this. The real issue for me is that ``View`` doesn't have the ability to pass errors back out to the middleware chain. I think maybe it could be useful beyond just handling 404s if it did. For instance, if someone needed to change the error handling for the whole site, they wouldn't have to think about the specific view code that was written.

[...]

PR #1257, opened 2015-03-26
pass queue errors to next middleware from view

qwales1:
This is in reference to issue #1249. It passes out errors from ``initQueue`` to the next middleware. I am not sure if this is something you want to implement as it changes the View function signature to include next as a third parameter and stores a reference to next within the view, but wanted to take a stab and it and see how it could be done.


creynders:
Some remarks:

- I'd drop the default value for `this.next` and add a check to the [conditional](https://github.com/keystonejs/keystone/pull/1257/files#diff-2ee7035f560888fa5c0f1c0e929032fdR373):
```js
if(err && this.next){
  this.next(err);
}else{
//...
```
- It has a possible side-effect we need to think through: if any of the `pre("render")` middleware throws an error, it will be passed along too. I'm not entirely sure this is always wanted.
- In general: it would be cool if you could try and clean up all the small code formatting changes (tabs swapped to space, semi-colon dropped, ...) Obviously for code hygiene, but also since it makes the PR easier to read, i.e. see what exactly has changed.
- I don't think the signature change is a big problem, since it's backwards compatible
- **_if_** it has no side-effects I'm :+1: with the PR, thanks!

qwales1:
Awesome thanks for looking at that. In response to your remarks:

1. Agree, much nicer with the conditional
2. I didn't think about this, but I will definitely take a closer look and see if there is a way to  account for both situations.
3. Ooops sorry about that. I thought I configured the editor correctly. I will clean that up.

I'll try and get those changes done soon. Thanks!

qwales1:
@creynders I looked at the effect on the ``pre('render')`` queue and I couldn't think of a way to distinguish the errors between those and the ``initQueue`` .  Some possible solutions I came up with:

1. Add to the documentation that if you want errors that occur when rendering this view to be handled by middleware then pass ``next`` as the third parameter when initializing the view.

2. Change this PR to still include the change to the function signature and store a reference to ``next`` in the view, but not call ``this.next`` unless an option is also passed to the view. Something like ``var view = new keystone.View(req,res,next, {renderOnError: false})`` so you can explicitly buy in to this behavior.

3. Scrap this PR and maybe make a module that extends view to have this behavior. I tried briefly, but haven't gotten it to work yet.  

What do you think?

creynders:
I _think_ 1. should be good enough. However, it does change functionality implicitly. Need to consult @JedWatson on this one. Thanks for the follow-up!

PR #1262, opened 2015-03-27
escape script tag from json

### LINKED ISSUE #1250: Embedded <script> in HTML field breaks react editor
I've been using keystone for a blog, and I've discovered that since the 0.3.x upgrade to react in the editor, an embedded script in an HTML field (that is to say, the blog content) breaks the editor. Specifically, the `</script>` tag does not seem to be properly escaped and leads to an `Unexpected EOF` error. I'm not very familiar with react (yet) so I'm not sure how to go about tackling this one. I'd really appreciate any help – my blog thirsts for embedded tweets!

lojack:
This fixes #1250. We need to escape </script> tags in the json response.


JedWatson:
Thanks @lojack

Issue #70, opened 2013-11-21
Forgot Password


mountainash:
No link for users on sign in screen to reset a forgotten password.

ekynoxe:
Hi @JedWatson I'm keep to have a go at this one. What do you mean by "Keystone's auth model"? Is there a particular file for this to go? Looking in the lib directory Would this be part of the session file? It seems that it's more of a utility, but it's related

Also, I believe it will also need a new route.

I'll have a go that way to get the ball rolling.

dcousens:
https://github.com/keystonejs/keystone/blob/823554c9b0e76dcac508765c59217c27b4e6bcc5/templates/views/signin.jade#L30

creynders:
Closing this in favor #1311

PR #1265, opened 2015-03-29
Password reset attempt, am I in the right direction?

ekynoxe:
Hi guys, I'm in need of the password reset for my project with keystonejs, so I figured I'd have a go at it.

I'm looking at issue https://github.com/keystonejs/keystone/issues/70 and I've tried to follow the comments left by @JedWatson  back then.

So far, this solution kind of works, but it's not really elegant, and there is still a lot to be done, I'm sure.
I'm struggling a little to find out how to organise the code in keystone, especially around how/when to set temporary password and validity dates on the user model.

With the suggestion on the issue comments, does it mean that the user model must have those fields in the schema? Or would there be a way to not have to add them there? @JedWatson you mentioned making them part of the password field. Would that mean on the PasswordType.js, maybe through new paths alongside the 'hash' and 'confirm' ones?

Thanks guys!


creynders:
hi @ekynoxe I just reviewed this PR. I couldn't get it to work i.e. the e-mail wasn't sent and the temporary password wasn't saved to the DB. Anyway, I could however review the flow and views et cetera. 

There was a minor issue with `user` not being passed to `sendResetPassword`, i.e. I think [this line](https://github.com/keystonejs/keystone/pull/1265/files#diff-0a6235c4445253f49367a39ab5ff3b7cR226)
should read: `user.sendResetPassword(user,onSuccess);`

However, I'm going to disagree with @JedWatson on this one (if I understood him correctly). I think the more user friendly  (and in the current implementation more secure) approach would be to generate a unique token instead of a password, send a reset password link to the user containing the token, then in the view use the token to look up the user.
The validity of the token should be no more than 10-20 minutes. Also, all the temporary data should _definitely_ be cleaned up afterwards.

If you don't mind I'm going to have a go at this, since I need this pretty urgently.

ekynoxe:
Hi @creynders Strange this doesn't work, as this is what I have in my deployed ( but test ) site. Maybe I forgot to commit some files? I'll be the first to admit I am still unsure about the structure of a keystone project, and I may well have some code in my app that's not in the keystone package.

Anyway, by all means, have a go at it, and let me know if I can be of any help. I only jumped on it as I too needed it pretty quickly, so what I have right now kind of works, but is, as you noticed, not ideal!

Cheers
Matt

creynders:
@ekynoxe yeah I think it was a combination of incorrect settings on my end and not having added the `temporaryPassword` and `temporaryPasswordValidUntil` to the User schema. Doesn't matter.

I branched off your PR, so your commits will still be included; credit where credit due.

bassjacob:
Hi @ekynoxe, from your comment on https://github.com/keystonejs/keystone/pull/1311#issuecomment-96359886 I'm going to close this PR and continue the conversation on that PR. Please re-open this one if you'd like to work on this instead.

PR #1277, opened 2015-04-02
MarkdownField: Uncaught ReferenceError: markdownOptions is not defined

albertogasparin:
At [line 92 of MarkdownField.js](https://github.com/keystonejs/keystone/blob/master/fields/types/markdown/MarkdownField.js#L92) there is `markdownOptions` which is not defined. Probably left from an older version of the script.
The variable should be `options`.


JedWatson:
Thanks @albertogasparin

Issue #1281, opened 2015-04-06
Relationship select field shuffling order of entries


lojack:
Using a Relationship field with `many` set to true, I've been running into an issue where the elements in the field are being reordered when adding and removing items from the list. 

I've only been able to recreate this on slow networks, and doesn't occur when I'm doing local development.

creynders:
I've had this too, but in my experience this was only a UI problem, i.e. the order _was_ maintained when persisted to the DB. But maybe I was just lucky.

JedWatson:
Fixed by #1287

PR #1287, opened 2015-04-07
Fix relationship field shuffling

lojack:
fix for #1281


JedWatson:
Thanks @lojack, looks good :)

PR #1290, opened 2015-04-08
Save original file name on AWS uploads

Subash:


Subash:
*aws, autocomplete messed up

morenoh149:
https://github.com/Subash/keystone/blob/master/fields/types/s3file/S3FileType.js#L406 Is the `protocol` var used here?

morenoh149:
https://github.com/Subash/keystone/blob/master/fields/types/s3file/S3FileType.js#L395 Is `path + filename` not the original filename?

Subash:
@morenoh149, filename is a temporary file name that looks like, `ff196d9499e3a4898874e65bc131d13f`

creynders:
@Subash we're re-writing all file fields, including s3. The original file name will automatically be saved in the new version.

JedWatson:
@creynders maybe we should merge this into 0.3, since we're deprecating the individual file fields in 0.4, what do you think? safe to do?

creynders:
@JedWatson you're right, there's no harm in merging it into v0.3

JedWatson:
Cool, I'll merge it then :)

Thanks @Subash!

PR #1291, opened 2015-04-08
Allow boolean evaluation of dependsOn for fields and headings

### LINKED ISSUE #1130: dependsOn truthy issue and S3 file no longer links in admin UI
In  version 2.x   I could use

```
dependsOn: { fieldName: true }
```
to match (I assume) a truthy value.

For example:

```
Foo.add({
	name: { type: Types.Text, required: true, index: true },
	optionalInfo: { type: Types.Text },
	moreAboutOptional: { type: Types.Text, dependsOn: { optionalInfo: true } },

	upload: { type: Types.S3File, s3path: 'uploads' },
	uploadExtendedInfo1: { type: Types.Number,  dependsOn: { upload: true } },
	uploadExtendedInfo2: { type: Types.Textarea, dependsOn: { upload: true } }
});
```

In version 2.x moreAboutOptional would appear as soon as you tabbed/clicked out of a non-empty optionalInfo box.  The uploadExtendedInfo fields would only appear after a save (but that's good enough for my use case).

screen shot 2015-03-01 at 2 09 59 pm

In version 3.x  neither shows up, even after the save.

screen shot 2015-03-01 at 2 08 09 pm

The second issue is the 2.x version gives a link to the s3 file and the 3.x version just gives a 
text box with the filename.

(sorry, not sure if github etiquette is to report these separately, I just noticed this when I was making the screenshots)

albertogasparin:
Allow the `dependsOn` option to be `true` or `false`. If the field is populated then it will return true.
Fixes #1130

### DIFF
--- a/admin/src/components/FormHeading.js
+++ b/admin/src/components/FormHeading.js
@@ -4,7 +4,11 @@ function evalDependsOn(dependsOn, values) {
 	if (!_.isObject(dependsOn)) return true;
 	var keys = _.keys(dependsOn);
 	return (keys.length) ? _.every(keys, function(key) {
-		var matches = _.isArray(dependsOn[key]) ? dependsOn[key] : [dependsOn[key]];
+		var dependsValue = dependsOn[key];
+		if(_.isBoolean(dependsValue)) {
+			return dependsValue !== _.isEmpty(values[key]);
+		}
+		var matches = _.isArray(dependsValue) ? dependsValue : [dependsValue];
 		return _.contains(matches, values[key]);
 	}, this) : true;
 }
--- a/fields/types/Field.js
+++ b/fields/types/Field.js
@@ -18,7 +18,11 @@ function evalDependsOn(dependsOn, values) {
 	if (!_.isObject(dependsOn)) return true;
 	var keys = _.keys(dependsOn);
 	return (keys.length) ? _.every(keys, function(key) {
-		var matches = _.isArray(dependsOn[key]) ? dependsOn[key] : [dependsOn[key]];
+		var dependsValue = dependsOn[key];
+		if(_.isBoolean(dependsValue)) {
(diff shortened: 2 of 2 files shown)


JedWatson:
Thanks @albertogasparin

lojack:
This doesn't seem to be backwards compatible and breaks dependsOn for me

albertogasparin:
Sorry @lojack, could you provide a failing scenario?
I have tested:

``` js
dependsOn: { state: 'published' } // single string
dependsOn: { state: ['published','needs review'] } // array
dependsOn: { texField: true } // field has string value
dependsOn: { relationField: true } // field has object value
dependsOn: { state: 'published', relationField: true } // multiple fields
dependsOn: { state: 'published', texField: ['','something'], relationField: true } 
```

lojack:
It doesn't work for me for the following:
```javascript
BasePage.add(
    {
        heading: 'Preview',
    },
    {
        hasPreview: {
            type: Types.Boolean,
            default: false
        },
        previewTitle: {
            type: String,
            collapse: true,
            dependsOn: { hasPreview: true }
        },
        previewDescription: {
            type: String,
            collapse: true,
            dependsOn: { hasPreview: true }
        },
        previewImage: {
            type: Types.CloudinaryImage,
            collapse: true,
            dependsOn: { hasPreview: true }
        },
    }
);
```

albertogasparin:
Gotcha. The bug is caused by `_.isEmpty`, because it returns `true` for objects with keys or vars with `length` > 0.  However, `_.isEmpty(true)` returns `false` :disappointed: 
I was missing this testing scenario:

```
dependsOn: { boolField: true } // field has boolean value
```

creynders:
Nice bug hunting @albertogasparin @lojack

JedWatson:
Indeed, thanks for tracking it down guys. Will submit a fix for #1303 in a moment.

PR #1297, opened 2015-04-11
Updated embedded tweets on keystone home page

leo7568:


JedWatson:
Nice @leo7568, thanks!

Issue #1301, opened 2015-04-13
v0.3.5 breaks `schema.path` field retrieval


creynders:
If I upgrade to v0.3.5 suddenly all `list.schema.path("field")` return `undefined`.
This is a major issue since it's mainly used for field value validation. E.g.:

```js
Document.schema.path( "link" ) .validate( function( value,
                       isValid ){
    if( !!value ){
      isValid( !this.file.size );
    } else {
      isValid( !!this.file.size );
    }
  }, "You need to supply either a local file OR an external link." );
```

On a side-note: it would be pretty useful if the `vx.x.x` tags were pushed to the repo too, it allows for easier diffing between versions.

JedWatson:
Hmm. May have been related to the new `inherits` feature... I'll check it out.

+1 for tagging versions, I'll implement that.

JedWatson:
Tagging versions is in, #1302 

On the downside this is definitely a side-effect of #1121

The paths aren't added to the schema before the list is registered, which in hindsight is not really a workable solution. I'll have to figure out a better way of handling it, _fast_.

We should really add a test to catch this in the future... I wasn't expecting to need to test Mongoose's functionality from within Keystone but this has demonstrated a need to ensure we don't accidentally break access to it.

lojack:
What we would need to do is add the base schema fields on instantiation and go back to the old way of adding fields to the schema on add(). 

There was a reason I was adding fields to schema in register(), but I don't believe thats really necessary with my final solution.

JedWatson:
Fixed it, you're right @lojack we can add the fields to the schema immediately with the way it's written now. I'm going to push a new version out with this fix so nobody else gets caught by it.

PR #1305, opened 2015-04-14
fixed an issue with schema inheritance and schemaPlugins

lojack:
This solves an issue brought on changes introduced by fixing #1301.

When using inheritance, if the parent list was using any of the schemaPlugins (i.e. had `track` set to `true`) then an exception would be thrown because of duplicate keys. These plugins should be added before any of the fields are inherited to prevent this.

Not sure if moving the plugin definition to the list initialization will have unintended side effects.

### DIFF
--- a/lib/list.js
+++ b/lib/list.js
@@ -104,7 +104,23 @@ function List(key, options) {
 		return initialFields || (initialFields = _.filter(this.fields, function(i) { return i.initial; }));
 	}});
 
-	if (this.options.inherits) {
+	if (this.get('sortable')) {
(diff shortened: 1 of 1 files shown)


JedWatson:
Damn, I thought I merged this one - must have read it and then it disappeared from my notifications. Sorry! Merging now, will publish an updated version soon.

SharadKumar:
#1363 works now.
We can close that, @JedWatson .

Thanks @lojack.

JedWatson:
Great, thanks for catching that @SharadKumar

PR #1308, opened 2015-04-15
Fix TinyMCE image upload fail due to "Bad response"

### LINKED ISSUE #1304: TinyMCE cloudinary image upload broken on 0.3.5/0.3.6
@JedWatson  Cloudinary image upload from TinyMCE (with the uploadimage plugin) got broken. It returns a "Bad server response" error message. 
Digging into the code, it looks like the plugin tries to read the body of the iframe, with the JSON payload in it. However, [a recent commit](https://github.com/keystonejs/keystone/commit/178db7f9cfc0ba7c4cbf023aeed7d8a65ee0c45c)  changed the response from `text/html` to `application/json`.

That kind of response should be avoided with iframes because in Firefox and Chrome the JSON payload is wrapped with a `<pre>` tag but in IE9+ it generates `.json` file download!

To fix Chrome and Firefox we can change one line in the plugin (replace `.innerHTML` with `.textContent`) but IE will remain broken. 
Another quick solution would be adapting the response type based on the request.
Obviously the best thing would be upgrading the TinyMCE plugin to use FileReader + AJAX for upload...

albertogasparin:
Adapt response based on the headers. TinyMCE upload plugin uses the iframe transport technique so the response type must be text/html. Fixes #1304

### DIFF
--- a/routes/api/cloudinary.js
+++ b/routes/api/cloudinary.js
@@ -12,11 +12,20 @@ exports = module.exports = {
 			}
 
 			cloudinary.uploader.upload(req.files.file.path, function(result) {
-				if (result.error) {
-					res.json({ error: { message: result.error.message } });
-				} else {
-					res.json({ image: { url: result.url } });
-				}
+				var sendResult = function () {
+					if (result.error) {
+						res.send({ error: { message: result.error.message } });
+					} else {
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @albertogasparin

PR #1311, opened 2015-04-16
Password resets

creynders:
This is a continuation on the excellent work of @ekynoxe in #1265 

Chose a slightly different approach in userland:
1. user clicks "Forgot your password?" on signin page
   screen shot 2015-04-15 at 22 57 07
2. when clicking "Reset password" a mail is sent with a unique, one-time, reset-link instead of a temporary password
   screen shot 2015-04-15 at 22 57 37
3. when clicking the link in the mail which leads him to something like `/keystone/changepassword/RVHnlNXEXpxecrjv52zcc2IAQ2c=`
   the user is logged in behind the screens and he's prompted for a new password immediately
   screen shot 2015-04-15 at 23 04 05
4. if the password change is successful he's redirected `/keystone`

So to get this up and running:
1. an `email` setting must be provided with `keystone.set('email', 'admin@example.com')` Probably I have to rename this to `noreply email` or something like that?
2. a `templates/reset-password.jade` template must be provided
3. in your UserModel you need to add `User.schema.methods.sendResetPassword = keystone.user.sendResetPassword;`

Food for thought:
1. I used a new `lib/user` module which is exposed as `keystone.user`. I'm not entirely convinced it's the best name, but it didn't feel right adding the various methods to `lib/session` and all of the functions are statics that _are_ applied to a user object (albeit indirectly): `attemptReset`, `attemptSignin`, `sendResetPassword`.
2. The unique tokens are stored in a `app_users` collection in the DB. I chose a separate collection for several reasons: didn't want to mess with user defined lists, separation of concerns in the DB and TBH I couldn't figure out on how to add fields to an already registered list.
3. I somewhat followed @jedwatson's train of thought in #70, i.e. the user could set a different `sendResetPassword` function to `User.schema.methods.sendResetPassword`, but since the other two functions `attemptReset` and `attemptSignin` are still used, it doesn't really make sense.
   I think a different approach is needed. I'm leaning towards allowing a bit of configuration like for instance the email subject, maybe the sender e-mail address and a few other things (and toggling the password reset functionality on and off of course) and leave it entirely up to the user to provide a different password reset mechanism should they want to.

@jedwatson @ekynoxe what do you guys think?


ekynoxe:
Hi guys, @creynders @JedWatson. Sorry I've been busy on other things lately.
Thanks for the update, this looks a lot better than what I had. I'm more comfortable with this approach.
Not having much experience with the overall architecture was one thing I knew would trip me over.
I also prefer the user lib approach, as I felt adding those methods to the session file wasn't the right place indeed.

Regarding attaching the methods to the user model, shouldn't an attempt to add them be made anyway once the user model is loaded? Password reset is (I think) one of those things that will be needed regardless? If users want to override them, I believe they can do it after the mount step?

I guess I mean: I'm not sure why we need to check whether or not to enable password reset. But again, this may well be because of my lack of understanding about the different modes Keystone can be ran in.
Maybe enable password reset by default and use a config flag if that needs to be disabled?

dcousens:
> when clicking the link in the mail which leads him to something like /keystone/changepassword/RVHnlNXEXpxecrjv52zcc2IAQ2c=
the user is logged in behind the screens and he's prompted for a new password immediately

Does this mean the user can effectively use the session to perform other actions?

lib/session.js
```diff
@@ -30,28 +30,28 @@ function hash(str) {
```
dcousens: All changes in this are unrelated to this PR.  They are good changes, but unrelated.  It'd be sad if something stopped this PR and these changes didn't make it in.  Please make another PR with just these and tag me so we can merge this.

lib/user.js
```diff
@@ -0,0 +1,111 @@
...
+	Reset.update({_id: user.id}, data, {upsert: true}, function(err) {
+
+		if (err) {
+			return callback(err);
+		}
+
+		return new keystone.Email({
+			templateName: 'reset-password'
+		}).send({
+				to     : user.email,
```
dcousens: Indentation is all screwy here

lib/user.js
```diff
@@ -0,0 +1,111 @@
...
+					userId=reset.id;
+				}
+				reset.remove(function(){
+					callback(err, userId);
+				});
+			}else{
+				callback(new Error("Reset token invalid"));
+			}
+		},
+		function(userId, callback){
```
dcousens: `){` or `) {`,  please be consistent with the formatting :)

lib/user.js
```diff
@@ -0,0 +1,111 @@
...
+			if(reset){
+				if(!reset.expires || moment().diff(reset.expires) >= 0 || !reset.token || reset.token !== token){
+					err = new Error("Reset token invalid");
+				}else{
+					userId=reset.id;
+				}
+				reset.remove(function(){
+					callback(err, userId);
+				});
+			}else{
```
dcousens: Please include spaces between `}else`

lib/user.js
```diff
@@ -0,0 +1,111 @@
...
+			var mongoose = keystone.mongoose,
+				Reset = mongoose.model('App_User');
+			Reset.findOne({token: token}).exec(function(err, reset){
+				callback(err, reset);
+			});
+		},
+		function(reset, callback){
+			var userId,
+				err;
+			if(reset){
```
dcousens: Please include spaces between `if(reset){`

lib/user.js
```diff
@@ -0,0 +1,111 @@
...
+		next(new Error('todo: e-mail must be string'));
+	}
+};
+
+exports.attemptSignin = function(token, req, res, next) {
+	async.waterfall([
+		function(callback) {
+			var mongoose = keystone.mongoose,
+				Reset = mongoose.model('App_User');
+			Reset.findOne({token: token}).exec(function(err, reset){
```
dcousens: Please include spaces between `{token: token}`,  like you did in your changes above.

lib/user.js
```diff
@@ -0,0 +1,111 @@
...
+				callback(err, reset);
+			});
+		},
+		function(reset, callback){
+			var userId,
+				err;
+			if(reset){
+				if(!reset.expires || moment().diff(reset.expires) >= 0 || !reset.token || reset.token !== token){
```
dcousens: Can you simplify this line instead to be:

``` javascript
if (!reset.expires || !reset.token) return callback (new TypeError("Invalid reset data"))
if (moment().diff(reset.expires) >= 0)) return callback(new Error("Reset token expired"))
if (reset.token !== token) return callback(new Error("Reset token does not match"))
```

There is also a timing attack possibility in this last line, not sure if that needs to be mitigated.

dcousens: There is also no tests for all of this.

dcousens: @JedWatson is there rate limiting implemented anywhere?

lib/user.js
```diff
@@ -0,0 +1,111 @@
...
+			if (err) {
+				return next(err);
+			}
+			if (user && user.sendResetPassword) {
+				return user.sendResetPassword(next);
+			}
+			next();
+		});
+	} else {
+		next(new Error('todo: e-mail must be string'));
```
dcousens: `new TypeError`, also, why is this `todo:`?  (Curious)

test/lib/session-test.js
```diff
@@ -53,7 +53,7 @@ describe('Keystone.session', function() {
 				sinon.assert.calledOnce(req.session.regenerate);
```
dcousens: All changes in this file are unrelated to this PR.  Please extract them and move to another PR.

lib/user.js
creynders: Re. simplification: good idea
Re. tests: Ummm, yes. TBH I'm waiting out until we have a solid testing strategy and approach.
Re. rate limiting: No, there's no rate limiting implemented AFAIK.

test/lib/session-test.js
creynders: Will do my best, they're a part of a commit by someone else.

lib/user.js
creynders: Re: the timing attack, could you elaborate @dcousens ?

creynders:
@dcousens thanks for the comments. 

>Does this mean the user could effectively use the session to perform other actions without changing the password?

Yes, it does. Probably that _should_ change.

JedWatson:
@creynders @dcousens re: the "change password" screen, let's authenticate access to that with the token then _after_ they successfully set the password, log them in and bounce them to `/keystone`.

re: timing / rate limiting, no that's not currently a thing. I've (historically) been trying to have a _very light touch_ on apps built with keystone - no keystone-specific schemas (except for `app_updates`) or assumptions made about user model (well, as few as possible).

Given how often Keystone is used with defaults though (from what I've seen) it might be better to make some more assumptions so that we can build in more functionality like this.

Let's tackle both those things as separate issues / PRs.

I'm reviewing this more thoroughly at the moment, will come back with more comments soon.

ekynoxe:
Hey guys, the lack of test and session test fixes are both my fault.
Lack of tests because I was really unsure of the direction I was going for my first PR. Really, it was only intended for a review to see if I wasn't completely wrong.

@creynders I'll remove the session messages changes and associated tests and submit another PR for them.

@JedWatson that's what I was hinting at regarding "defaults" and password reset. It seems to me that it should be part of the defaults provided anyway.

ekynoxe:
@creynders @dcousens I've extracted the session error messages changes in their own PR as requested (https://github.com/keystonejs/keystone/pull/1356).

creynders:
@dcousens @JedWatson 

>re: the "change password" screen, let's authenticate access to that with the token then after they successfully set the password, log them in and bounce them to /keystone.

What would be the best way to verify it is still the same user? I.e. token verified -> changePassword view rendered. User enters new password and clicks "submit". What would be the most secure way to verify this is a valid request? 

1. upon initial token verification, generate a new token, write token to DB and set it as hidden field in the passwordChange form
1. when receiving the POST with `user_id`, `password`, `password_confirm` and `token`, see if `token` matches the token in the DB, if so -> update password and sign user in. If not -> :bomb:

creynders:
So, I squashed @ekynoxe to one commit. Dropped the formatting changes to lib/session.
Rewrote the logic as described above to:

* user clicks email link, lands on /keystone/changepassword/<token>
    1. `req.param.token` is verified with the token in the DB
    1. reset data is updated with a `verified:true`
    1. view is rendered with a hidden `nonce` field, containing the value of `nonce` from the reset data in the DB
* user fills out password and confirmation and clicks submit:
    1. `req.param.token` is verified with the token in the DB
    1. `req.body.nonce` is matched with the nonce in the DB
    1. reset data is removed from the DB
    1. user password is updated
    1. user is automatically signed in.

Any mismatch with `token`, `nonce`, et cetera will blow the whole thing off. 
The `verified` field was added to the reset data to enforce the user to start all over again if anything goes wrong in between token verification and password updating.

Maybe I _could_ bump the expiry with 15 minutes once the user lands on the /changepassword page, just to preclude the edge-case in which a user requests a new password, waits 59 minutes, then clicks the email link, and right before him finishing completion of the changepassword form the token expires.

Question I still need some guidance for: how do we let the developer configure this?
Just a flag? `keystone('allow password resets', true);` or???

bscheibe:
Any updates on when this might get pushed through?

jacargentina:
@JedWatson Would be nice to have this on final v0.4. Is it mergeable btw?

jacargentina:
@JedWatson I can take a shot on this, to make mergeable with master. I really **need** this for 0.4 final. please!

larsha:
Any updates on this PR? Let me know if I can help somehow.

bassjacob:
Hi @creynders, is this PR something you're still interested in exploring? If so, is there anything I can do to help with that?

creynders:
@bassjacob as far as I'm concerned this can be closed. It was for KS 0.3 anyway.

VinayaSathyanarayana:
Is this functionality available in 4.0?

export-mike:
@VinayaSathyanarayana no it is not, however I have created a plugin which you can use in your application. https://github.com/thinkmill/keystone-password-reset

christroutner:
ConnextCMS also has a Keystone View that allows for password reset. Feel
free to copy the code from that project and implement it in your own
Keystone installation. Here's the documentation page for it:

http://connextcms.com/documentation/basics.html#passwordReset

VinayaSathyanarayana:
@export-mike  - I tried to use the plugin https://github.com/Thinkmill/keystone-forgotten-password
I am getting an error at `enhanceUser(User);` 

Any suggestions?

mcclowes:
@VinayaSathyanarayana I think the readme is wrong. From looking at the index.js for the plugin, try enhanceUserModel(User).

export-mike:
Can you share a stack trace?

acmoune:
@creynders , is [this](https://github.com/keystonejs/keystone/pull/1311#issue-33419477) working, can we have the sample project ? If it works I will just use it as is.

Issue #126, opened 2014-01-17
Field validation: Max characters


jossmac:
It'd be nice to be able to add { maxChars: x } to a field

albertogasparin:
+1 :+1: 
However I think that `crop` is bad UX. Cropping text without notice could lead to undesired strings.
I think the default behaviour should be `validate` (aka Twitter-like): the counter goes red, displays a -N and the item cannot be saved.
The alternative behaviour is `warn`, which makes the counter orange but nothing more.

bradwbradw:
+1

ericelliott:
> I think the default behaviour should be validate (aka Twitter-like): the counter goes red, displays a -N and the item cannot be saved.

:+1:

bradwbradw:
@ericelliott  agreed!

prachi-159:
In which file do we have to add this feature ?

mxstbr:
We're closing all feature requests to keep the issue tracker unpolluted. I've moved this over, from now on submit feature requests on [productpains.com](https://productpains.com/post/keystonejs/field-validation-max-characters)!

shankarjois:
What is status of this?

VinayaSathyanarayana:
is this not in latest beta?

sheldon-sminq:
There is a bug in the beta version
current code
`
text.prototype.validateInput = function (data, callback) {
        var max = this.options.max;
        var min = this.options.min;
        var value = this.getValueFromData(data);
        var result = value === undefined || value === null || typeof value === 'string';
        if (max && typeof value === 'string') {
                result = value.length < max;
        }
        if (min && typeof value === 'string') {
                result = value.length > min;
        }
        utils.defer(callback, result);
};
`
With fix
`
text.prototype.validateInput = function (data, callback) {
        var max = this.options.max;
        var min = this.options.min;
        var value = this.getValueFromData(data);
        var result = value === undefined || value === null || typeof value === 'string';
        if (max && typeof value === 'string') {
                result = value.length < max;
        }
        if(result) {
           if (min && typeof value === 'string') {
                result = value.length > min;
           }
        }
        utils.defer(callback, result);
};

`

tsuNammi:
Does this work for the WYSIWYG/tinymce too?

PR #1313, opened 2015-04-16
Implement text field max length w/ validation and counter

albertogasparin:
As discussed on issue #126.
New option for field type Text:

``` js
max: 50 // limit the length to 50. Produces validation error if longer
max: { chars: 50, mode: 'limit' } // same as above
max: { chars: 50, mode: 'warn' } // shows a "warning" counter if longer
```

### DIFF
--- a/fields/types/Field.js
+++ b/fields/types/Field.js
@@ -53,7 +53,23 @@ var Base = module.exports.Base = {
 	},
 	
 	renderField: function() {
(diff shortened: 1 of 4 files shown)


richardleggett:
I'm looking to add this myself, is it something shelved due to the 0.4 admin UI?

JedWatson:
It is shelved due to the 0.4 Admin UI, apologies this isn't going to make i tin @albertogasparin! We'll make sure the feature turns up in the new version.

PR #1347, opened 2015-04-27
Fix linting errors

creynders:
Fixes _all_ linting errors. Definitely needs to be reviewed. 
I chose rather to ignore a line, than to risk breaking things, so this consists out of a lot of ignores.

Things of note: 

- this still seems to be a problem: https://github.com/yannickcr/eslint-plugin-react/issues/50#issuecomment-96708326
- encountered a new bug with incorrectly regarded missing spaces after commas, e.g.
  ```js
  //error thrown for this
  links.push(<span key={'ds' + i} className="separator">,</span>);
  ```
- [this should be rewritten I suppose](https://github.com/keystonejs/keystone/compare/keystonejs:master...creynders:fix_linting_errors?expand=1#diff-69385a658d55bf29d4f61c144d9bc77eR31)


JedWatson:
The did-mount-set-state thing isn't actually a problem because the callback will be async. It's just throwing the linter off. I'll jump in and fix it after we merge this, don't want to mess you up with conflicts.

JedWatson:
:+1: AWESOME work @creynders. Merging now.

PR #1353, opened 2015-04-28
Change default formatString for Types.Datetime

WingedToaster:
The default display for the Types.Datetime field is only a date, as the control is technically for a date and time I kindly suggest that it should display both the date and the time. I found this issue when trying to display a create and modify date by using the <code>noedit: true</code> option on the field.

createdDate: {type: Types.Datetime, required: true, index: true, default: Date.now(), noedit: true}

### DIFF
--- a/fields/types/datetime/DatetimeType.js
+++ b/fields/types/datetime/DatetimeType.js
@@ -22,7 +22,7 @@ function datetime(list, path, options) {
 	this._properties = ['formatString'];
 	
 	this.typeDescription = 'date and time';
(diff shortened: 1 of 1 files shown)


WingedToaster:
Looks like the build timed out, can it be restarted please? This was a simple text change and it shouldn't fail...my first pull request ever is failing :-(

JedWatson:
@WingedToaster don't worry about the build, it's happened to a bunch of them in the last few hours (to me and @creynders also). This is a good change, thanks!

PR #1356, opened 2015-04-28
Fixing session error messages

ekynoxe:
(pedantic alert) I found those while working on my password-reset branch, so extracting that out to be committed on its own.


dcousens:
ACK

@JedWatson good to merge.  Just a question of whether you think this is a breaking change or not.

JedWatson:
Looks good to me, don't expect it will break anything. Thanks @ekynoxe!

Issue #1355, opened 2015-04-28
Convert the List view from Jade to React


JedWatson:
Currently working on this with @pieer and @leo7568 in the `react-list` branch.

Moving ahead with converting the Admin UI to React, one of the next goals is to convert the List view from Jade to React.

There are several things that need to be done in preparation:

- [x] Implement an additional List script (for now, filters will remain in Jade - it's our next target!)
- [x] Set up the Field Type Columns infrastructure
- [ ] Load the list results from a JSON API, will also fix #1348
- [ ] Turn the Pagination and Header into React components
- [ ] Re-implement the Sortable and Delete functionality in the React Table

Then we need to port all the Field Types to React column classes:
- [x] Text (needs to support being the name field, and linking to the item)
- [ ] Name (needs to support being the name field, and linking to the item)
- [ ] Email
- [ ] Boolean
- [ ] Key
- [ ] Textarea
- [ ] File/s (all providers)
- [ ] CloudinaryImage/s
- [ ] Relationships (including subfield handling)
- [ ] HTML
- [ ] Markdown
- [ ] Code
- [ ] Color
- [ ] Date / Datetime
- [ ] Date / String / Number Array
- [ ] Embedly
- [ ] Geopoint
- [ ] Location
- [ ] Money
- [ ] Password
- [ ] Select
- [ ] URL

mxstbr:
Closing this as it's done.

PR #1360, opened 2015-04-30
Converted boolean field from jade to react

leo7568:
Converted boolean field from jade to react. This change relates to the issue:
Convert the List view from Jade to React #1355

### DIFF
--- a/admin/src/columns.js
+++ b/admin/src/columns.js
@@ -1,7 +1,7 @@
 var React = require('react');
 
 var Columns = {
-	text: require('../../fields/types/text/TextColumn')
+	text: require('../../fields/types/text/TextColumn'),
 	boolean: require('../../fields/types/boolean/BooleanColumn')
 };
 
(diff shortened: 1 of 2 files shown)


admin/src/columns.js
```diff
@@ -2,6 +2,7 @@ var React = require('react');
 
 var Columns = {
 	text: require('../../fields/types/text/TextColumn')
```
JedWatson: missing comma here

fields/types/boolean/BooleanColumn.js
```diff
@@ -0,0 +1,23 @@
+var React = require('react');
+
+var BooleanColumn = React.createClass({
+
+	render: function() {
+		var value = this.props.data.fields[this.props.col.path];
+		if (value === true) {
```
JedWatson: not sure that value will always be a boolean, the truth check would be safer (and more consistent with the jade implementation)

i.e `if (value) {`

JedWatson:
:+1: thanks @leo7568!

PR #1361, opened 2015-04-30
added 'image/svg+xml' mime type to cloudinaryImage(s)

drlogout:
just added 'image/svg+xml' to 
"fields/types/cloudinaryimage/CloudinaryImageField.js" and
"fields/types/cloudinaryimages/CloudinaryImagesField.js"


JedWatson:
Good call @drlogout, thanks!

PR #1222, opened 2015-03-17
File fields

JedWatson:
@sebmck has contributed a functionally complete (?) implementation of generic `File` and `Files` fields.

The job from here is to merge the latest changes in and ensure the fields work correctly across all the use-cases (including the implementation of the backwards compatible field types for `LocalFile` etc, which will remain until `0.4.0`)

Once these are up and running, we'll create branched `Image` versions of them which will resolve the issues floating around to make `File` fields play more nicely with images (including previews, etc)

Please sound out here if you're interest / capable of helping with the effort to get this merged!

PR #1362, opened 2015-04-30
S3Files Field

sarriaroman:
A field to upload multiple files to Amazon S3 using the LocalFiles Field for the UI


JedWatson:
@sarriaroman I really appreciate you adding this, is there any chance you'd be interested in contributing to the `file-fields` branch with myself and @creynders?

We're actually working to get rid of all the specific file fields and replace them with just *two*:

- `Types.File`
- `Types.Files`

... and each field can be bound to a storage configuration.

Unfortunately merging this PR at this stage may not be a good idea because it will conflict with our transition to the generic file field, I'm interested in @creynders view on this though...

JedWatson:
The issue relating to the File fields is #1222

sarriaroman:
@JedWatson I'm agree with the point of view of @creynders but I see the solution as a future solution. I will try to put work to get it faster.

This is a fast solution for a problem I had for a long time ;).

JedWatson:
Thanks for the PR @sarriaroman and sorry for leaving it hanging so long. We'll be coming back around to file fields soon with a different approach (cc/ @mxstbr @creynders), so I'm going to close this.

PR #1364, opened 2015-05-03
Adding React Column Types for the List View

LennyBoyatzis:
As per issue - Convert the List view from Jade to React #1355

Converted Number and Password fields


admin/src/columns.js
```diff
@@ -2,7 +2,9 @@ var React = require('react');
 
 var Columns = {
 	text: require('../../fields/types/text/TextColumn'),
-	boolean: require('../../fields/types/boolean/BooleanColumn')
+	boolean: require('../../fields/types/boolean/BooleanColumn'),
+	number: require('../../fields/types/number/NumberColumn'),
```
LennyBoyatzis: Realised there's no need for a number field it's text

fields/types/password/PasswordColumn.js
```diff
@@ -0,0 +1,24 @@
+var React = require('react');
+
+var PasswordColumn = React.createClass({
+
+	render: function() {
+		var value = this.props.data.fields[this.props.col.path];
+		if (value) {
+			return (
+				<td>
+					<p>******</p>
```
JedWatson: For most columns, you want this inside the `<td>`:

```
<div className="col-value">{value}</div>
```

the `col-value` class is important for styling.

Checkboxes were an exception

dcousens:
@LennyBoyatzis if this PR is no longer relevant.  Please close.

PR #1365, opened 2015-05-05
GeoPoint display fixes for Type.GeoPoint and Type.Location

WingedToaster:
This fixes a crash when a GeoPoint is added as a column on the index page and there is no GeoPoint saved for the record. This also updates the location field to display the coordinates in Lat/Lng instead of Lng/Lat when editing the record.

### DIFF
--- a/fields/types/geopoint/GeoPointType.js
+++ b/fields/types/geopoint/GeoPointType.js
@@ -62,7 +62,10 @@ geopoint.prototype.getData = function(item) {
  */
 
 geopoint.prototype.format = function(item) {
-	return item.get(this.path).join(', ');
+	if (item.get(this.path)) {
+		return item.get(this.path).reverse().join(', ');	
+	}
+	return null;
 };
 
 
--- a/fields/types/location/LocationField.js
+++ b/fields/types/location/LocationField.js
@@ -136,14 +136,14 @@ module.exports = Field.create({
 		return (
 			<div className="row">
 				<div className="col-sm-2 location-field-label">
-					<label className="text-muted">Lng / Lat</label>
+					<label className="text-muted">Lat / Lng</label>
 				</div>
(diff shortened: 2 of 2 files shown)


morenoh149:
@WingedToaster why should we prefer Lat/Lng over Lng/Lat ?

WingedToaster:
Other humans (read: non-MongoDB/GeoJSON programmers) would typically read and enter the data in it in Lat/Lng format, no?

Also, in it's current state the render of Lat/Lng when one is displaying and editing Types.Location is inconsistent with how Types.GeoPoint is rendered when it is being displayed and edited. <a href="https://github.com/keystonejs/keystone/blob/master/fields/types/geopoint/GeoPointField.js#L18-36">Types.GeoPoint is rendered in Lat/Lng format</a>

JedWatson:
Thanks @WingedToaster, looks good.

dcousens:
@Winged this caught me, expected `Lat/Lng`, got `Lng/Lat`. 

However, this is a breaking change @JedWatson.

WingedToaster:
Sorry about that...what's the bug?

dcousens:
I had code that relies on them being the other way.  :)

JedWatson:
@dcousens they were just changed for display, what are you doing that would break with this? the data is stored in the same format...

The location field displays `Lng / Lat` so we should ensure it is consistent everywhere, I'll open another issue for that.

WingedToaster:
Just to be clear there is a bug fix on this commit/PR so if the Lat/Lng display change ends up getting reverted back, the <code>if</code> statement on <a href="https://github.com/keystonejs/keystone/blob/890ddd7268a59d0196702749b4dfaa0905b7bcc7/fields/types/geopoint/GeoPointType.js#L65">Line 65</a> should remain.

morenoh149:
@WingedToaster for future PRs please do try to make them as granular as possible. So the project can selectively integrate proposals.

PR #1373, opened 2015-05-07
Fix #1370 - inconsistencies between LocalFileType.js & LocalFilesType.js

### LINKED ISSUE #1370: Types:Localfiles missing originalname schema, different returns between Type:Localfiles vs. TypeLocalfile
Just noticed this:

```javascript
// LocalFilesType.js
var schemaPaths = new mongoose.Schema({
	filename:		String,
	path:			String,
	size:			Number,
	filetype:		String
});
```
```javascript
// LocalFileType.js
var schemaPaths = this._path.addTo({}, {
		filename:		String,
		originalname:   String,
		path:			String,
		size:			Number,
		filetype:		String
	});
```

I think this is why I'm having trouble passing a function as the value of Gallery.add.images.filename. I'd like to inherit the file's name & extension where possible, and append 1, 2, 3 etc. where that file already exists. 

```javascript
Gallery.add({
// extra attributes removed for brevity
images: {
type: Types.LocalFiles,
dest: 'public/uploads/images',
filename: function(item, file){
      console.log(**Item**, item)
// item returns entire Gallery object BEFORE upload is added, not the item just uploaded
      console.log(**file**, file)
// file returns the file AFTER upload, with undefined.undefined as file & type
 return item.filename + '.' + file.extension
}
}
});

Nuwen:
Hi!

This is just a quick normalization fix to #1370 - the [documented examples](http://keystonejs.com/docs/database/#fieldtypes-localfile) for the field type LocalFile can now be used with field type Localfiles.

 I added originalname to the schamaPath used in LocalFilesType, and also made sure that the entire file object is passed to field.options.filename.

### DIFF
--- a/fields/types/localfiles/LocalFilesField.js
+++ b/fields/types/localfiles/LocalFilesField.js
@@ -161,9 +161,9 @@ module.exports = Field.create({
 		var value = '';
 		var remove = [];
 		_.each(this.state.items, function (thumb) {
-			if (thumb && thumb.props.deleted) remove.push(thumb.props.public_id);
+			if (thumb && thumb.props.deleted) remove.push(thumb.props._id);
 		});
-		if (remove.length) value = 'remove:' + remove.join(',');
+		if (remove.length) value = 'delete:' + remove.join(',');
 
 		return <input ref="action" className="field-action" type="hidden" value={value} name={this.props.paths.action} />;
 	},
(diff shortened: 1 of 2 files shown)


Nuwen:
Commit 66f8f622d99b734278dab7b39628cbccfeed7f9e was my mistake - I copied over changes made in a different version of LocalFilesTypes.js. 

f74095b7b6133deab901dd60a71397e9f8d23c9f is the correction. 

Keystone is awesome, by the way. Thanks for all the hard work on it. :)

morenoh149:
could you rebase this please? essentially squash the two commits into one.

Nuwen:
Sure. About to finish up for the day but I'll do this first thing tomorrow.

Nuwen:
Okay!

I've also included my fix to keystone#1238 in this rebase.

JedWatson:
Nice - thanks for this @Nuwen :)

Issue #1269, opened 2015-03-31
Add custom validation for sign in


ttsirkia:
The ready-made `signin` function is quite handy if you override the original login view. However, it might be a good idea to add here

```javascript
exports.signin = function(lookup, req, res, onSuccess, onFail) {
```

sixth parameter for a custom validation function which retrieves the `User` and returns `true` or `Error` object. For example, you might have a field for `accountLocked` and that should be checked when user logs in.

I can make a PR if the idea sounds good.

creynders:
Sounds like a good idea, however I'm not sure about passing it as a 6th parameter. Maybe there's a better approach? Maybe we could provide `pre` and `post` "signin" hooks which could be used?
Basically there's a number of situations that might be of interest:

* _before_ the whole shebang
* _after_ the matching user is retrieved, but _before_ authentication
* _after_ authentication

What do you think?

ttsirkia:
Sounds better but requires more work. :smiley:

creynders:
LOL. 

_If_ you're interested in tackling this, you should know I've extracted all hooking functionality from keystone into a separate module: [grappling-hook](https://github.com/keystonejs/grappling-hook) which you can use to add/call hooks. The documentation is pretty thorough, but I you need to know anything, let me know! Keystone v0.4 will use grappling-hook, i.e. in `master` it's not in use yet.

ttsirkia:
Is there need for hook before the whole authentication? Could `pre` be when `User` is retrieved and password is checked but not yet registered to session and `post` after `User` is in Session?

creynders:
Hmm, yeah I don't know. So, for instance (just free-wheeling here) something like this:

```js
keystone.pre("signin", function(user, auth, next){
  if(user.accountLocked){
    return next( new Error("ACCOUNT LOCKED, MOFO!"));
  }
  next();
});
```

It _does_ make sense to have the user object available. I'm just wondering whether there are situations where you'd want to hook into _before_ the user is retrieved. But I must confess I can't come up with a viable use case.

ttsirkia:
Looks exactly what I was thinking.

ttsirkia:
@creynders are you working with this? I might have time next week to implement this feature unless you have already started.

creynders:
@ttsirkia Totally forgot. `grappling-hook` landed into keystone, i.e. this should take me very little time to add. Should have a moment or two today, will keep you updated.

creynders:
Landed. Usage is:

```js
[...]

PR #1378, opened 2015-05-09
Extra hooks: `updates`, `signin` and `signout`

creynders:
I added 3 new hooks:

* `updates`
* `signin` (see #1269)
* `signout`

Some things of note: (will update docs in a separate PR BTW)

* `signin` middleware is executed in the user object scope, i.e. `this` refers to the user object. It means the `pre:signin` middleware are executed _after_ pasword verification. TBH @ttsirkia it _is_ a bit weird though, if the user enters a wrong password the `signin` middleware isn't called, which is **really** odd IMO.
    @dcousens and @jedwatson what do you guys think?
* Had to modify `views/signin` too, since it always outputs "Sorry, that email and password combo are not valid." if an error occurs, which is no longer appropriate.

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,7 @@ var moduleRoot = (function(_rootPath) {
  */
 var Keystone = function() {
 	grappling.mixin(this)
-		.allowHooks('pre:routes', 'pre:render');
+		.allowHooks('pre:routes', 'pre:render', 'updates');
 	this.lists = {};
 	this.paths = {};
 	this._options = {
@@ -205,7 +205,18 @@ Keystone.prototype.import = function(dirname) {
  */
 
 Keystone.prototype.applyUpdates = function(callback) {
-	require('./lib/updates').apply(callback);
+	var self = this;
+	self.callHook('pre:updates', function(err){
+		if(err){
+			callback(err);
+		}
+		require('./lib/updates').apply(function(err){
+			if(err){
+				callback(err);
+			}
+			self.callHook('post:updates', callback);
+		});
+	});
 };
 
 
--- a/index.js
+++ b/index.js
@@ -25,7 +25,7 @@ var moduleRoot = (function(_rootPath) {
  */
 var Keystone = function() {
 	grappling.mixin(this)
-		.allowHooks('pre:routes', 'pre:render', 'updates', 'signout');
+		.allowHooks('pre:routes', 'pre:render', 'updates', 'signout', 'signin');
 	this.lists = {};
 	this.paths = {};
 	this._options = {
--- a/index.js
+++ b/index.js
@@ -25,7 +25,7 @@ var moduleRoot = (function(_rootPath) {
  */
 var Keystone = function() {
 	grappling.mixin(this)
-		.allowHooks('pre:routes', 'pre:render', 'updates');
+		.allowHooks('pre:routes', 'pre:render', 'updates', 'signout');
 	this.lists = {};
 	this.paths = {};
 	this._options = {
--- a/lib/session.js
+++ b/lib/session.js
@@ -55,9 +55,19 @@ function signinWithUser(user, req, res, onSuccess) {
 		}
 		onSuccess(user);
 	});
+
 }
 
 exports.signinWithUser = signinWithUser;
+
+var postHookedSigninWithUser = function(user, req, res, onSuccess, onFail) {
+	keystone.callHook(user, 'post:signin', function(err){
+		if(err){
+			return onFail(err);
+		}
+		exports.signinWithUser(user, req, res, onSuccess, onFail);
+	});
+};
 
 /**
  * Signs in a user user matching the lookup filters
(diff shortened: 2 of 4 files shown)


creynders:
Ouch... Travis complains...

ttsirkia:
Looks great! Could it work so that `pre:signin`is executed before the password comparison and then `post:signin`if the password matched? In this case, `post` function could reject the login if the user account is disabled, for example. And `pre` function could prevent the login in all cases, if this is required because of some maintenance work, for example.

creynders:
Ok, the tests failed, since they assumed `session.signinWithUser` to finish synchronously... So, @JedWatson do you think we need to delay this till v0.4 just in case someone's code is depending on a synchronous `signinWithUser`? It's due to the dezalgofication (what a word, pure poetry! :grin: Remember, you heard it here first!) of `grappling-hook` obviously.

creynders:
@ttsirkia yes, that's a really good idea!

creynders:
Rewrote the last commit so that

>pre:signin is executed before the password comparison and then post:signin if the password matched

index.js
```diff
@@ -205,7 +205,18 @@ Keystone.prototype.import = function(dirname) {
  */
 
 Keystone.prototype.applyUpdates = function(callback) {
-	require('./lib/updates').apply(callback);
+	var self = this;
+	self.callHook('pre:updates', function(err){
+		if(err){
+			callback(err);
+		}
+		require('./lib/updates').apply(function(err){
```
dcousens: That use of inline `require`, is that sane?

lib/session.js
```diff
@@ -115,10 +133,25 @@ exports.signin = function(lookup, req, res, onSuccess, onFail) {
 
 exports.signout = function(req, res, next) {
 
-	res.clearCookie('keystone.uid');
-	req.user = null;
+	keystone.callHook(req.user, 'pre:signout', function(err) {
+		if (err) {
+			console.log("An error occurred in signout 'pre' middleware", err);
```
dcousens: I think I understand why, but @creynders could you elaborate on why this `err` is non-terminating?

dcousens:
@creynders ACK, this LGTM. 

> do you think we need to delay this till v0.4

Absolutely,  if anyone relied on the sign-in being synchronous,  this could become a large security floor in their application.

index.js
creynders: `applyUpdates` and the callback to 'pre:updates' are both one-shots, so yeah it should be ok. Whether it's clean is debatable, but since it already had an inline `require` I left it as is.

lib/session.js
creynders: It didn't seem like a good idea to potentially block a user from signing out just because some 'pre:signout' middleware throws. Safety seemed more important than allowing blocking logic. Can't really come up with a use case where you'd _really_ want to prevent a user from signing out TBH.
Obviously this could be solved by next-ticking the real signout call and throwing the error, sth like:

```js
res.clearCookie('keystone.uid');
req.user = null;

process.nextTick(function(){
	req.session.regenerate(function(err) {
		if (err) {
			return next(err);
		}
		keystone.callHook({}, 'post:signout', function(err) {
			if (err) {
				console.log("An error occurred in signout 'post' middleware", err);
			}
			next();
		});
	});
})

if (err) {
	throw err;
}
```

creynders:
>Absolutely, if anyone relied on the sign-in being synchronous, this could become a large security floor in their application.

@dcousens Sorry, should've added a comment that this is no longer the case (was a comment to an outdated commit). The hooks wrap the "normal" `signin` now and no longer `signinWithUser`, in fact if you use `signinWithUser` directly no hooks are called at all, which means this is stil synced. The "normal" signin is always async anyway, so nothing changes in that case. I.e. there's no real reason to postpone to v0.4

creynders:
Updated the original comment with strike-out to point out it's no longer true.

lib/session.js
creynders: @dcousens do you think that would be a better solution (to `nextTick` it)?

JedWatson:
I think this is good to merge, @creynders can you rebase and confirm we shouldn't wait for 0.4?

JohnnyEstilles:
It would be nice to have this before `0.4.x` ...

creynders:
@JedWatson Rebased and confirmed!

lib/session.js
dcousens: @creynders sorry for the delay,  yes I think anything that more clearly outlines what is actually going on,  is always an improvement.

dcousens: I think that solution is good,  in terms of a graceful handling of a terminating error, its a good outcome; and this definitely makes it more [obviously] ugly, which typically means it will likely to be addressed better later haha.

Any reason for the 2nd non-terminating error in the post hook?

dcousens:
LGTM

JedWatson:
:+1: Thanks! In it goes :)

creynders:
@JedWatson Thanks !!

PR #1381, opened 2015-05-10
Support 'session store' setting as a function

pentode:
This small change will allow keystone to support pluggable express session stores.  I need to be able to use connect-memcached, and looked into adding it as one of the named/supported session store implementations, but as I went further down that rabbit hole, it occurred to me that maybe it was a bad idea to attempt to add support for every session store implementation into keystone.

This allows one to specify the 'session store' setting as a function which will be given keystone's internal express-session instance.  It must return a working express session store.  For example, configuring keystone to use connect-memcached is as straightforward as

```
var keystone = require('keystone'),
    ConnectMemcached = require('connect-memcached')

keystone.init({
  //...
  'session store': function(session){
    return new (ConnectMemcached(session))({
      hosts: [
        'localhost:11211'
      ]
    });
  }
});
```


pentode:
Ugh.  git f-up.  will submit another PR after I learn how to use git...

JedWatson:
@pentode please do, this looks like an excellent addition to Keystone!

JedWatson:
Aah, nevermind, just found #1396

PR #1388, opened 2015-05-12
Admin: Moving assets

milosdakic:
I've been having a look at how to create the admin interface as a fully reusable module, and I think there needs to be a few small steps before we get there.

This pull request moves the public directory within the admin directory, as all assets are for the admin interface.

Possible other steps:
* Move /routes/api to /admin/api
* Move /routes/download to /admin/api
* Move /routes/views to /admin/

One step closer to SPA.


JedWatson:
Nice. Thanks @milosdakic

Issue #903, opened 2015-01-15
Make AdminUI an SPA


morenoh149:
The AdminUI needs to become an SPA. Currently updating the UI in realtime is almost impossible. Saving changes to a List (model) triggers a post request and page refresh. Ghost is a heavy client-side ember app and I think KeystoneJS should have a heavy client-side adminUI as well.

Moreover, features like https://github.com/keystonejs/keystone/issues/816 are very hard to implement given the current architecture. I believe @JedWatson has hinted previously that this is in the pipeline.

JedWatson:
@morenoh149 you're absolutely correct this is in the pipeline, it's the end goal of our move to React.

There is a *lot* of functionality that will need to be rewritten in order to get there, which is why we're taking the current approach of rebuilding each part to function the way it did (i.e. react-generated forms submit the same way the jade ones did).

This means that we can continue to iterate and release while replacing elements in-place with versions that will be (relatively) trivial to migrate into an SPA framework.

Now that the create and edit forms are (basically) done, the major elements left to rebuild are:

- Filters (for the list view)
- Loading list results via API (including filters and pagination)
- Rendering the list from the API (on the list view)
- Rendering item relationships from the API (on the item details view)
- Creating items via API

Then we'll be able to replace the surrounding framework with an SPA that brings it all together, probably using `react-router`.

Now that the field types have all been transitioned I'm making much more rapid progress through rewriting the UI in React (create forms have come together this week, and are nearly completely functional). @jossmac and I are also taking the opportunity to improve the UI as we go (e.g. create forms now show in a dialog on both the item details and list view).

I expect we'll release 0.3.0 as soon as the current functionality is solid (i.e. just using React for the field create and edit forms) so we can get Express 4 out as well, and get back to regular stable releases from `master`.

Then we'll continue to migrate parts of the UI (filters are my next target) throughout the 0.3.x release line, as these should not be major or breaking changes.

adamscybot:
Related to #897

gaurav21r:
@JedWatson @morenoh149  Any timeline on this? I think it would be great if we have a central wiki or something with the roadmap, their status and subdivision of tasks. I am more than sure that the loving keystone community will chip in for any task. But I believe planning is currently being done in silos :cry:

morenoh149:
@gaurav21r See https://github.com/keystonejs/keystone/blob/master/ROADMAP.md no timeline atm. get on the slack if you want to take ownership of a major feature like this.

[...]

Issue #627, opened 2014-09-11
Custom Admin UI routes


steveacalabro:
Is there a way to keep the default Admin UI but change the routes from /keystone to something else ie (/admin)?

frederik:
+1

edword:
+1

JedWatson:
This isn't simple because the `/keystone` path is hard-coded in resources that are checked in to the repo (js / css). We need to get a dynamic build on app init happening first for both js and css that can take into account keystone settings, including generating our css from a template.

All of this is on the roadmap, just wanted to explain why it hasn't been done yet.

frederik:
I added my workaround to https://github.com/frederik/keystone-admin-routes (just a temporary solution since I needed to get this going with subdirectories). Feedback and pull requests are welcome.

claudiosmweb:
+1

digitalcraft:
@frederik I like your solution, I'm thinking about modifying it a bit so I can swap /keystone to /admin(rather than a subdirectory), I would also need to use an nginx rewrite to then proxy anything from /admin back to /keystone.

Until there is an easy way to decouple the path from the assets, I can't think of a better solution.

JohnnyEstilles:
I'd been working on a solution for this as well, not sure if it's worth submitting now that @milosdakic submitted his PR. My implantation was a bit different though.

- treat the Keystone API, the Admin UI routes and the assets used by Admin UI (i.e. css, js plugins, images, fonts, etc.) as separate resources
- leave the Keystone API routes as-is (i.e. `/keystone/api/*`)
- decouple Admin UI assets from the Admin UI routes (i.e. create an Express router to handle routes to all Admin UI assets, mounted as `/keystone/assets/` or something along those lines)
- point all asset the their new location using absolute URIs
- now we're free to mount the Admin UI router anywhere

I have a working version of this that I'm still testing (successfully so far) on a local dirty branch. If you guys think this may be a viable solution I can package it in a PR and submit it.

milosdakic:
@JohnnyEstilles my PR will allow us to do that in stages. Maybe you could take a look and see if my PR could be amended with your changes?

JohnnyEstilles:
@milosdakic! Like yours, my first attempt was to do it in stages, but ran into a number of issues (including the one mentioned by @JedWatson in your PR). What I describe in my comment above is the minimum number of changes I could come up with to implement this. That said, I still have more testing to do to ensure I haven't inadvertently introduced any bugs.

[...]

Issue #471, opened 2014-07-15
Restrict access to keystone admin by domain


JedWatson:
We've got a use-case where a keystone app is running on multiple domains, but we only want `/keystone/` to be accessible through one of them.

To fix this, we could add the ability to restrict access to the `/keystone/` routes based on `req.headers.host`.

milosdakic:
I'll probably do this one in the next day or so.

Which solution do you guys prefer:
- https://github.com/bmullan91/express-subdomain
- http://www.senchalabs.org/connect/vhost.html

Would be cool not only to restrict the admin to a particular domain, but have it load up only on a subdomain. eg. http://keystone.mysite.com/ will only load the admin UI.

grabbou:
Been using express-subdomain in my previous project, but I am not sure if we need another dependency to implement that feature. Should be pretty easy. As long as Keystone do not have (not sure, maybe it's been already done) real multi-tenancy (create and manage multiple sites) that's not needed.

webteckie:
https://github.com/expressjs/vhost looks pretty handy.  Not sure if it is applicable here.

creynders:
We're closing all feature requests to keep the issue tracker unpolluted. From now on submit feature requests on [productpains.com](https://productpains.com/product/keystonejs)!

https://productpains.com/post/keystonejs/restrict-access-to-keystone-admin-by-domain

PR #1389, opened 2015-05-12
Admin: Standalone app + ability to mount different entry point

milosdakic:
- Admin is now standalone Express app that mounts itself onto the Keystone app.
- You can now change the path/route/entry to the admin

```javascript
keystone.init({
  'admin uri': 'admin' // default is keystone
});
```

This should close off #627, paves the way for #471 to be implemented easier, and gets #903 on its way.


admin/public/styles/keystone/list.less
```diff
@@ -282,7 +282,7 @@
 			width: 16px;
 			height: 18px;
 			padding: 0px;
-			background: url('/keystone/images/icons/list-controls.png') top left no-repeat;
+			background: url('../images/icons/list-controls.png') top left no-repeat;
```
JedWatson: this won't work, because pages in the Admin UI are not at a consistent depth. This file needs to be dynamically built with the path to the Admin UI in it when Keystone starts...

unless, @jossmac, could we use less's data-uri feature here to get around this issue?

jossmac: Absolutely, just replace:
`background-image: url( "../images/icons/16/field-text.png" );`
with
`background-image: data-uri("../images/icons/16/field-text.png");`

The caveat being a heavier stylesheet

milosdakic:
@JedWatson @jossmac Fixed issue with images.

gaurav21r:
Any status?

creynders:
@JedWatson @jossmac is this good to go?

JedWatson:
Unfortunately we've moved on from this with the React rebuild; hopefully once we've got that out there the gap to implement custom paths to Keystone will be one we can bridge. Thanks for the work @milosdakic and sorry this isn't being merged!

ajayvishwanath:
Hey @JedWatson  How long until this much-needed feature is merged? :)

JedWatson:
@ajayvishwanath this will be available in Keystone 0.4.x, we're restructuring Keystone's internals to be much more modular in the next release.

ipengyo:
I now need this feature, otherwise I will use admin UI route nginx is mapped into another domain。
@JedWatson 
@milosdakic

thinq4yourself:
Clearly just left everyone standing in the wind with this. Ugh.

thinq4yourself:
Nevermind, it is indeed:

`'admin path': 'admin'`

PR #1392, opened 2015-05-13
Docs: Update settings docs for logger/favicon/compress/static

JohnnyEstilles:
Updating docs that still still reference pre-Express 4.x middleware.


JedWatson:
Thanks @JohnnyEstilles!

Issue #1197, opened 2015-03-11
Slow response times for Admin UI pages


ttsirkia:
I noticed that even with an empty database it takes about 700 ms before browser gets response from the Admin UI. I've tested this with a few virtual machines and the results are the same. Keystone is running in production mode and the other views (my own) are served in less than 30 ms.

Any ideas what makes those pages so slow?

morenoh149:
could you share a screencap of your network waterfall in dev tools. Both FF and chrome have this.

ttsirkia:
There are three users in my database.

`GET /keystone/users 200 1228.269 ms`

![users](https://cloud.githubusercontent.com/assets/2797729/6613393/ca61ec58-c894-11e4-8d91-bb89cebb0628.png)

At the same time, serving my own index view takes about 30 ms.

morenoh149:
😦 somethings wrong. What api endpoint (route) is `GET users` hitting? Could you also share the output of running the query in the mongo shell. http://docs.mongodb.org/manual/tutorial/analyze-query-plan/ aka `explain`. You're probably missing an index somewhere.

ttsirkia:
`users` is something that the Admin UI generates out of the box. It has not my own code. As there are only three users (and actually nothing else), indices cannot be the problem.

I just used `yo` to create an empty project and I'm running it on a virtualized server. The database is empty: `GET /keystone/users 200 757.409 ms` This is a real virtualized server running Ubuntu 14.04. In the screenshot, there is a similar server but it runs my own laptop inside Virtualbox.

After getting the response from the server, it still takes many seconds before the admin view is fully rendered and JavaScript is ready.

So the response time from server is 20 - 40 times slower than any of my own views and the complete render time much worse.

ttsirkia:
Can anyone else confirm this or is there just a common factor in my two environments which causes this?

tivac:
I'm using io.js 1.5.1 with MongoDB 2.6.8 and seeing about a 2.5s wait before I get `/users` HTML back. I have 136 users in my DB.

sreyemnayr:
I'm having the same issue.  Even on a local server, the navigation will appear fairly quickly, then like 6 seconds later the edit form finally shows up.

lojack:
Same issue here, also getting a warning on the backend

```
Note: The code generator has deoptimised the styling of "{redacted}/node_modules/moment/moment.js" as it exceeds the max of "100KB".
```

[...]

PR #1395, opened 2015-05-13
fix babel warning issue #1197

GlennGeenen:


JedWatson:
Thanks @GlennGeenen, this has been supeceded by my new browserify process. The underlying issue has been resolved because all large dependencies have been separated out into the `packages` build, which is not run through babel (and is regenerated with versions in package.json are updated or new common packages are added)

PR #1396, opened 2015-05-13
support custom session store implementation

pentode:
This small change will allow keystone to support pluggable express session stores. I need to be able to use connect-memcached, and looked into adding it as one of the named/supported session store implementations, but as I went further down that rabbit hole, it occurred to me that maybe it was a bad idea to attempt to add support for every session store implementation into keystone.

This allows one to specify the 'session store' setting as a function which will be given keystone's internal express-session instance. It must return a working express session store. For example, configuring keystone to use connect-memcached is as straightforward as

```
var keystone = require('keystone'),
    ConnectMemcached = require('connect-memcached')

keystone.init({
  //...
  'session store': function(session){
    return new (ConnectMemcached(session))({
      hosts: [
        'localhost:11211'
      ]
    });
  }
});
```

This PR now includes documentation for this option.

### DIFF
--- a/lib/core/mount.js
+++ b/lib/core/mount.js
@@ -121,6 +121,8 @@ function mount(mountPath, parentApp, events) {
 	var sessionStorePromise;
 
 	if (sessionStore) {
+		sessionOptions.store = sessionStore(session);
+	} else if (sessionStore) {
 		
 		var sessionStoreOptions = this.get('session store options') || {};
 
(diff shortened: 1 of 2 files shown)


lib/core/mount.js
```diff
@@ -121,6 +121,8 @@ function mount(mountPath, parentApp, events) {
 	var sessionStorePromise;
 
 	if (sessionStore) {
+		sessionOptions.store = sessionStore(session);
+	} else if (sessionStore) {
```
JedWatson: This is actually mean to be
```
+	if (typeof sessionStore === 'function') {
+		sessionOptions.store = sessionStore(session);
+	} else if (sessionStore) {
```

... but it's fine, I'll merge + fix it myself, want to get this in.

JedWatson:
Thanks @pentode!

pentode:
Thanks for merging this...and fixing my error.

PR #1402, opened 2015-05-15
Update less-middleware invocation as to be inline with version 2.X

francesconero:
https://github.com/emberfeather/less.js-middleware/commit/2fa4ce8106178c2ee427b7d68f258df42df35e66

### DIFF
--- a/lib/core/mount.js
+++ b/lib/core/mount.js
@@ -283,9 +283,7 @@ function mount(mountPath, parentApp, events) {
 	// when set, we configure the less middleware
 	
 	var lessPaths = this.get('less'),
-		lessMiddlewareOptions = this.get('less middleware options') || {},
-		lessParserOptions = this.get('less parser options') || {},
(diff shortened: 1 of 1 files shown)


JedWatson:
This is going to be a breaking change... @francesconero is it already broken and I missed it? Or is this forward-looking?

francesconero:
It's currently broken as far as I can tell, as the required dependency is currently 2.0.1 in the package.json. So if you don't want a breaking change we should at least downgrade the dependency?

JedWatson:
@francesconero looks like it's been 2.x since Feb, I missed that change when it was upgraded. I'll just push it through and add some warnings in for the depurated options... the lack of issues reported means it probably hasn't been a broad problem in the wild.

JedWatson:
Thanks for the PR :)

francesconero:
Thanks for Keystone! :P

Issue #1401, opened 2015-05-15
Provide a means to pass options to morgan (logger) middleware


JohnnyEstilles:
At the moment, Keystone consumers can set the output format used by [morgan](https://github.com/expressjs/morgan) via the `logger` option. In addition to the output format, morgan takes an `options` object as it's second parameter with which one can set:

* `skip` - a function which allows us to filter the entries logged, and
* `stream` - an output stream to which morgan will write the log.

I propose we create a new option (e.g. `logger options`) that when set will be passed to morgan as it's second parameter.

```js

var logFilter = function(req, res) {
  return res.statusCode < 400
};

var logStream = fs.createWriteStream(__dirname + '/access.log', {flags: 'a'});

keystone.init({
  ...
  'logger options': {
    skip: logFilter,
    stream: logStream
  },
  ...
});
```

If you guys agree, I'll submit a PR adding this feature. Your thoughts?

JedWatson:
Closed by #1407

PR #1407, opened 2015-05-17
Adding optional `logger options` feature

JohnnyEstilles:
In this PR we address issue #1401
- Add `logger options` and pass as second argument to `morgan` middleware during `.mount()`
- Document `logger options`

And an unrelated issue found while documenting `logger options`
- Fixing `unexpected text |` error caused my missing trailing space (from 000b05455b37aef6ad0795d855d1518212288551)

I apologize for not including any unit tests. Given the current structure of the `.mount()` method, I found it difficult to create an isolated test for this functionality.


JedWatson:
Thanks @JohnnyEstilles. Totally understand about `mount`, I'm really looking forward to breaking it down into manageable pieces with 0.4!

JohnnyEstilles:
That's great news. Let me know if I can be of assistance. :-)

JedWatson:
Certainly - please do see if you can extract any of it into smaller modules without breaking backwards compatibility, that would be really helpful. Similar to the work I did with the new `staticApp` and `browserify` functionality.

PR #1409, opened 2015-05-18
- Allow user to pass own mongoose instance and mongoose dbConnection …

gemscng:
- Allow user to pass own mongoose instance and mongoose dbConnection which will allow multiple database connection in keystone application.

 Simply pass the mongoose and DBConnection to the options when create List.

### DIFF
--- a/admin/api/s3.js
+++ b/admin/api/s3.js
@@ -1,0 +1,48 @@
+var knox = require('knox');
+var keystone = require('../../');
+var Types = keystone.Field.Types;
+
+exports = module.exports = {
+
+	upload: function(req, res) {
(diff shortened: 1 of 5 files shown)


gemng850303:
Recommit using master branch. Seems first commit use old version of keystone.

JedWatson:
Wow... is it really this simple? Happy to merge, just slightly concerned that there may be other ways of accessing the db / connection that this won't work with.

gemng850303:
Yo Jed.  I am not 100% sure yet. I am having a project using keystone. Will update the pull after more test. I create the pull request to see if you guys have and comment of suggestion since i am pretty new to keystone.

Issue #1366, opened 2015-05-05
Chinese autokey skipped all chinese charaters 


arthurtalkgoal:
I set up the ***autokey*** of the model as follows:
```
var Deal = new keystone.List('Deal', {
    map: { name: 'name' },
	autokey: { path: 'slug', from: 'name', unique: true },
  track: true,
  defaultSort: '-createdAt'
});
```
When user input chinese name , the slug will skip all the chinese:
```
維他命E補濕系列、異國麝香、異國麝香沈木、白麝香、白麝香LIBERTINE
```
becomes
***e-e-libertine-***

JedWatson:
This isn't great. Can you suggest a way to translate Chinese characters into URL-safe characters?

It looks like the [pinyin](https://www.npmjs.com/package/pinyin) package might do the trick...

arthurtalkgoal:
Hi @JedWatson, 

worse is that some whole Chinese title will not be able to create the slug at all (it happens in most of my posts).

I know Rails have a very good slug gems doing it greatly:
https://github.com/norman/friendly_id

The related issue is: 
https://github.com/norman/friendly_id/issues/304

Some people want the pinyin, but it is better to do the friendly_id generic way - handle any non-Latin chars (instead of particular Chinese)

And they mention this function in the issue:
http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-transliterate

JohnnyEstilles:
@JedWatson, I've been using a little module called [limax](https://github.com/lovell/limax) for this purpose. It currently support Chinese, Japanese, Cyrilic and Latin scripts. Use cases I've ran into that it doesn't support are Thai and Arabic (there's currently an issue open requesting Arabic). 

This is the most complete transliteration utility (at least for my use-cases) that I have found for node so far. Though I should say my search has been, by no means, exhaustive.

arthurtalkgoal:
Do we all agree with @JohnnyEstilles suggestion.  limax.

If so, I can join the PR.

JedWatson:
I do - having checked it out, limax looks great (although it has a really unfortunately autocorrect, btw - I just had to try 3 times to type that!)

PR would be very welcome implementing it. Thanks!

arthurtalkgoal:
Hi, I updated the ***keystone-utils*** and ***keystone***.

Notice that I added the optional field "defaultLocale" for autokey.

[...]

PR #1415, opened 2015-05-19
Slug for multilanguages

arthurtalkgoal:
Apply new keystone-utils slug method. 

The cld usually failed to find the locale, I allow ***autokey.defaultLocale*** to set up default locale.

### DIFF
--- a/lib/schemaPlugins/autokey.js
+++ b/lib/schemaPlugins/autokey.js
@@ -94,7 +94,7 @@ module.exports = function autokey() {
 		if ((!modified || autokey.fixed) && this.get(autokey.path)) {
 			return next();
 		}
-		var newKey = utils.slug(values.join(' '), null, {locale: autokey.defaultLocale}) || this.id;
+		var newKey = utils.slug(values.join(' '), null, {locale: autokey.locale}) || this.id;
 		if (autokey.unique) {
 			return getUniqueKey(this, newKey, next);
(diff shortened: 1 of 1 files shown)


lib/schemaPlugins/autokey.js
```diff
@@ -94,9 +94,7 @@ module.exports = function autokey() {
 		if ((!modified || autokey.fixed) && this.get(autokey.path)) {
 			return next();
 		}
-
-		var newKey = utils.slug(values.join(' ')) || this.id;
-
+		var newKey = utils.slug(values.join(' '), null, {locale: autokey.defaultLocale}) || this.id;
```
JedWatson: what do you think of just making this option `autokey.locale`?

arthurtalkgoal:
Hi @JedWatson , that's good for me.

notice that cld2 actually just use it as "hints" not absolute locale setting.  Is it still ok to change the name to "locale"

dcousens:
Does this resolve #1366?

arthurtalkgoal:
Hi @JedWatson , I have renamed the field to ***locale***

Hi @dcousens , the #1366 is solving the same issue. Since the keystone-utils function definition has changed, so this keystone has changed accordingly.

JedWatson:
Thanks @arthurtalkgoal, merged & will be released in a new version shortly.

PR #1420, opened 2015-05-20
Add gitter button

morenoh149:
Change gitter link to button. Increase it's visibility.


JedWatson:
Good idea @morenoh149

PR #1423, opened 2015-05-21
Faster travis builds

fearphage:
This should speed up builds considerably in travisci.

[Reference](http://blog.travis-ci.com/2013-12-05-speed-up-your-builds-cache-your-dependencies/)


JedWatson:
Thanks @fearphage, that's great

PR #1436, opened 2015-06-01
Just proposal. Imposing new eslint rule "object-curly-spacing".

AlexKVal:
Though I'm not sure about the space on the last line in such snippets:
```js
Object.defineProperty(azurefile.prototype, 'azurefileconfig', { get: function() {
  return this.options.azurefileconfig || keystone.get('azurefile config');
} });
```

Should I re-write them as ?:
```js
Object.defineProperty(azurefile.prototype, 'azurefileconfig', {
  get: function() {
    return this.options.azurefileconfig || keystone.get('azurefile config');
  }
});
```

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -23,6 +23,7 @@
 		"no-trailing-spaces": 0,
 		"no-underscore-dangle": 0,
 		"no-use-before-define": 0,
+		"object-curly-spacing": [1, "always"],
 		"quotes": [2, "single", "avoid-escape"],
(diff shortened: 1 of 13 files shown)


JedWatson:
@AlexKVal I like it, thanks - have been looking forward to `"object-curly-spacing"` being released.

I do think the second snippet is clearer, yes.

AlexKVal:
You could just say it and I then reformat and `git push -f` it :wink:

PR #1438, opened 2015-06-02
Adds more detailed message to when logging in with wrong email or password

WingedToaster:
Adds the message "Incorrect email or password." when the user account might exist but does not match the password sent since the err object would be null in this case.

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -92,6 +92,7 @@ var doSignin = function(lookup, req, res, onSuccess, onFail) {
 					if (!err && isMatch) {
 						postHookedSigninWithUser(user, req, res, onSuccess, onFail);
 					} else {
+						err = new Error('The email and password you entered do not match.');
 						onFail(err);
 					}
 				});
@@ -108,6 +109,9 @@ var doSignin = function(lookup, req, res, onSuccess, onFail) {
 			if (user && (!passwordCheck || scmp(passwordCheck, hash(user.password)))) {
 				postHookedSigninWithUser(user, req, res, onSuccess, onFail);
 			} else {
+				if (!err) {
+					err = new Error('The email and password you entered do not match.');
+				}
 				onFail(err);
 			}
 		});
--- a/lib/session.js
+++ b/lib/session.js
@@ -110,7 +110,7 @@ var doSignin = function(lookup, req, res, onSuccess, onFail) {
 				postHookedSigninWithUser(user, req, res, onSuccess, onFail);
 			} else {
 				if (!err) {
-					err = new Error('The email and password you entered do not match.');
+					err = new Error('The password you entered do not match the account.');
 				}
 				onFail(err);
 			}
--- a/lib/session.js
+++ b/lib/session.js
@@ -110,7 +110,7 @@ var doSignin = function(lookup, req, res, onSuccess, onFail) {
 				postHookedSigninWithUser(user, req, res, onSuccess, onFail);
 			} else {
 				if (!err) {
-					err = new Error('The password you entered do not match the account.');
+					err = new Error('The password you entered is incorrect.');
 				}
 				onFail(err);
 			}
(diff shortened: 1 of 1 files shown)


lib/session.js
```diff
@@ -92,6 +92,7 @@ var doSignin = function(lookup, req, res, onSuccess, onFail) {
 					if (!err && isMatch) {
 						postHookedSigninWithUser(user, req, res, onSuccess, onFail);
 					} else {
+						err = new Error('The email and password you entered do not match.');
```
ttsirkia: You are overriding here the original error. This may not always be a wise choice.

lib/session.js
```diff
@@ -108,6 +109,9 @@ var doSignin = function(lookup, req, res, onSuccess, onFail) {
 			if (user && (!passwordCheck || scmp(passwordCheck, hash(user.password)))) {
 				postHookedSigninWithUser(user, req, res, onSuccess, onFail);
 			} else {
+				if (!err) {
+					err = new Error('The password you entered is incorrect.');
```
ttsirkia: Here you will expose to a possible hacker that the user account exists but the password is incorrect.

lib/session.js
WingedToaster: Good catch. Updated.

lib/session.js
WingedToaster: Looking at the code it doesn't really confirm one way or another if the user account exists. The return message could be a little more vague, I guess? Suggestions?

Additionally it appears the password is optional for this function anyways, so I'm really not sure anymore if there needs to be a more detailed error for the else side of this function. I'd be curious generally how secure this function is since it appears I can just pass in a user id and no password. It's entirely possible I'm missing something though...

ttsirkia: That's true. But in general, at least in my opinion, it is always a bad idea to state that the password is incorrect. It implies that the username was correct which is bad for two reasons:
1) The username may still be wrong
2) It may leak information about the existing user accounts and give better opportunities to hackers

WingedToaster: I updated the file to be more vague but I'll just say that on a lot of sites you're already exposed just by virtue of using the site, e.g. I already know what your username is on GitHub because you're posting messages to me. Github, Facebook, Twitter, et.al. leak information about their users everywhere (usernames appear on friends/connections lists, contributor lists, etc.) and they are still considered pretty secure as far as I know.

WingedToaster: Created https://github.com/keystonejs/keystone/issues/1448 to discuss possible security issue

lib/session.js
```diff
@@ -89,9 +89,13 @@ var doSignin = function(lookup, req, res, onSuccess, onFail) {
...
 			if (user) {
 				user._.password.compare(lookup.password, function(err, isMatch) {
+					
 					if (!err && isMatch) {
 						postHookedSigninWithUser(user, req, res, onSuccess, onFail);
 					} else {
+						if (!err) {
+							err = new Error('Incorrect email or password.');
+						}
 						onFail(err);
```
dcousens: `onFail(err || new Error('Incorrect email or password')`?

lib/session.js
```diff
@@ -108,6 +112,9 @@ var doSignin = function(lookup, req, res, onSuccess, onFail) {
 			if (user && (!passwordCheck || scmp(passwordCheck, hash(user.password)))) {
 				postHookedSigninWithUser(user, req, res, onSuccess, onFail);
 			} else {
+				if (!err) {
+					err = new Error('Incorrect user or password.');
+				}
 				onFail(err);
```
dcousens: Again

creynders:
@dcousens is this good to go?

dcousens:
@creynders yup :+1:

creynders:
Thanks @WingedToaster for the PR and @dcousens for merging!

PR #1445, opened 2015-06-03
Timezone fix

### LINKED ISSUE #1345: DateTime field type timezone bug
Hi. We've discovered a bug when using a DateTime field type. It occurs when Keystone admin UI is accessed by a client from a timezone different to the one on the server. Say the client machine is 2h ahead of server (EET), and the server itself is in UTC. Steps to reproduce:

1. Fill out a DateTime input with any valid value.
2. Save the document.
3. Edit the document - you'll see previously saved value change by timezone delta.

This is a presentation thing - the value in the database, and the value returned to Keystone UI view is correct - but it is then rendered in user's timezone. When you hit "save" - the value rendered (e.g. 2 hours ahead) gets saved down to the database as a new UTC time.

Here's the model we used:

```
var Banner = new keystone.List('Banner', {
  map: { name: 'imageTitle' },
  autokey: { from: 'imageTitle', path: 'key', unique: true },
  sortable: true
});

Banner.add({
  imageTitle:           { type: Types.Text, label: 'Title' },
  alt:                  { type: Types.Text ,  label: 'Alternative Text'},
  filename:             { type: Types.LocalFile, dest: 'public/images/uploads' },
  inApp:                { type: Types.Boolean, default: true },
  targetUri:            { type: Types.Text },
  validityPeriodStart:  { type: Types.Datetime, required: true, initial: true },
  validityPeriodEnd:    { type: Types.Datetime, required: true, initial: true },
  disabled:             { type: Types.Boolean, default: false , label: 'Inactive'},
  uriSmall:             { type: Types.Url, hidden: true },
  widthSmall:           { type: Types.Number, hidden: true },
  heightSmall:          { type: Types.Number, hidden: true },
  uriMedium:            { type: Types.Url, hidden: true },
  widthMedium:          { type: Types.Number, hidden: true },
  heightMedium:         { type: Types.Number, hidden: true },
  categoryId:           { type: Types.Relationship, ref: 'sportCategory' },
  categoryName:         { type: Types.Text, default: 'default', hidden: true },
  brand:                { type: Types.Text, default: 'bma', hidden: true },
  lang:                 { type: Types.Text, default: 'en', hidden: true }
});
```

lojack:
This fixes #1345, and adds timezone support. Default timezone is set to UTC, and this adds a new keystone option, 'timezone'.

Additionally this fixes an edge case of #1038 that wasn't solved by e7a005866be458086a539d6855835a3523a68db9. If a content editors local machine was in a different timezone as the servers such that the dates don't line up, selecting "Today" would enter an infinite loop.

### DIFF
--- a/fields/components/DateInput.js
+++ b/fields/components/DateInput.js
@@ -1,6 +1,6 @@
 var React = require('react'),
 	Pikaday = require('pikaday'),
-	moment = require('moment');
+	moment = require('moment-timezone');
 
 module.exports = React.createClass({
 	
@@ -14,6 +14,7 @@ module.exports = React.createClass({
 	},
 	
 	getInitialState: function() {
+		moment.tz.setDefault(this.props.timezone);
 		return {
 			value: this.props.value,
 			id: Math.round(Math.random() * 100000)
@@ -25,7 +26,7 @@ module.exports = React.createClass({
 		this.setState({
 			value: newProps.value
 		});
-		this.picker.setMoment(moment(newProps.value, this.props.format));
+		this.picker.setMoment(moment(newProps.value, this.props.format), true);
 	},
 
 	componentDidMount: function() {
--- a/fields/types/date/DateField.js
+++ b/fields/types/date/DateField.js
@@ -2,7 +2,7 @@ var React = require('react'),
 	Field = require('../Field'),
 	Note = require('../../components/Note'),
 	DateInput = require('../../components/DateInput'),
-	moment = require('moment');
+	moment = require('moment-timezone');
 
 module.exports = Field.create({
 	
@@ -14,7 +14,8 @@ module.exports = Field.create({
 	inputFormat: 'YYYY-MM-DD',
 
 	getInitialState: function() {
-		return { 
+		moment.tz.setDefault(this.props.timezone);
+		return {
 			value: this.props.value ? moment(this.props.value).format(this.inputFormat) : ''
 		};
 	},
--- a/fields/types/date/DateField.js
+++ b/fields/types/date/DateField.js
@@ -60,7 +60,7 @@ module.exports = Field.create({
 		if (this.shouldRenderField()) {
 			input = (
 				<div className={fieldClassName}>
-					<DateInput ref="dateInput" name={this.props.path} format={this.inputFormat} value={this.state.value} onChange={this.valueChanged} yearRange={this.props.yearRange} />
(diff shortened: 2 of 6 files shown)


ttsirkia:
How does this behave if the server's timezone is not UTC and no timezone is defined in Keystone? For example, my server is in UTC+2 and Keystone/Mongoose/Mongo saves everything as UTC and converts them back to local time based on server's timezone when reading data back from database.

ttsirkia:
Or maybe better question is, that is the new `timezone` option required in order to prevent any unwanted timezone changes?

lojack:
Right now, if timezone is not set in the config then it defaults to UTC. It will (and should) always save to the database in UTC. I'm open to setting this to fall back to the system timezone.

There is one potential edge case that probably won't outweigh breaking backwards compatibility. Potentially, someone could be using a multi-host instance with servers in geographically separate locations. With multiple timezones being used, this could result in wonky behavior.

Even if we allow using system timezone as a fallback, I think explicitly setting the timezone in keystone's config should be the recommended approach.

ttsirkia:
In my opinion server's timezone would be better default choice if the timezone setting is not set. Of course, for new installations with yo, there should be a default setting for this. In a same way as Django has for empty projects.

creynders:
Is this still relevant?

lojack:
I'll retest this against master when I get a chance to verify if its an issue that's still present.

temojin:
any idea when this will be merged? we're on 0.3.16

ericelliott:
I haven't looked at this or the problem in depth, but in general:

Servers should always be timezone-agnostic. Because servers may be distributed across a variety of different timezones, deciding anything based on the server timezone is a recipe for bugs and inconsistent behaviors.

All dates should be saved in UTC, and *only sometimes* adjusted for the local user's timezone, depending on what the date is used for. For example, events that are happening in a particular location should always be displayed in that particular location's timezone, regardless of where the local user is located, and regardless of where the server is located.

Trying to apply the same timezone logic to every date is a bad idea. =)

riyadhalnur:
I agree with @ericelliott. Servers should be timezone agnostic. The time should be localised for the user, after reading back the date/time from the DB, when displaying it on a page.

morenoh149:
@lojack mind rebasing this on master? sorry for the lag but this seems to still be a present bug

JedWatson:
Thanks for the effort here @lojack 

I'm going to close this specific PR but we're going to do a final check on timezone handling for the 4.0 release, to make sure the issue here is resolved.

pourmesomecode:
@lojack thanks for the `timezone` fix, however, nowhere does it state/cover this in the docs. Am I missing something? http://keystonejs.com/docs/database/#fieldtypes-datetime

PR #1449, opened 2015-06-04
Fix bug saving geo on Types.Location field

WingedToaster:
Now that the admin display is in Lat,Lng format (https://github.com/keystonejs/keystone/pull/1365) the newGeo variable needs to be reversed in the updateItem function to put it back into GeoJSON format when saving to the db.

### DIFF
--- a/fields/types/location/LocationField.js
+++ b/fields/types/location/LocationField.js
@@ -140,10 +140,10 @@ module.exports = Field.create({
 				</div>
 				<div className="col-sm-10 col-md-7 col-lg-6 location-field-controls"><div className="form-row">
 					<div className="col-xs-6">
-						<input type="text" name={this.props.paths.geo} ref="geo1" value={this.props.value.geo ? this.props.value.geo[1] : ''} onChange={this.geoChanged.bind(this, 1)} placeholder="Latitude" className="form-control" />
+						<input type="text" name={this.props.paths.geo + '[1]'} ref="geo1" value={this.props.value.geo ? this.props.value.geo[1] : ''} onChange={this.geoChanged.bind(this, 1)} placeholder="Latitude" className="form-control" />
 					</div>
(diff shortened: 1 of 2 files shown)


JedWatson:
I'm going to have to take a look into this - not all the places that submit to the update handler will have it switched.

WingedToaster:
Yikes...sorry...perhaps that if statement should be reversed?

<code>
	if (valuePaths.geo_lat in values && valuePaths.geo_lng in values) {

		var lat = utils.number(values[valuePaths.geo_lat]),
			lng = utils.number(values[valuePaths.geo_lng]);

		item.set(paths.geo, (lat && lng) ? [lng, lat] : undefined);

	} else if (valuePaths.geo in values) {

		var oldGeo = item.get(paths.geo) || [],
			newGeo = values[valuePaths.geo];

		if (!Array.isArray(newGeo) || newGeo.length !== 2) {
			newGeo = [];
		}

		if (newGeo[0] !== oldGeo[0] || newGeo[1] !== oldGeo[1]) {
			item.set(paths.geo, newGeo);
		}

	}
</code>

WingedToaster:
Shoot...that doesn't work either. 

The <code>Types.GeoPoint</code> type appears to save correctly, can the <code>geo</code> field within the <code>Types.Location</code> field be switched to use <code>Types.GeoPoint</code> instead?

WingedToaster:
@JedWatson I reviewed the code in <code>Types.GeoPoint</code> and found the difference in the way it was passing the data over. I added the appropriate code in LocationField instead and undid my original change.

It looks like the test passes now too, so hopefully this is good to go on your end.

JedWatson:
@WingedToaster that is the right way to fix it - nice work! Thanks.

WingedToaster:
glad to hear it...thanks for bearing with me!

PR #1472, opened 2015-06-17
Update required Elemental version

NullEntity:
The 0.0.9 Elemental version doesn't causes Keystone to fail to compile the less files giving `Error: .u-text-truncate is undefined` on list.less:67.


JedWatson:
Thanks @NullEntity.

PR #1475, opened 2015-06-17
Fix height property for TextArea not being used

SlashmanX:
[The docs](http://keystonejs.com/docs/database/#fieldtypes-textarea) mention that a `height` property can be passed to a textarea to set the height in pixels. However I found that this doesn't work and all textarea's are the same (default) height. This small pull request fixes that.

I followed the same standard used in the MarkdownType to implement this.


morenoh149:
thanks

JedWatson:
Perfect, thanks @SlashmanX.

PR #1483, opened 2015-06-19
Fix #1479, no confirmation warning on delete by binding the event on …

### LINKED ISSUE #1479: No confirmation warning on delete
The "delete {type}" button on the item details view used to have a confirmation warning... in the latest version, it just immediately deletes, which is really unsafe. We need to put the confirmation back in.

gerotakke:
…document

### DIFF
--- a/admin/public/js/common/ui.js
+++ b/admin/public/js/common/ui.js
@@ -11,11 +11,9 @@ jQuery(function($) {
 
 
 	// generic confirm
-	
-	$('a[data-confirm]').click(function(e) {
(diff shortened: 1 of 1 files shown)


gerotakke:
I also removed the return false statement since it unnecessarily breaks event propagation. It might be a good idea to change all the bindings to document instead of the matched DOM elements directly.

JedWatson:
Thanks @gerotakke! appreciate it :)

JedWatson:
Also don't worry about the failing test on Travis - that's a linter issue, I'll clear it up.

PR #1485, opened 2015-06-19
Passing "originalname" to S3Type filename function

moesian:
Allows the use of the original file name when saving to S3.


JedWatson:
Thanks @moesian

PR #1488, opened 2015-06-21
S3Type fixed "orignalname" being undefined

moesian:
Sorry I had quickly copied and pasted the changes in my last commit and have realised that the "orignalname" would be be undefined. This fixes this.


JedWatson:
Thanks @moesian

PR #1499, opened 2015-06-27
Added support for importing an array of model directories

webteckie:
When a model array is specified, each directory in the array is assumed to be an absolute path.  When the specified model directory is a string, the import will be behave as before and resolve its full path using the module root.

NOTE:  This may need documentation.

### DIFF
--- a/admin/public/styles/keystone/forms.less
+++ b/admin/public/styles/keystone/forms.less
@@ -299,6 +299,9 @@ label.checkbox {
 		background: transparent;
 		border-radius: @input-border-radius;
 	}
+	.CodeMirror-linenumber{
+		margin-left:-30px;
+	}
 }
 
 // Related Items
--- a/index.js
+++ b/index.js
@@ -154,29 +154,34 @@ keystone.security = {
 keystone.utils = utils;
 
 /**
- * returns all .js modules (recursively) in the path specified, relative
+ * returns all .js modules (recursively) in the path(s) specified.
+ *
+ * If the input dirname is a string then it will be assumed to be a relative path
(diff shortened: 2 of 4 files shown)


webteckie:
Apologies for the Travis CI failure...didn't realize it had happened!  @JedWatson do you see any negatives with the changes?  They worked fine for me and thought it may benefit the community.  If not you may close the the PR and I'll just keep the changes on my side.

webteckie:
@JedWatson Not sure how to separate the different commits into their own PR; maybe you can just pull the independent commits assuming you agree with the change.  Anyways, the latest commit https://github.com/webteckie/keystone/commit/3bc37c53ef40adc02d4512c05d7c282bdf5825f4 is a very simple change with possibly big impact for some since currently CodeMirror is being loaded in all admin pages.  With this change it will only be loaded in the item form page.

JedWatson:
@webteckie the trick is to submit PRs from branches, not your `master` - any commits to the branch after the PR is created are added to the PR automatically.

I really like your code mirror change, would be great to get that in a separate PR.

About the array of paths, not sure if it's that intuitive, but I'm happy to add support for _something_ like that. Main sticking point right now is the absolute path behaviour change when using an array - this seems like something that could be resolved before the `import` method is called.

Also, seems like multiple calls to `import` would do the same trick?

webteckie:
@jedwatson ok I will split the PR now that you have thought me how :-)  Also it didn't ocurr to me to do multiple calls to import.  I will try that and if it works in my use case then I'll be all set and not need for that particular PR.  Perhaps we can add more documentation around the different ways models can be configured in keystone.

PR #1505, opened 2015-06-30
Check if response has been sent, this enables the view logic to send 404 page during operating

opelhoward:
Check if response has been sent, this enables the view logic to send 404 page during operating

### DIFF
--- a/lib/view.js
+++ b/lib/view.js
@@ -47,18 +47,18 @@ module.exports = exports = View;
  */
 
 View.prototype.on = function(on) {
-	
+
 	var req = this.req;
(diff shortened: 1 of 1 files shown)


JedWatson:
I can see the usefulness of this, but isn't it safer to simply not call the `next` callback from whichever view logic is sending the 404? That would be inline with the way express middleware works.

Otherwise there may be subsequent view logic that would _still_ run, even when the 404 had been sent, and could lead to other issues.

JedWatson:
Going to close this now, I think keeping with the standard express pattern of handling 404s is probably the best option.

PR #1507, opened 2015-07-01
Fixes an error caused by the default value being an empty string

NullEntity:
The default value for an empty name field is '' and not { first: '', last: '' } which the field expects. Typing any character in the first/last name boxes creates a console warning saying "Uncaught TypeError: Cannot assign to read only property 'first'/'last' of ".


JedWatson:
Thanks @NullEntity

PR #1510, opened 2015-07-03
Add uploadcare image field type

alanshaw:
This PR adds support for Uploadcare image fields (https://uploadcare.com/).

Uploadcare is a service similar to cloudinary so the code for the field is very similar.

Configure the field by adding a public and private key to your keystone config:

``` js
{
  "uploadcare public key": ""
  "uploadcare private key": ""
}
```

Note that the field does not yet support the "select" option that cloudinary image fields support because of an [outstanding PR on the uploadcare node module](https://github.com/RexMorgan/uploadcare-node/pull/9) meaning that we can't get a list of uploaded files yet.

If you're happy to merge this then I'll send a PR for uploadcare image**s** asap.


morenoh149:
this looks interesting

mxstbr:
Thanks for taking the time to make this @alanshaw and sorry we never got back to you! :-1:

With #3052 landing soon, I don't expect we'll merge this as-is and rather make this an external npm module – @keystonejs/core-devs what do you think?

bassjacob:
@alanshaw apologies for the delay, is this feature still something you're interested in?
@mxstbr given #3052 is this PR still actionable directly, or should we explore a more generic solution that this feature can be built on top of?

JedWatson:
Thanks for this @alanshaw

We've re-architected the field types for 4.0 which means this can't be merged as-is, so that everyone can implement and publish field types as plugins independently of keystone core.

If you're still using uploadcare and update the implementation for 4.0, let me know so we can help get it working!

Issue #1280, opened 2015-04-04
LocalFiles Type Images have no href


CMarker:
There is no .href property on .image property of LocalFiles Type. Given the example in: https://github.com/keystonejs/keystone/issues/700 there should be a .href property for each image of LocalFiles Type with the correct routable address.

jorisw:
Any word on this? I too am trying to simply link to or download a stored file of type localFile. How do I link to one? Concatenating the prefix and the filename in the object data give me a 404.

CMarker:
<img src="<%= "/files/" + data.post.image.filename %>"class="img-responsive" style="">

This worked for me after some trial and error. It's not ideal but I haven't found a better way yet. The image will be stored at /data/files/ but is exposed via the /files/ route.

CMarker:
`
<img src="<%= "/files/" + data.post.image.filename %>"class="img-responsive" style="">
`

This worked for me after some trial and error. It's not ideal but I haven't found a better way yet. The image will be stored at /data/files/ but is exposed via the /files/ route.

jorisw:
Yeah concatenaning will work. Thanks.

tylkomat:
I can confirm this. It is not related to just images, but any localfiles. I provided a fix for that (#1515)

tylkomat:
also exists does not work. There is no error but also no output.

Issue #700, opened 2014-11-06
LocalFiles image path contains 'public'


davidpaulrosser:
Hi,

First of all I'd just like to say thanks for an amazing cms : )

After implementing the localFiles field type within my model the images do not correctly show within the admin ui.

The path to the images contains 'public' which obviously can't be accessed frontend.

I've worked about this by adding this line above live 15 in `keystone/templates/localfiles/form.jade`

```
- value.path = value.path.replace('public/', '')
```

Is there a nicer way to solve this?

Thanks,

tylkomat:
On windows the `href` path has backslashes instead of slashes. I tested with `LocalFile`.
```javascript
thumbnail: {
    type: Types.LocalFile,
    dest: baseDir + '/data/thumbnails',
    prefix: '/thumb/',
    allowedTypes: [
        'image/jpeg', 'image/png', 'image/gif', 'image/svg+xml'
    ],
    filename: function(item, file) {
        return item.id + '.' + file.extension;
    },
    format: function(item, file) {
        return '<img src="' + file.href + '" style="max-width: 300px">';
    }
}
```
`{{ item.thumbnail.href }}` for example outputs `\thumb\5542306ff9800d1c23a3cf3b.png`

tylkomat:
This happens because `path.join` creates a platform specific path, but you want to create a URL path, string concatenation should be sufficient.

JedWatson:
Thanks for pointing that out @tylkomat, I've fixed it and will push a new version soon.

PR #1515, opened 2015-07-08
Fix for #1280, href was not added to the paths #700

tylkomat:


tylkomat:
Sorry I was too fast something else is missing.

PR #1516, opened 2015-07-08
Fix for #1280 href working now on LocalFiles Type

tylkomat:
Virtual `href` method was missing on nested schema. This fixes the initial issue of #1280. 

I left the rest of the methods how they were, while `exists` for example does not work. I was not clear to me how the expected behavior is.


JedWatson:
Thanks @tylkomat.

JedWatson:
> I was not clear to me how the expected behavior is.

I'm refactoring this fairly heavily for the next release, so might be OK to leave them as they are until then - would have to go digging in there to work this out comprehensively

PR #1519, opened 2015-07-09
Updated TinyMCE Icons (& .ie7) to correct value for BlockQuote button…

jeffreypriebe:
Tiny PR fixing the icon for the blockquote that is included when you set Keystone Admin UI configuration "wysiwyg additional buttons" to include "blockquote" per http://keystonejs.com/docs/configuration/#options-ui

Note Also:

1. The resulting CSS file differs slightly as the compilation step was slightly different (e.g. ".3" instead of "0.3").
2. There is a big chunk of commented rules in Icons.less - likely not needed, but not removed as I wasn't sure of their purpose (I updated the value in the comments as well). These rules were commented out Oct 14, 2013 in keystonejs/keystone@c29642f0fecbdcffa20050245709416a537181c9


JedWatson:
Looks good @jeffreypriebe, thanks!

Issue #1524, opened 2015-07-14
Can't inherit from a model that has history enabled.


jeffreypriebe:
From the inheritance example in the [Keystone docs](http://keystonejs.com/docs/database/), if you turn on the history plugin on the BasePage, you will get an error when trying to launch keystone.
Mongoose throws
```
throw new mongoose.Error.OverwriteModelError(name);
OverwriteModelError: Cannot overwrite `basepages_revisions` model once compiled.
at Mongoose.model (website\node_modules\keystone\node_modules\mongoose\lib\index.js:334:13)
```

As I understand it, this is because the "history" plugin (keystone: lib\schemaPlugins\history.js) creates a model ("basepages_revisions" in the example) whose name is based off the Base model. When it comes around to creating the inherited model, the history plugin tries to create another model - using the same Base information and thus a model with the same name. Mongoose catches the repeat model and throws an error.

See [Gist](https://gist.github.com/jeffreypriebe/d2e3d04303370cf99a00) for example case.

I expected to be able to turn on history, configure inheritance and for it to just work.

### How to fix this?

Note: you can't even disable the history on BasePages and then enable it on ChildPages. The history plugin will still try to write its data into "basepages_revisions."
So, you could get away with this for a single ChildPages (where ChildPages has history, BasePages does not).
But, you cannot create a second "SecondChildPages" inheriting from BasePages with history enabled. Same error as above for the same reason.

jeffreypriebe:
Note sure if [previous closed issue "Plugins seems to break when using List inheritance"](https://github.com/keystonejs/keystone/issues/1363) is related.

lojack:
What version of keystone are you using @jeffreypriebe?

jeffreypriebe:
[v0.3.12 / Jun 25, ecbf4ae](https://github.com/keystonejs/keystone/tree/ecbf4ae309f97a9627390757945c66c33a682410) - current latest on [npm](https://www.npmjs.com/package/keystone).

Have you seen different results with a different version?

JedWatson:
Closed by #1526

PR #1526, opened 2015-07-14
Added check for duplicate _revisions models.

jeffreypriebe:
This fixes Issue #1524.

To avoid collision of inherited models (which reuse the same _revisions historyModel as the model inherited from), check for existence of _revisions model for inherited models and skip attempt to create a new model. (Inherited _revisions model will be used.)

**NOTE**: I'm not at all suggesting this is the RightWay&trade; to solve this issue.
* It does "work on my machine" (ugg) and the related '_revisions' collection in Mongo seems to end up with the expected data.
* At my early understanding of the keystone codebase, it seems like a reasonable solution.

Final note: I couldn't get hub to work to convert Issue to PR, so I apologize if that is still the desired route (per PRO TIP in contributing.md).

### DIFF
--- a/lib/schemaPlugins/history.js
+++ b/lib/schemaPlugins/history.js
@@ -1,8 +1,14 @@
 var keystone = require('../../');
 
+var historyModelSuffix = '_revisions'; 
+
+function getHistoryModelName (list) {
+	return list.options.schema.collection + historyModelSuffix;
+}
+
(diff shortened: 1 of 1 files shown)


JedWatson:
Looks good to me @jeffreypriebe, happy to go with your solution. Thanks!

Issue #1533, opened 2015-07-16
Use  other geocoding-service for location enrichment when no Google Maps API-Key available


stosorio:
As much as I enjoy using Google's APIs, maybe it would be nice to have something that works out of the box without having to provide an API-Key?
Mapquest for example: http://open.mapquestapi.com/nominatim/

stosorio:
Processwire (a PHP CMS/CMF which i used a couple of times) uses google geocoding without needing an API Key.
https://github.com/ryancramerdesign/FieldtypeMapMarker/blob/48090d01a524ed33c1de71d61003e848067e5152/MapMarker.php#L79

Also Google Developer Guide sounds like an API Key isn't strictly neccessary?
https://developers.google.com/maps/documentation/geocoding/intro
> All Geocoding API applications should use an API key. Including a key in your request:
> * Allows you to monitor your application's API usage in the Google Developers Console.
> * Enables per-key instead of per-IP-address quota limits.
> * Ensures that Google can contact you about your application if necessary.

morenoh149:
+1

JedWatson:
You're correct, not sure if this changed but it looks like you don't actually need an api key for a limited number of geocodes with google. Also happy to bring this to the client, but I don't think we should get rid of the server-side implementation because it's useful in other cases. No problem having both options.

I guess we should get rid of the check for the API Key to enable the geolocation functionality, and we can update the generator as well.

stosorio:
Yes, the server-side implementation would definitely stay relevant.
I wouldn't get rid of the check completely though - perhaps add an option to Type.Location to allow geocoding without key?
This is because Google has [certain restrictions](https://developers.google.com/maps/terms#section_10_1) about the ways in which you can use data gained through the API (e.g. you aren't allowed to use it with other map providers or in se), so activating it should still be a deliberate choice.

stosorio:
I suppose if i want to tackle a client-side implementation it would be best to do this in the elemental-integration branch?

stosorio:
Apparently, the check that enables the API [checks for the wrong key](https://github.com/keystonejs/keystone/search?utf8=%E2%9C%93&q=%22google+api+key%22&type=Code)?
But then again, the provided key [isn't used anyway](https://github.com/keystonejs/keystone/blob/123e4f980b5e6ae8583ec1cb03479667b6445daf/fields/types/location/LocationType.js#L371).

PR #1537, opened 2015-07-18
Add Maps API key to request, Allow override per model, check for server instead of browser key

stosorio:
Regarding #1533 
- Added an option to Types.Location, which allows to explicitly enable/disable geocoding in the model, without breaking the current behaviour. 
- Added the Server API key - if available - to the request
- Check GOOGLE_SERVER_KEY instead of GOOGLE_BROWSER_KEY for availabilty of a server API key

### DIFF
--- a/fields/types/location/LocationType.js
+++ b/fields/types/location/LocationType.js
@@ -24,7 +24,7 @@ function location(list, path, options) {
 	this._underscoreMethods = ['format', 'googleLookup', 'kmFrom', 'milesFrom'];
 	this._fixedSize = 'full';
 
-	this.enableMapsAPI = keystone.get('google api key') ? true : false;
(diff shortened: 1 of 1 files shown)


JedWatson:
Great. Thanks for cleaning this up @stosorio :+1:

fields/types/location/LocationType.js
```diff
@@ -368,6 +368,10 @@ function doGoogleGeocodeRequest(address, region, callback) {
 		options.region = region;
 	}
 
+	if (keystone.get('google server api key')){
```
jordanyim1: This addition broke location auto-detect on my site - pressing 'Autodetect and improve location on save' would no longer geolocate the address. Tracked it down to this commit in version 0.3.13 and up.
If I delete the above lines of code it works again.

Could this be removed/tweaked?

Issue #1323, opened 2015-04-21
E-mail not treated as case-insensitive during Admin UI login


JohnnyEstilles:
A Keystone consumer reported on [StackOverflow](http://stackoverflow.com/questions/29759761/i-cant-access-the-admin-ui-in-keystonejs) not being able to log in to the Admin UI with an e-mail address that had uppercase characters. (Thanks to @mikeLspohn for finding this bug.)

I found that `email.updateItem` coverts e-mail addresses to lowercase (see [fields/types/email/EmailType.js line 66](https://github.com/keystonejs/keystone/blob/master/fields/types/email/EmailType.js#L66)), but `session.login` does no such conversion prior to lookup (see [lib/session.js line 78](https://github.com/keystonejs/keystone/blob/master/lib/session.js#L78)).

Submitting PR to fix.

creynders:
I wouldn't convert it to lower case since if the e-mail address isn't in lowercase in the DB, it won't match. Rather use a regex query in mongoose: 

```js
var email  = "info@Creynders.be";
User.model.findOne({email: new RegExp(email, "i")})
```

JohnnyEstilles:
@creynders ... that is exactly what my PR does. Submitting in a few.

creynders:
ok, just wanted to make sure :smile:

JohnnyEstilles:
No worries. I know we all love Keystone and want to protect it. :-)

JohnnyEstilles:
Closed via #1326 ...

PR #1542, opened 2015-07-20
Escape special characters in email before creating a RegExp for case-…

alobodig:
…insensitive lookup

Fixes a bug with emails addresses containing special characters introduced after the fix for #1323. These characters were not being escaped and were interpreted as part of the regular expression.

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -1,6 +1,7 @@
 var crypto = require('crypto');
 var keystone = require('../');
 var scmp = require('scmp');
+var utils = require('keystone-utils');
 
 /**
  * Creates a hash of str with Keystone's cookie secret.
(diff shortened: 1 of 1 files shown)


NullEntity:
If we're going to stay with the regex case insensitive search, at least anchor it to the start of the string by prefixing it with `^`. I'd personally prefer if all emails were just all converted to lower before being stored. Maybe it could be done as a pre save hook.

alobodig:
Sticking with the regex also precludes the use of an index on the field (at least in mongo currently). I am definitely in favor of saving a lower case version instead.

morenoh149:
double check travis ci

JedWatson:
Saving a lowercase version isn't something we can easily do with backwards compatibility so this is a better option for now. Needs some fixes though, I'll comment inline.

Travis errors are unrelated linter errors, not sure why those are showing up for this PR !? -- is it up to date with master?

lib/session.js
```diff
@@ -85,8 +85,11 @@ var doSignin = function(lookup, req, res, onSuccess, onFail) {
 	}
 	var User = keystone.list(keystone.get('user model'));
 	if ('string' === typeof lookup.email && 'string' === typeof lookup.password) {
+		// create regex for email lookup with special characters escaped
+		var emailRegExp = new RegExp('^' + lookup.email.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, '\\$&') + '$', 'i');
```
JedWatson: we have an `escapeRegExp` method in `keystone-utils`, use that please.

NullEntity:
@JedWatson how about storing the current keystone version in the db? That'd allow for updating db schema when Keystone's updated and an update script can be triggered. This script could convert all emails to lower.

alobodig:
Yes, it's up to date with master. Looks like there's a problem there too.

JedWatson:
@alobodig I found the issues in mater and fixed them - thanks for the update!

alobodig:
@JedWatson Excellent!

PR #1555, opened 2015-07-24
added table field(with variable rows and columns)

mohitjindal1992:
This is to add table in field types.
now you can add table with variable number of rows and columns.
In dashboard you will have option to add/delete rows/columns.
No need to hard code number of columns(can be different for different items.)

Note: Will work best if you provide default value as single row and column.
ex:
capture


creynders:
There's a bunch of linter warnings, could you fix those?

JedWatson:
Thanks for the PR @mohitjindal1992 and sorry we never merged it. We've re-architected keystone fields so you will be able to implement this as a plugin external to keystone core, if you're interested in updating this let me know.

PR #1557, opened 2015-07-28
Fix error when trying to use 'file limit' configuration

### LINKED ISSUE #1412: Most middleware (like limit) is no longer bundled with Express
I tried to use `keystone.set('file limit', '10mb')`, but express is complaining, saying:

```
/Volumes/Macintosh HD 2/Projects/project/platform/node_modules/keystone/node_modules/express/lib/express.js:89
      throw new Error('Most middleware (like ' + name + ') is no longer bundle
            ^
Error: Most middleware (like limit) is no longer bundled with Express and must be installed separately. Please see https://github.com/senchalabs/connect#middleware.
```

according to [https://github.com/senchalabs/connect#middleware](https://github.com/senchalabs/connect#middleware) [`raw-body`](https://www.npmjs.com/package/raw-body) should be used now.

jacargentina:
Fixes #1412, caused when trying to use 'file limit' configuration

### DIFF
--- a/lib/core/mount.js
+++ b/lib/core/mount.js
@@ -348,14 +348,17 @@ function mount(mountPath, parentApp, events) {
 		app.use(morgan(this.get('logger'), this.get('logger options')));
 	}
 	
+	var bodyParserParams = {};
+	
 	if (this.get('file limit')) {
 		debug('adding file limit');
-		app.use(express.limit(this.get('file limit')));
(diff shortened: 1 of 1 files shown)


morenoh149:
lgtm

JedWatson:
Nice one @jacargentina, thanks.

PR #1562, opened 2015-07-30
Load CodeMirror only in the item form

webteckie:
This change updates keystone's Admin UI to load CodeMirror only in the item form if and only if a code field is defined.  This should speed up loading of all other pages in the Admin UI as well as the item form if no code field is defined.


JedWatson:
Thanks @webteckie! really happy about this.

PR #1579, opened 2015-08-11
Add S3 support to wysiwyg editor

### LINKED ISSUE #1574: Wysiwyg support for S3
There is support for Cloudinary uploads in wysiwyg but none for S3 uploads, so I just added that on my fork. This seems like it would be a useful feature, but some of the setup I have isn't that elegant right now:
- The `doUpload` function in `fields\types\s3file\S3FileField.js` has some core S3 handling code that feels like it should be shared if uploads are going to happen in other places, like the new `admin\api\s3.js` file that I added to enable this support. (Matching the routing and directory structure for cloudinary)
- In the configuration, there is a property of `wysiwyg cloudinary images` so the reasonable setting to add for S3 would be `wysiwyg s3 images` - I really just wanted to see if you agree with that.

Let me know if you have any feedback.

Ref https://github.com/jrit/keystone/commit/74aba49aa7132b6aec0a529680655588e66b5001

jrit:
This resolves #1574 including the concern about CSRF. The image upload plugin for tinymce expects to find a meta tag with the CSRF value, so that needed to be added to the admin jade template. The rest is modeled on the existing Cloudinary API upload point and uses as much of the existing S3 file handling as seemed practical, mainly for header management.

Please let me know if you have any changes you'd like to see before this can be merged.

### DIFF
--- a/admin/api/s3.js
+++ b/admin/api/s3.js
@@ -1,0 +1,43 @@
+var knox = require('knox');
+var keystone = require('../../');
+
+exports = module.exports = {
+
+	upload: function(req, res) {
+		if(req.files && req.files.file){
+
+			var s3Config = keystone.get('s3 config');
+
+			var file = req.files.file,
+				path = s3Config.s3path ? s3Config.s3path + '/' : '';
+				originalname = file.originalname,
+				filetype = file.mimetype || file.type;
+
+			var s3Client = knox.createClient(s3Config);
+
+			s3Client.putFile(file.path, path + file.name, function(err, s3Response) {
+				var sendResult = function () {
+					if(err){
+						return res.send({ error: { message: err.message } });
+					}
+
+					if (s3Response) {
+						if (s3Response.statusCode !== 200) {
+							return res.send({ error: { message:'Amazon returned Http Code: ' + s3Response.statusCode } });
+						} else {
+							return res.send({ image: { url: "https://s3.amazonaws.com/" + s3Config.bucket + "/" + file.name } });
+						}
+					}
+				};
+
+				res.format({
+					html: sendResult,
+					json: sendResult
+				});
+			});
+
+		} else {
+			res.json({ error: { message: 'No image selected' } });
+		}
+	}
+};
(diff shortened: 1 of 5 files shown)


jrit:
Maybe instead of validating CSRF in the API as `keystone.security.csrf.validate(req, req.body.authenticity_token)` the `keystone.security.crsf.validate` method should just be updated to check `authenticity_token` also? Again, this is just to deal with tinymce's configuration, but this will come up again with the Cloudinary API point if CSRF is added there.

jrit:
@jeffreypriebe this look ok to you?

jrit:
@JedWatson are there issues with this PR or is it just something you'll take a look at later?

JedWatson:
Sorry for the delay @jrit - I've been caught up getting the admin UI rebuild branch ready to merge, and need to work out how to merge this w/o conflicting with the work being done there. PRs and issues will move much faster once we're past that! I'll get back to you soon.

jrit:
Great, thanks for all your hard work. Best CMS I've ever used.

JedWatson:
@jrit I'm about to merge the elemental branch which will introduce a lot of changes. I'd like to review this once that's done so we can fix it properly there, anything else is just adding more work before that release. I think you'll find the updates are worth pushing through asap! If you can check this out again once that lands it would be great.

andrew-kennedy:
Did this get merged into the 0.3 branch? it appears that I can't use this functionality in the stable version of keystone

jrit:
It isn't in 0.3

alechewitt:
Can we merge this functionality into 0.3?

admin/api/s3.js
```diff
@@ -0,0 +1,48 @@
...
+				var sendResult = function () {
+					if(err){
+						return res.send({ error: { message: err.message } });
+					}
+
+					if (s3Response) {
+						if (s3Response.statusCode !== 200) {
+							return res.send({ error: { message:'Amazon returned Http Code: ' + s3Response.statusCode } });
+						} else {
+							return res.send({ image: { url: 'https://s3.amazonaws.com/' + s3Config.bucket + '/' + file.name } });
```
alechewitt: It appears if the bucket was created in a different region to US standard, it prefers the format:
```'https://' + s3Config.bucket + 's3.amazonaws.com/' file.name```

PR #1582, opened 2015-08-13
Fixed sortable so order is set correctly during initialization

Adam14Four:
The sortOrder field needs to start at 1 or the max check will always fail and the sortOrder field for models will be set to 0 when they are created, until the admin reorders the items.


JedWatson:
Thanks @Adam14Four

Issue #1493, opened 2015-06-24
cannot update Address field 


arthurtalkgoal:
Hi Guys,

I cannot save my model in Admin UI, it always pop up an error like follows:

```
 Can't extract geo keys: { _id: ObjectId('55555555'), updatedBy: ObjectId('5555555'), updatedAt: new Date(1435143947013), createdBy: ObjectId('66666'), createdAt: new Date(1430191417150), slug: "cowork", name: "lai chi kok CoWork", address: { street1: "660 Castle Peak Rd", state: "Kowloon", country: "Hong Kong", geo: [ 22.338566, 114.147899 ], postcode: null, suburb: "lai chi kok" }, status: "active", images: [], __v: 1, excerpt: "", website: "", business_hours: "", email: "", phone: "1234444", banner: { public_id: "banner/hturosogsgr4wbzlkgpf", version: 1434011818, signature: "somesecret", width: 960, height: 638, format: "jpg", resource_type: "image", url: "my.jpg", secure_url: "my.jpg" } } can't project geometry into spherical CRS: [ 22.338566, 114.147899 ] (MongoError) 

```

I don't know why an existing address got the ***Can't extract geo keys*** errors. 

It happens for my another address.  

The original working address is:

```
Street Address: 660 Castle Peak Rd
Suburb:  Lai chi kok
State: --
Country: Hong Kong
```

But after save, it turns to the following which ***prohibit my update***:

```
Street Address: 660 Castle Peak Rd
Suburb:  --
State: Kowloon
Country: Hong Kong
```

worse, I got another error if i have not checked the ***Autodetect and improve location on save***

```
There was a problem saving your changes:

    Address is required

```

aks0020:
@arthurtalkgoal +1 @JedWatson I am facing similar issue, I found temporary workaround to swap Lat/Long value, actually the form expect Lat first & then Long. but the `Cordinate` datatype expect Long/Lat. but keystone should internally handle this.

arthurtalkgoal:
Hi @aks0020 , That issue is very annoying in my application.

I allow user input in the frontend. And I verify the user address in backup. However, I am not able to update in Admin UI!

PR #1590, opened 2015-08-19
Validation fix for GeoLocation (for FieldType - Location)

vintesh:
GeoLocation (for FieldType - Location) Fix: Validating new Geolocation correctly when updating field of type location


vintesh:
#1493 should be fixed by this validation fix.

JedWatson:
Thanks @vintesh!

PR #1596, opened 2015-08-20
Update to grappling-hook v3.0.0

creynders:
Major reason for updating grappling-hook would be the addition of thenable middleware. 
However during development of v3.0.0 I realised I'd made a serious mistake when sketching out v2.x: all `callHook` parameters are flattened, i.e. this:

```js
this.callHook('pre:upload', [item, file], function(err) {
```

was translated automatically to 

```js
this.callHook('pre:upload', item, file, function(err) {
```

Which means that if you passed an array directly all of its elements were passed as separate parameters to the middleware functions. Ouch! :imp: 

Since I changed this in v3.0.0 all `callHook` calls in keystone that used an array to pass parameters needed updating, which is what this PR does.

Just to be clear: to the consumers of the middleware, i.e. keystone users, no backwards incompatible changes were introduced, i.e. they do NOT have to update anything, i.e. this is NOT a breaking change.

### DIFF
--- a/fields/types/azurefile/AzureFileType.js
+++ b/fields/types/azurefile/AzureFileType.js
@@ -270,7 +270,7 @@ azurefile.prototype.uploadFile = function(item, file, update, callback) {
 		});
 	};
 
-	this.callHook('pre:upload', [item, file], function(err) {
+	this.callHook('pre:upload', item, file, function(err) {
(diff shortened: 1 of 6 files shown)


JedWatson:
Nice work, @creynders. Thanks for catching that 2.x issue too!

PR #1598, opened 2015-08-21
Fix: Do not update null length strings

ChristianRich:
When updating records with empty fields they are saved as empty strings. When no value is present in a Types.Text field it should not be updated.


JedWatson:
@ChristianDen either this is working as designed, or the fix doesn't look right. You should definitely be able to submit a blank string and have it save to the database; otherwise, you could never remove the value from a field.

The `text` field type has a small number of tests, could you add some that demonstrate the behaviour you are fixing so we can validate what's happening?

See the `number` field type for an example of more comprehensive tests, including ones that would catch (I think) the behaviour you're addressing in `text`.

JedWatson:
We've standardised and corrected this behaviour with the recent push to get fields all unit tested. Thanks for raising it, @ChristianRich!

Issue #670, opened 2014-10-13
In admin UI virtuals can't access field values not included in `defaultColumns`


creynders:
If you want to include a virtual to your `defaultColumns` the fields it accesses _must_ be included in `defaultColumns` as well.

E.g.:

```js
//not working:
Foo.schema.virtual( 'foo' ).get( function(){
  return this.baz;
} );
Foo.defaultColumns = "name, foo";
```
```js
//works:
Foo.schema.virtual( 'foo' ).get( function(){
  return this.baz;
} );
Foo.defaultColumns = "name, baz, foo";
```

I assume some filtering of the document happens for speed reasons, which makes sense, but is there some way we can force the document to have specific fields inside our virtual getters?

morenoh149:
solution proposal?

creynders:
@morenoh149 I suppose the easiest solution would be to allow the explicit exclusion of fields from filtering or another option would be  to allow adding fields to `defaultColumns` yet hiding them. Something like: `MyModel.defaultColumns = "name, someVirtualField, usedByVirtual:hidden"`

pricco:
I'm using the following patch:

```javascript
var keystone = require('keystone');

keystone.List.prototype.defaultSelectColumns = keystone.List.prototype.selectColumns;
keystone.List.prototype.selectColumns = function(query, columns) {
  var list = this;
  var allColumns = columns;
  columns.forEach(function(col){
    var virtual = list.model.schema.virtuals[col.path];
    if(virtual && virtual.depends) {
      // TODO: Check if the column was already added
      allColumns = allColumns.concat(list.expandColumns(virtual.depends));
    }
  });
  return this.defaultSelectColumns(query, allColumns);
};
```
Example:
```javascript
Ticket.schema.virtual('total').get(function() {
  return this.limit || '∞';
}).depends = 'limit';
```

creynders:
@pricco aw man, that's a very cool solution! Maybe turn it into a PR?

morenoh149:
closing as elemental UI merge makes this irrelevant.

sebastiancarlsson:
+1 this bug is still present in 0.3.19

PR #1602, opened 2015-08-22
Fix adminUI virtuals can't access field values

morenoh149:
* fixes https://github.com/keystonejs/keystone/issues/670

### DIFF
--- a/lib/list/selectColumns.js
+++ b/lib/list/selectColumns.js
@@ -5,11 +5,21 @@
  * @param {Array} columns
  */
 function selectColumns (q, cols) {
+	var list = this;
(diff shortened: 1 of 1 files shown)


morenoh149:
@creynders @pricco could you check if this looks right? I just inlined @pricco's patch logic into the `selectColumns` method

JedWatson:
@morenoh149 I'm about to merge the elemental branch and start working on 0.4 in master, which deprecates this function. Let's revisit and address this there, anything else is just pushing out that release.

morenoh149:
is there an open issue for that? should ref to it here.

PR #1604, opened 2015-08-22
enforce semicolons

morenoh149:
fixed syntax to enforce semicolons http://eslint.org/docs/rules/semi


JedWatson:
Thanks @morenoh149!

PR #1605, opened 2015-08-22
Fix datetime timezone bug

WofloW:
backend:
    Add a new hidden input to hold the UTC ISOString.
    Save this real time to database.

frontend:
html:
convert utc to local time
span.timeZone #{blog.publishedDate.toISOString()}
script.
    $( document ).ready(function() {
        $('.timeZone').each(function(index,data){
            var time = $(data);
            time.text(moment.utc($(data).text()).local().format('LLL'));
            time.css('visibility', 'visible');
        })
    });

css:
prevent showing the utc time
.timeZone {
  visibility: hidden;
}

### DIFF
--- a/fields/types/datetime/DatetimeField.js
+++ b/fields/types/datetime/DatetimeField.js
@@ -20,7 +20,8 @@ module.exports = Field.create({
 	getInitialState: function() {
 		return {
 			dateValue: this.props.value ? this.moment(this.props.value).format(this.dateInputFormat) : '',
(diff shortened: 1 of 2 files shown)


JedWatson:
Thanks @WofloW! I'm about to merge a massive set of changes into master as we prepare for 0.4, since this modifies the data structure we're storing in the database let's revisit this once that's in. Would be good to get in that release.

JedWatson:
Hey @WofloW, thanks again for this - we're going to take a very similar approach to this in 4.0 but I'm going to close the PR because we need to implement it in the new architecture.

PR #1614, opened 2015-08-25
Add S3 support to wysiwyg editor (0.4)

jrit:
This is the same as #1579 but for the v0.4 version.

@JedWatson if you don't want to bother with the v0.3 PR it can just be closed. I had to branch to make sure we are stable on v0.3 for now.


JedWatson:
@jrit I really like what you've done here, nice work :smile:

PR #1616, opened 2015-08-26
Fixed bug in Number type column view displaying blank cell for value of 0

tomasztunik:
Number fields with value of 0 would be treated as falsy value, ignored and not displayed as valid value of 0 but rather as a blank


JedWatson:
:+1: thanks @tomasztunik!

PR #1634, opened 2015-09-01
Don't use regular expressions for signin

markokocic:


markokocic:
Ok, after reading a tests seems like it's a feature to have case insensitive email lookup.

While I don't agree with that, I'll let the decision to project maintainers. I'll rework the patch to fix test if we are to drop the case insensitivity requirement. Otherwise, just reject the PR.

creynders:
@markokocic why exactly would you want to make it case-sensitive? AFAIK that's definitely exception rather than rule.

markokocic:
@creynders 
First impulse was to make `TingoDB` and `Tungus` driver work with `keystone` without need to modify keystone source, since those don't work with such regex as used here. Details here: https://github.com/sergeyksv/tingodb/issues/84

Only after running tests, I realised that it was actually a `keystone` feature to use case insensitive email as login. It's nothing wrong per se, just a bit confusing since it's the first time that I actually realised somone implemented it that way.

morenoh149:
I'm seeing over here http://stackoverflow.com/a/9808332/630752
One person would go as far as to say "you're unsafe to treat email-addresses as case-sensitive [in] manner".

creynders:
Seems to me this specific problem is rather an issue with tingodb than keystone, no?
The only other approach I can think of to achieve case insensitive logins is by lower casing the email addresses when they're stored to the DB (and when logging in), but that's not really an option IMO, since this would break almost every single keystone installation in the wild.

markokocic:
@creynders Yes, the original problem is spicific with `TingoDB`, and it should be fixed there.

What I don't understand is why `keystone` is trying to force case insensitive logins at all. By the spec, domain name is case insensitive, but first part of the email address, the one before `@` is not. There can be a difference between `MyName@domain.com` and `myName@domain.com`, it is not always the same email address, and it's up to email provider to treat it at will. Even gmail considers `name@gmail.com` and `nAmE+something@GMail.Com` and `n.a.me@gmail.com` as the same account, and will happily deliver all three emails to the same mailbox `name@gmail.com`. Should that provider specific case also be handled?

There's no way to actually validate email except by sending the test mail with confirmation link. Since here we are not talking about email alone, bus as an email as login key, we can as well check only if email address that is stored into db is the same as whatever user typed during login attempt. If I used `me@mail.com` to register myself to some random web site, I don't have any expectation that `ME@MaIl.CoM` will also work as a login.

creynders:
It's certainly up to discussion, but seriously, try any big name service and authentication is always case insensitive when it comes to usernames even when e-mail addresses are used.
I just tested it out with github, twitter and facebook

markokocic:
Since it's a common consensus that this works as intended, I'm closing this ticket.

PR #1648, opened 2015-09-04
Fixed double-escaping of Amazon S3 File URLs

jeffreypriebe:
Added unescaping of percent symbol to avoid double-escaping from Amazon S3.

If you have special characters in the filename or path that you send to Amazon, when the url comes back it will have them double-escaped (escaping the character then escaping the percent). Not sure if this is Amazon or Knox who is the cause, but unescaping the '%' will fix the URLs.


JedWatson:
Thanks @jeffreypriebe!

Issue #1615, opened 2015-08-26
Filters not working anymore?


SlashmanX:
It seems the latest version doesn't allow filters to work correctly.

Here's a screenshot from 0.2.42:

<img width="637" alt="screen shot 2015-08-26 at 15 34 39" src="https://cloud.githubusercontent.com/assets/871743/9496426/35e27014-4c08-11e5-8bd1-b7cffd2a2f0e.png">

And essentially the same in 0.3.14:

<img width="807" alt="screen shot 2015-08-26 at 15 35 12" src="https://cloud.githubusercontent.com/assets/871743/9496437/3f5b22d0-4c08-11e5-9b87-e7040c277ad7.png">

Note that 'Routes' can be selected even if Airline isn't selected, also when an airline is selected, Routes still shows all available routes, so it seems the filter is not doing anything at all.

Relevant code is as follows: 

```
Cycle.add({
	name: {
		type: Types.Text,
		trim: true
	},
	ID: {
		type: Types.Text,
		trim: true,
		required: true,
		initial: true
	},
	airline: {
		type: Types.Relationship,
		required: true,
		initial: true,
		ref: 'Airline'
	},
	route: {
		type: Types.Relationship,
		required: true,
		initial: true,
		ref: 'Route',
		filters: {
			airline: ':airline'
		}
	}
});
```

JedWatson:
Looks like this was something we've missed for a while in the 0.3.x line of releases. We're prepping a major re-work of that whole system in 0.4 right now so I'll add this to the list of things to confirm support for in the new version. Thanks for the report!

mxstbr:
Definitely fixed in 0.4!

PR #1661, opened 2015-09-09
Fixed relationship filters not working on _id field

Adam14Four:
At least a partial fix for issue #1615.  This allows the relationships filters to at least work with the _id field of the model.


morenoh149:
@Adam14Four looks like linting fails on an unrelated file

Adam14Four:
It looks like a fix was committed a little while ago...can we re-run the check?

morenoh149:
try rebasing onto master if you can. That could trigger another CI check

PR #1666, opened 2015-09-09
Allow negative number in NumberField

lojack:


morenoh149:
could you briefly elaborate on what was wrong on the front end. And what the new regex does?

lojack:
Prior to this change, any time you typed a character into the number field the regex would strip any characters that aren't `0-9` or a '.'. This means any time you typed a '-' it would be removed. This change simply includes a '-' in the regex so it can be typed into the field.

Essentially, there's sanitation occurring on the frontend with this regex, and validation occurring on the backend (that's being tested). A case can probably be made to remove this sanitation all together, but I do see the value.

creynders:
Could this get merged? Travis is failing because of totally unrelated code.
Ping: @JedWatson @jossmac

JedWatson:
Great! thanks @lojack, and @creynders for the review!

Issue #1663, opened 2015-09-09
Email helper fails when utilizing Nunjucks templates + Fix


chriseaton:
If you switch the email template engine to Nunjucks, the e-mail will fail to render. The reason is that the compile() function returns an object that is not a function so the following line fails:

lib/email.js Line 215
```
var html = template(locals);
```
I was able to overcome the problem for now by doing the following:
```javascript
var html = null;
var template = templateCache[this.templateName];
if (typeof template === 'function') {
    html = template(locals);
} else if (typeof template.render === 'function') {
    html = template.render(locals);
}
```

Performance-wise you could probably pre-cache the template rendering option on initialization, or set the rendering option as a "default" property of the Email object, but this is the essence of it

chriseaton:
So above is actually only a partial fix. It turns out the initial compile function call (lib/email.js Line 319)  causes issues as well due to the different argument arrangement. Nunjucks doesn't like the jade-style 2nd argument... I don't have a resolution at the moment other than break the e-mail compatibility w/ jade.

mxstbr:
We're closing all questions and support requests to keep the issue tracker unpolluted. Please ask this question on [Stackoverflow](https://stackoverflow.com/questions/tagged/keystonejs) or [Gitter](https://gitter.im/keystonejs/keystone) instead!

mxstbr:
The entire `Keystone.Email` class was rewritten as an external module, `keystone-email`. This should be fixed over there, if this issue still exists for you with the new code feel free to comment again!

PR #1670, opened 2015-09-10
Updated Email helper to better support custom template engines.

chriseaton:
Updated support for custom template engines in the Email helper. Specifically, by adding the capability to provide a custom compile and/or render call (via the new 'templateCompile' and 'templateRender' options). 

If either of these options (compile/render calls) are omitted, the Email helper will fallback to it's original methods (designed for Jade).

Reference Issue: [1663](https://github.com/keystonejs/keystone/issues/1663)

### DIFF
--- a/lib/email.js
+++ b/lib/email.js
@@ -158,8 +158,8 @@ var Email = function(options) {
 	this.templateExt = options.templateExt || Email.defaults.templateExt;
 	this.templateEngine = options.templateEngine || Email.defaults.templateEngine;
 	this.templateBasePath = options.templateBasePath || Email.defaults.templateBasePath;
-	this.templateCompile = options.templateCompile;
-	this.templateRender = options.templateRender;
+	this.templateCompile = options.templateCompile || Email.defaults.templateCompile;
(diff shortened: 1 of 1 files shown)


chriseaton:
I also noticed there is currently no documentation for the Email helper on the official site. I'd be willing to  help with that if desired.

chriseaton:
Here's an example of leveraging Nunjucks (template engine) for e-mail processing using this update.
```
var Email = new keystone.Email({
    templateName: 'contact',
    templateExt: 'html',
    templateEngine: require('nunjucks'),
    templateCompile: function(engine, contents, opts, path) {
        return engine.compile(contents, opts, path); //calling compile() on the nunjucks engine, and returning the compiled template
    },
    templateRender: function(template, locals) {
        return template.render(locals); //calling the nunjucks compiled template render()
    },
    mandrill: {
        //mandrill options here...
    },
    // Mandrill template
    templateMandrillName: null,
    templateForceHtml: false
});
```

JedWatson:
:+1: looks great, thanks @chriseaton

PR #1688, opened 2015-09-24
Show 'reverse' relationship in item details

davibe:


JedWatson:
:+1: neat, thanks @davibe!

Issue #1676, opened 2015-09-15
DateField always displays current date


lojack:
This is for 0.4. You can select a Date with the date selector, and it appears to save correctly to the database, but upon loading any page, the date displayed on the frontend is always the current date.

JedWatson:
Fixed by #1703

PR #1703, opened 2015-10-03
Fix date field issue #1676

w01fgang:
Hi, I just make a fix of Date and DateTime fields. issue #1676
And I have make a link to "front page" which is present on 0.3.x.
I forgot the number if issue.


JedWatson:
Thanks @w01fgang! Nice work :)

PR #1719, opened 2015-10-08
Incorrect user error fix #1697

### LINKED ISSUE #1697: Error: Incorrect user or password
I'm getting this stack:

````
------------------------------------------------
Error handling 500 (error):
[TypeError: res.err is not a function]

------------------------------------------------

Error thrown for request: /
Error: Incorrect user or password
    at /home/javiercastro/javascript/website/node_modules/keystone/lib/session.js:113:19
    at Query.<anonymous> (/home/javiercastro/javascript/website/node_modules/keystone/node_modules/mongoose/lib/query.js:2119:28)
    at /home/javiercastro/javascript/website/node_modules/keystone/node_modules/mongoose/node_modules/kareem/index.js:177:19
    at /home/javiercastro/javascript/website/node_modules/keystone/node_modules/mongoose/node_modules/kareem/index.js:109:16
    at doNTCallback0 (node.js:407:9)
    at process._tickCallback (node.js:336:13)
GET / 500 67.312 ms
````

I think it may be related to this: after many months, i've opened my Firefox browser (normally i use Chrome every day) and pointed to my dev keystone site. Then i get the stack.

I assume it is related to the old session cookie stored on Firefox which is now invalid (may be i've changed the secret after my last Firefox use on this site)

Which would be the way to avoid/correct this?

### PR DESCRIPTION
**Pull request by w01fgang**
Hello, I have made the fix of the issue #1697.
When DB is changed it is impossible to login. Throw error 500.
That fix clears cookies and redirects to login page without any errors.

### CODE REVIEW

File: lib/session.js [hunk 1]
```diff
@@ -110,7 +115,9 @@ var doSignin = function(lookup, req, res, onSuccess, onFail) {
 			if (user && (!passwordCheck || scmp(passwordCheck, hash(user.password)))) {
 				postHookedSigninWithUser(user, req, res, onSuccess, onFail);
 			} else {
-				onFail(err || new Error('Incorrect user or password'));
+				res.clearCookie('keystone.uid');
```
JedWatson: shouldn't this still call `onFail`?

w01fgang: It falls with error 500 if calls `onFail`. I didn't found function of update cookies when it expires.
Error occurs if the user has cookies from the old session. I think the better way is to delete it.
In this `if` statement `else` works only if `user` is defined. And if it defined `err` is null.

JedWatson:
Thanks for your work here @w01fgang. I ended up fixing it independently of your PR, but you definitely pointed me in the right direction. To see the final solution, check out 3dbbe46

PR #1750, opened 2015-10-27
Fixed typo when creating auth url locals

tmcgoo:
Couldn't log into admin when auth was true because signout url and signin url locals where missing '/keystone' in createDynamicRouter.js.


JedWatson:
Thanks @tmcgoo!

Issue #1612, opened 2015-08-25
Version 0.4.0


JedWatson:
Today, I'm really excited to be merging our work on the `elemental-integration` branch into master.

We're really close to releasing a major new version which will finish our transition to React for the Admin UI, and be a strong foundation for a lot we've been wanting to do with Keystone for a long time, including opening up the Admin UI for customisation, and making the project more modular.

Once this lands, I expect we'll move much more quickly through the next few minor releases (0.5, 0.6, ...) as we introduce new functionality and potentially some breaking changes while we refactor the project more thoroughly. Our 1.0.0 release will follow as soon as we're ready to lock down the customisation and other APIs for long term support.

This means that `master` may be unstable for a while as the new UI is finished, but getting the project into a state where it is easier for others to contribute to the effort, and for the broader community to engage with our progress is really important. Until we get through this push, PRs and issues will continue to come in against a version that we've really moved on from, and I don't want to waste any of the valuable time and effort that's being dedicated to this project.

The other change that will happen from today is we'll start creating issues to track our progress towards 0.4.0 more granularly, and that will make more sense when the primary working codebase is the `master`  branch. Up to now, @jossmac and I have worked largely behind the scenes on tasks that really only the two of us are across, and I'm keen to facilitate a much broader engagement with what's left to do. (we'll  be rearranging the milestones and issue tags a bit to better support this today as well)

There will be a new `0.3.14` release today with all the current work in `master` before the merge, and we'll create a `0.3.x` branch which we'll continue to release from if any critical issues pop up in that version.

Thanks everyone for your patience as we've been working towards this huge update, and for all your support and engagement over the last two years that Keystone has been out in the world. We're tremendously excited to get these changes we've been working on released, and move on to the next stage of Keystone's development.

Once this release settles down I'll be publishing a new roadmap for what's to come, and outline the vision we've got for the future of Keystone. In the meantime:

#### 0.4.0 TODO

This list will be replaced with individual issues as we add them, but generally the entire Admin UI and back-end support for it has been rewritten. This means many features won't be completely working, and there's a few things left to add. The major areas are:

- Finalising the new design (feedback please!)
- Implementing `react-router` for complete client-side routing (plus possibly a flux/etc implementation)
- Creating a new `items/update` API endpoint and using it to save item changes
- Finishing the Filters implementation across all the field types
- Finalising the React Fields

[...]

Issue #1315, opened 2015-04-17
upgrade to mongoose v4


creynders:
We need to determine how drastic the change is first though.

morenoh149:
https://github.com/Automattic/mongoose/wiki/4.0-Release-Notes

creynders:
Yeah, I'd seen the release notes already, but we need to determine how this impacts keystone.

morenoh149:
The notes sound intimidating. First step I'll take is to see what happens to the keystone demo site if I change this dependency. Will share results then.

creynders:
@morenoh149 yeah, the list of backward breaking changes is substantial. Let's hope it's easily fixed though

morenoh149:
I've deployed a demo with mongoose v4 https://fast-hollows-3420.herokuapp.com/keystone/ nothing seems broken. I also reviewed the release notes in depth and couldn't find any instances in keystone's codebase where it'd be affected.

I'd wait for further testing or until someone requests the new mongoose features before bumping keystone's dependencies.

dcousens:
So do we close this issue until then?

JedWatson:
I've added it to the 0.4 milestone - happy to do the upgrade when we publish the next release w/ other (potentially) breaking changes in it.

morenoh149:
@JedWatson if we do a test release for 0.4 similar to the 0.3 release there would be no need to do test release for just this feature. If that's the plan then this issue can be closed.

adamscybot:
(Y) yes to this. I desperately need the mongoose query hooks (pre/post) find so I can modify what keystone shows in list view (though may open another issue for this, this is a decent hack).

JedWatson:
@adamscybot heads up, the 0.4 release is still a week or two away, so if you need this urgently you should be able to use mongoose 4.x today by including it in your project's dependencies, then providing keystone with the updated version:

```
// in your keystone.js file
var keystone = require('keystone');
var mongoose = require('mongoose');
keystone.set('mongoose', mongoose);
// ...
```

as far as I know there are no incompatibilities between the new version and Keystone itself, we're just not bumping a major dependency on a patch version of keystone.

adamscybot:
Thanks Jed. I actually managed this through a (hacky) postinstall script on package.json

```
find node_modules -type d -name mongoose | xargs rm -fr; npm install mongoose@4.0.3
```

This is nice as it means I can keystone vanilla. Can give the all clear so far on mongoose v4 with keystone.

[...]

PR #1753, opened 2015-10-27
updated mongoose to version 4, fix #1752

### LINKED ISSUE #1752: 0.3 Error: Cannot find module '../build/Release/bson'
```bash
{ [Error: Cannot find module '../build/Release/bson'] code: 'MODULE_NOT_FOUND' }
js-bson: Failed to load c++ bson extension, using pure JS version
```

node: 4.2.1
npm: 2.14.7
keystone: 0.3.15

mahnunchik:
all tests passed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "marked": "^0.3.5",
     "method-override": "^2.3.5",
     "moment": "^2.10.6",
-    "mongoose": "~3.8.35",
+    "mongoose": "^4.2.3",
     "morgan": "^1.6.1",
     "mpromise": "^0.5.5",
     "multer": "^0.1.8",


nikashitsa:
:+1:

morenoh149:
I think we get mongoose 4 in the 0.4.x version of keystone. See https://github.com/keystonejs/keystone/issues/1612  Ref https://github.com/keystonejs/keystone/issues/1315

mahnunchik:
If mongoose 4 does not contain breaking changes related to keystone then why not to upgrade dependency in 0.3 branch while 0.4 branch is not so stable?

morenoh149:
we can definitely do that as well.

JedWatson:
I'm really mixed on this because while mongoose 4.x doesn't break _keystone_, it could quite easily break projects _build on_ keystone which is the point of more than a patch version bump. So merging this would definitely break sermver.

Having said that not releasing 0.4 is really holding us back and there's still work to do :disappointed:

creynders:
As I wrote in #1315 

>master is on v4 already and if you're still running KS v0.3, just add mongoose v4 as a dependency in your project and hook it up with keystone.init or keystone.set('mongoose', require('mongoose'))

Closing this, since it's landed for v0.4 and I don't think it's worth the hassle for v0.3

PR #1757, opened 2015-10-29
Moved initList middleware into its own file

tmcgoo:
Completed TODO in admin/server/app/createDynamicRouter.js.

InitList middleware had to be moved out into it's own file in the middleware directory.


JedWatson:
:+1: neat, thanks @tmcgoo!

PR #1766, opened 2015-11-01
Server Testing

snowkeeper:
Fixes path to `server/startSocketServer.js` in `start.js`

Change `keystone.httpsServer` to match `keystone.httpServer` by assigning to listen: `keystone.httpsServer = server.listen(port, host, ready)`

Add comments

Add simple tests for http, https, socket servers


snowkeeper:
The Travis CI build is calling each test successively without exiting and is failing since it calls the global Keystone instance again.  

Do I need to refactor or are we good?

JedWatson:
Thanks for this! We're good for now but let's work out a way to get the tests working. Fixing Keystone so it can return a new instance (instead of the default one) is one of the big things left to get this core refactor completed for 0.4.

PR #1775, opened 2015-11-06
WIP: allow creation of multiple keystone instances

creynders:
**WORK IN PROGRESS - DO NOT MERGE**

A first attempt at allowing the creation of multiple keystone instances. Biggest hurdle ATM:

>The problem is that for instance `lib.core.list` requires `index.js` which means it ​_always_​ refers to the singleton instance, i.e. if you create a new instance, it’ll look like it’s working OK, but in fact you’ll be working with one instance of keystone in your tests for instance, while `List` will still be referring to the singleton instance.
>In other words, we’ll have to eliminate ALL `require('../')` etc in ALL of the files and switch to DI through the constructor for example
>If I’m not mistaken there’s approximately 39 injection sites: https://github.com/keystonejs/keystone/search?utf8=%E2%9C%93&q=%22var+keystone%22&type=Code

I started out as we'd agreed on slack: moving the whole Keystone class to a `prototype.js`. I also added a `registry.js` which functions as a factory and service locator.

@snowkeeper Feel free to comment/add/tear apart


JedWatson:
Note: this has been merged into a new `prototype` branch in the main keystone repo so we can work on it together, going to close this PR to avoid confusion

PR #1778, opened 2015-11-09
Remove dependency on underscore

jonathanp:


JedWatson:
Thanks @jonathanp! :+1:

PR #1779, opened 2015-11-10
Refactor admin UI dashboard icons class generation

jonathanp:
I DRYed up a block of code and got rid of a bunch of unneeded constants.


JedWatson:
Nice @jonathanp, thanks!

PR #1781, opened 2015-11-10
Align require calls

jonathanp:
Align the require calls by colons in `columns.js` to match the style in `fields.js` and `filters.js`


JedWatson:
Cheers!

morenoh149:
should we add this rule to the linter? http://eslint.org/docs/rules/key-spacing.html#3-vertically-align-colons-align-colon

snowkeeper:
Thats a lot of work for no true benefit.  Everytime something long is added the entire block would have to be redone.

jonathanp:
@snowkeeper the benefit is increased readability as it's much easier to skim over, and increased consistency as this style is already used in `fields.js` and `filters.js` for similar blocks of `require` statements.

Besides, whatever editor you're using it probably isn't too hard to set up a "align by first colon" shortcut which is useful if you are to add a longer name to the block.

JedWatson:
@jonathanp good tip with the shortcut.

I don't think we should add this to the linter as it only provides value in _long_ code blocks; I don't want to do it for every multiline object, but in files like this I think it's worth doing.

Maybe, for clarity, we should enable the linter rule on a per-file basis (or around code blocks) so it's clear for contributors when the convention should be followed, but I am also happy to just keep an eye on it.

JedWatson:
On a related note, I don't know a great way to express "provisional" standards... if anybody has a suggestion for this I'm open to it.

In other words I'm not ready to commit to lining up code blocks everywhere but I'm open to experimenting with it in a few files where, like @jonathanp says, it improves readability.

I've also started introducing trailing commas for multi-line arrays and objects to "see how it feels" and am liking it. Now that all our code is either being run in node or transpiled by Babel it's not going to cause any issues. And as a benefit, it makes sorting and managing lines _much_ less typing overhead, especially with long lists like this file or React PropTypes / state / etc.

So we kind of need to turn of any rules relating to that in our linter until there's an "official adoption" and we're ready to go refactor a bunch of code... in the meantime I'm introducing them as I work in a file.

Same with var statement blocks; I've been removing any `var one, two...` and putting "one var per line" as I go. But this could be confusing, especially for new contributors, or make it look like we don't follow standards so I'm honestly not sure what the best approach is for experimental use (or transitions to new conventions).

jonathanp:
I'm all for adding both "one var per line" and "require dangling commas" to the linting rules. They both makes is easier to add/remove/reorder rows in arrays and objects and as you say doesn't break in any environments anymore.

Regarding how to best transition to new conventions; why not just edit `.eslintrc` and let the new rules trigger errors all over the place? Whenever you work in an old file that uses the old conventions you probably feel like cleaning it up to get rid of the linting errors, and after a while the whole code base has been transitioned. ;)

I think I prefer that than to have different conventions in different files and they all pass the linting rules.

PR #1782, opened 2015-11-10
Remove redundant `Object.assign()` polyfill

jonathanp:
Since [`babel-plugin-object-assign`](https://www.npmjs.com/package/babel-plugin-object-assign) is already used in the codebase [this polyfill](https://www.npmjs.com/package/object-assign) is redundant.


jonathanp:
Oops, didn't think about that this polyfill is needed for old Node versions, should I close the PR?

JedWatson:
Correct

PR #1783, opened 2015-11-10
Replace `_.extend` with `Object.assign`

jonathanp:
To increase consistency I have replaced the client side uses of `_.extend` with `Object.assign`.


JedWatson:
:+1: great

Issue #1723, opened 2015-10-09
0.4:  Using the + in the Admin UI Home Page to create does not work


webteckie:
Using the + sign next to each item type as in the following screenshot does not work:

![image](https://cloud.githubusercontent.com/assets/5872515/10383501/ac7c01b0-6dfd-11e5-923c-c08a03e3fbce.png)

It goes to an empty page:

![image](https://cloud.githubusercontent.com/assets/5872515/10383505/c7450e74-6dfd-11e5-9471-8a0d1891647b.png)

OS:  Windows 10
Browser: Chrome

jonathanp:
This has been resolved, thanks for reporting the issue @webteckie!

PR #1784, opened 2015-11-10
Check if param value is defined in filtersFromUrlParams (#1723)

kumo:
The '+' links in the Admin UI home page (#1723) weren't working because the ``filtersFromUrlParams`` function in CurrentListStore.js had an undefined value for the search param ``?create``


JedWatson:
Thanks @kumo

PR #1788, opened 2015-11-10
Remove unused code for search input visibility toggling

jonathanp:
Since the toggling of the visibility of the search input field is (very elegantly) handled by CSS this commit removes some no longer needed code.


JedWatson:
You're on fire... to save a bunch of "thanks @jonathanp" I'm just going to start silently merging the simple ones :)

Issue #1711, opened 2015-10-07
0.4:  Creating a new user throws error


webteckie:
![image](https://cloud.githubusercontent.com/assets/5872515/10326628/775e62a4-6c6b-11e5-9dd8-d7df0cfcccb0.png)

The following input reproduces the issue:
![image](https://cloud.githubusercontent.com/assets/5872515/10326859/26ef95c4-6c6e-11e5-8b89-9b96cb626940.png)

OS:  Windows 10
Browser: Chrome

JedWatson:
Fixed by @kumo's PR

PR #1789, opened 2015-11-10
Using error keys to map error messages (#1711)

kumo:
The errors from the modal ``CreateForm.js`` are stored in an Object rather than an Array, so the map function wasn't working. This was causing the error in #1711


JedWatson:
Thanks @kumo

PR #1790, opened 2015-11-10
Focus the focus target in the create form modal when opening the modal

jonathanp:
This PR fixes a bug that caused the focus target in the modal in `CreateForm` to not be focused when the modal was opened.

screen shot 2015-11-10 at 10 16 17

Another approach (probably better) would be to submit a PR to [Elemental UI](https://github.com/elementalui/elemental/blob/master/src/components/Modal.js) and refactor the `Modal` component there to use the low level API of `ReactTransitionGroup` instead of `ReactCSSTransitionGroup`, which has a lifecycle method `componentDidEnter` that could be used for the focusing. [More info on how this API could be used.](https://facebook.github.io/react/docs/animation.html)

### DIFF
--- a/admin/client/components/CreateForm.js
+++ b/admin/client/components/CreateForm.js
@@ -41,9 +41,9 @@ var CreateForm = React.createClass({
 	componentDidUpdate (prevProps, prevState) {
 		if (this.props.isOpen !== prevProps.isOpen) {
 			document.body.style.overflow = (this.props.isOpen) ? 'hidden' : '';
-			if (this.refs.focusTarget) {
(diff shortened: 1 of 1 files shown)


JedWatson:
A PR for elemental implementing that would be very welcome, in the meantime I'll merge this because it's much nicer to use in Keystone.

JedWatson:
p.s. not sure if you've seen it but another more elegant solution here might be to use the `autofocus` prop on Elemental FormInput components. We could add it to the Field components and pass it through... let me know if you want to look into that.

jonathanp:
Good call, just fixed it in #1799

PR #1791, opened 2015-11-10
Blur search input field on escape key press

jonathanp:


admin/client/components/EditFormHeader.js
```diff
@@ -46,6 +46,14 @@ var Header = React.createClass({
 		});
 	},
 
+	handleEscapeKey (event) {
+		const escapeKeyCode = 27;
+
+		if (event.which === escapeKeyCode) {
+			this.refs.searchField.getDOMNode().blur();
```
JedWatson: We're using React .14, I believe this will throw a warning to use `ReactDOM.findDOMNode(ref)` instead. To see warnings, rebuild keystone's packages with `npm run build-dev`

jonathanp: Correct, I've replaced that call now.

JedWatson:
eek, conflicts! mind re-syncing this?

jonathanp:
Np, it's re-synced :)

JedWatson:
Sweet!

PR #1795, opened 2015-11-10
Removed unused variables

jonathanp:
I've gone through the components and removed all unused variables I could find.

### DIFF
--- a/fields/mixins/ArrayField.js
+++ b/fields/mixins/ArrayField.js
@@ -1,3 +1,4 @@
+var React = require('react'); // this is needed since we use JSX, even though the `React` name is not used in our code
 var _ = require('underscore');
 
 var Button = require('elemental').Button;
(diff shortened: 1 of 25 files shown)


fields/mixins/ArrayField.js
```diff
@@ -1,4 +1,3 @@
-var React = require('react');
```
JedWatson: React will get used when the JSX is transpired... there are a few of these.

JedWatson:
Awesome, thank you! I wonder why our linter isn't complaining about these. I've noticed it missed quite a few things and suspect our config may not be correct (maybe related to an ESLint update...?)

I think there are a couple of errors in here because `React` is used when JSX is transpiled, but aside from that this looks great!

jonathanp:
Good catch! I've re-added those `var React = require('react')`'s now, except in `admin/client/columns.js` where there's no JSX.

JedWatson:
:+1:

PR #1800, opened 2015-11-11
Use `_.extend` instead of `Object.assign` in Node code

jonathanp:
This PR uses `_.extend` instead of `Object.assign` in Node code to support environments without ES2015 support.


JedWatson:
good catch, cheers!

PR #1801, opened 2015-11-11
Remove direct mutation of state and props in React components

jonathanp:
This PR removes direct mutation of state and props in a couple of components.

[Some info on why this is bad can be found here.](https://medium.com/@cassiozen/a-brief-talk-about-immutability-and-react-s-helpers-70919ab8ae7c)


JedWatson:
Great, been meaning to do this - thanks!

By the way, in environments with ...spread support it can also be more simply written:

```
let values = { ...this.props.values };
```

PR #1802, opened 2015-11-13
Add drag/drop via react-dnd

### LINKED ISSUE #1633: Implement Drag & Drop Reordering in React
For 0.4.0, we need to reimplement list sorting with drag and drop in React. Not sure if there's an appropriate library we can implement; the last (jQuery) implementation was quite simple and specific because we're sorting `table` rows rather than `div` elements. It should be pretty easy to port.

The Component to start with is `/admin/src/components/ItemsTable`; it may make sense to break out each `tr` into its own Component before too long.

We'll need a new sort endpoint as well in the API, I want to remove the old (`admin/api/list.js`) handler completely, the new endpoint could be significantly based on the code in the `order` case statement in that file.

snowkeeper:
Fixes #1633 

Used http://gaearon.github.io/react-dnd/ to add in drag drop implementation with React.  

Able to be futher configured to deal with drags across pages.


JedWatson:
Looks great @snowkeeper! Thanks. Would you mind rebasing? I'm getting conflicts on the merge (probably due to the other PR I just merged that replaces `require` with `import` in the client-side scripts)

Also a couple of syntax quirks that I'll mark up inline.

admin/client/components/ItemsTable.js
```diff
@@ -1,20 +1,149 @@
...
+
 
 const CONTROL_COLUMN_WIDTH = 26;  // icon + padding
 
-var ItemsTable = React.createClass({
+/**
+ * Implements drag source.
+ */
+const itemSource = {
+	beginDrag(props) {
```
JedWatson: function names and argument definitions should have a space between them, i.e

`beginDrag (props) {`

admin/client/stores/CurrentListStore.js
```diff
@@ -17,50 +18,59 @@ var active = {
...
+		fail: false
+	};
+}
+
 var CurrentListStore = new Store({
 	getList () {
 		return _list;
 	},
-	getAvailableColumns () {
+	getAvailableColumns() {
```
JedWatson: Mind undoing these whitespace changes please? as per above, trying to get consistency.

admin/client/components/ItemsTable.js
```diff
@@ -1,20 +1,149 @@
...
+		
+		var style = this.props.isDragging ? {
+			borderLeft: '2px solid #1385e5',
+			color: 'black',
+			backgroundColor: 'white',
+			cursor: 'move',
+			opacity: 0
+		} : {};		
+		return (
+			this.props.connectDragPreview(this.props.connectDropTarget(<tr style={{ ...style }} key={'i' + item.id} onClick={this.props.manageMode ? (e) => this.props.checkTableItem(item, e) : null} className={rowClassname}>{cells}</tr>))
```
JedWatson: we probably only need to do this when `this.props.isDragging`? there may be some other calls in this file that can be similarly optimised. Not sure that I'm correct though, don't have a lot of experience with react-dnd (yet)

admin/client/stores/CurrentListStore.js
snowkeeper: no problem, my bads

admin/client/components/ItemsTable.js
snowkeeper: Agree.  I have an update with the row creation moved into its own component.  I will run through some changes and see what happens.

PR #1803, opened 2015-11-13
ES2015 module import syntax for loading dependencies

jonathanp:
The way dependencies were loaded in the client side JS was a bit fragmented so to increase the consistency I've replaced the `require()` calls with the ES2015 `import` syntax.

I'm fine with either syntax but I think it should be consistent, and between the two I think the ES2015 syntax feels a bit more modern (and future proof). :)


JedWatson:
Thanks @jonathanp. Glad to have this cleaned up.

PR #1809, opened 2015-11-13
Remove dependency on underscore

jonathanp:
This PR removes the underscore dependency in a couple of files by replacing `_.findWhere` with `[].find`


JedWatson:
Thanks!

PR #1815, opened 2015-11-15
Single page drag and drop

snowkeeper:
Drag and drop rows on the current page

PR #1816, opened 2015-11-15
drag drop across pages

snowkeeper:
Can replace previous PR.  
Adds the ability to drag items across pages.
Adds `perPage` list configuration option.
#### Drag single page
dnd1
#### Drag across pages
dnd3
#### Drop onto a page
dnd2

### DIFF
--- a/admin/client/components/ItemsTable.js
+++ b/admin/client/components/ItemsTable.js
@@ -1,19 +1,20 @@
 import React from 'react';
+import classnames from 'classnames';
+
 import Columns from '../columns';
 import CurrentListStore from '../stores/CurrentListStore';
 import ListControl from './ListControl';
+import TableRow from './ItemsTableRow'
+import DrapDrop from './ItemsTableDragDrop'
 
 const CONTROL_COLUMN_WIDTH = 26;  // icon + padding
 
-var ItemsTable = React.createClass({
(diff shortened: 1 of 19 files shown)


JedWatson:
@snowkeeper looks great, could you rebase though? there are conflicts, probably from recent PR merges. Makes sense that this would replace #1815 so maybe close that if it's not needed anymore?

jonathanp:
Looks really good @snowkeeper!

snowkeeper:
Thanks @jonathanp !

This should be ready to merge @JedWatson 

How do we get travis to pass checks again?

snowkeeper:
We used to have a `perPage` option per the docs.  I assume it got removed at some point.  The docs have the default at **50**.  We have the current set to **100**.  Which one should we use?

snowkeeper:
When testing I found that Microsoft Edge aggressively cached everything, including links not yet clicked.

 So changing order worked correctly, but when you switched pages the initial results were **always** displayed.  Using F12 you can verify that the correct list info is being fetched, but not applied.

You can load page 1.  In another browser make a change to an item from page 2.  Back in Edge click on page 2 and the results are stale, from the initial page load.  Again F12 shows that the correct info is fetched, but not applied.

PR #1818, opened 2015-11-15
Fix z-index issue in `Select` component in `UpdateForm`

### LINKED ISSUE #1814: Modal footer covers select
When select field is used on modal form, the footer of the form covers select field options.
To fix that behavior I need to change z-index from 1 to 2 on selector .Select-menu-outer in keystone.min.css
But I can't find this selector :( It is only in minified version.

jonathanp:
This fixes #1814


JedWatson:
Nice, thanks :)

PR #1820, opened 2015-11-15
Support for using a custom admin path

jonathanp:
This PR adds support for using a custom path to the admin UI, as discussed in #627.

If I would want to to use e.g. `/foobar` instead of `/keystone` as the path to the admin UI I would need to add the `admin path` config parameter to the `keystone.init()` call, as such:

```
keystone.init({
	'admin path': 'foobar'
});
```

I think this PR needs to be thoroughly reviewed as there's alot of files involved, but it should be feature complete AFAICT.

### DIFF
--- a/admin/client/components/CreateForm.js
+++ b/admin/client/components/CreateForm.js
@@ -98,7 +98,7 @@ var CreateForm = React.createClass({
 		if (!this.props.isOpen) return;
 		var form = [];
 		var list = this.props.list;
-		var formAction = this.props.adminPath + '/' + list.path;
+		var formAction = `${this.props.adminPath}/${list.path}`;
 		var nameField = this.props.list.nameField;
 		var focusRef;
 
--- a/admin/client/components/CreateForm.js
+++ b/admin/client/components/CreateForm.js
@@ -98,7 +98,7 @@ var CreateForm = React.createClass({
 		if (!this.props.isOpen) return;
 		var form = [];
 		var list = this.props.list;
-		var formAction = '/keystone/' + list.path;
+		var formAction = this.props.adminPath + '/' + list.path;
 		var nameField = this.props.list.nameField;
 		var focusRef;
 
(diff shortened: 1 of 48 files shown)


celiao:
Is the documentation updated too?

jonathanp:
@celiao Not yet

jonathanp:
@JedWatson have you had any chance to look at this yet? :)

JedWatson:
Sorry for the delay @jonathanp - have been travelling a lot lately & wanted some focus time to review this.

Looks great :+1:

The only thing I might change about it is passing the `adminPath` prop through components - seems like the code structure would be cleaner (and easier to refactor) if we referred to the `Keystone` global inside the components, rather than passing the prop down through the component hierarchy?

If you think it's better / clearer with props, I'm happy to merge as-is.

There are also some conflicts, I think it's just the changes to the links in the footer, would you mind updating your branch? Thanks!

jonathanp:
Thanks, it's rebased now!

I agree, the passing down of `adminPath` through the component tree feels a bit weird but I think I prefer that than to reference a global from inside of the components. I think it's good to keep them as pure (only dependent on their own props) as possible.

JedWatson:
Oh man, sorry about this - conflicts again. Master is moving fast at the moment... can I get one more rebase? I'll merge as soon as it lands so we break the cycle :)

jonathanp:
Haha np, rebased now!

JedWatson:
...and we're done!

Thanks :)

jonathanp:
:boom:

ipengyo:
this pr  Currently in which version is valid ？
@joerter

morenoh149:
@ipengyo this was merged into master. So it's probably available in the 0.4.x line.

tomheathershub:
Is it possible to have no admin slug or in other words have keystone run off the base url of the domain name ? e.g tomscms.co.uk/signin

wmertens:
@tomheathershub try it, there's a good chance it will work.

flzozaya:
Since this is now implemented as an option for `keystone.init();`, maybe it should be added to the docs?? 😄 . 
There is no mention regarding this anywhere on the Keystone Documentation page.

PR #1821, opened 2015-11-16
correct popout body max-height

snowkeeper:
The popout body height of **360px** was pushing the footer into hidden overflow for `Types.Date`.  This sets correct max-height for the Type filter options page.

Current | Fixed
-----------|--------
bad-filterheight | fixed-filerterheight


JedWatson:
Thanks for getting to the bottom of that @snowkeeper

PR #1823, opened 2015-11-16
DateType Filter

snowkeeper:
Adds filter for Date field type


snowkeeper:
I have datetime coming later today

JedWatson:
Nice work @snowkeeper!

PR #1825, opened 2015-11-16
Remove redundant `module.exports`/`exports`

jonathanp:
This PR removes redundant `module.exports`/`exports` statements.

If these have any meaning that I'm not aware of please enlighten me. :)


JedWatson:
The reason we reassign `exports` when reassigning `module.exports` is so that any references to `exports.x` in the file won't be mistakenly pointed at the previous `module.exports` object It's really just a safety precaution.

However, I think I'm OK with this change. Let's go with it.

PR #1826, opened 2015-11-17
Datetime UI Filter

snowkeeper:
#### ui filter
dtbetween

The popout doesn't currently resize when the content gets bigger.  This adds a forced minimum until a better solution.

### DIFF
--- a/admin/client/components/ListFiltersAdd.js
+++ b/admin/client/components/ListFiltersAdd.js
@@ -15,7 +15,8 @@ var ListFiltersAdd = React.createClass({
 	displayName: 'ListFiltersAdd',
 	getDefaultProps () {
 		return {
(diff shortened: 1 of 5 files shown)


snowkeeper:
Added an `onlyTime` field option to **DateTime**.  Displays the time option only.  

Switched time field to selects.

JedWatson:
@snowkeeper thanks! I think we're better off creating a new `time` field though, rather than having `onlyTime` as an option on a timestamp; otherwise there's unmanageable data in the database, and it's a bit misleading. A new `Time` could simply store its value as a number in 24h time.

snowkeeper:
I was just about to say the same.  Working with it was difficult at best.  I'll fix this pr.

Issue #1827, opened 2015-11-17
react-day-picker uses "stage":0  in .babelrc


megheaiulian:
There is backward-incompatible change from Babel 5.x to 6 as stage option is replaced with presets.
This breaks the admin section because fields.js,list.js,item.js are not compiling.

JedWatson:
Hmm, any idea on how we are supposed to stop babel from transpiling anything in the `node_modules` folder? that seems like a really silly thing for it to be doing automatically.

Aside from that, it's odd you're seeing this as I can't reproduce any issues compiling `fields.js`, `list.js` or `item.js`.

megheaiulian:
@JedWatson I am using node v5.0.0 and npm 3.3.6 on Windows 10. This version of node and npm seems to place node_modules a bit different. e.g. react-day-picker is not place at "project"/node_modules/keystone/node_modules/react-day-picker but at "project"/node_modules/react-day-picker. Because of this maybe browserify runs react-day-picker thru babelify transform and triggers the error.

megheaiulian:
Here is a screenshot of the error.
https://db.tt/JnH5dZw9

morenoh149:
@megheaiulian I get 404 on that link

megheaiulian:
here is a update screenshot
https://db.tt/0qZxPLkS

megheaiulian:
Has anyone encounter this issue? It seems to be a problem with browserify. I think transforms shouldn't be applied to dependency modules normally unless they are global transforms?

jacargentina:
I've got that error when updated my own fork with today code from master; i've reverted as it breaks the admin ui

megheaiulian:
@jacargentina 
For a quick fix you can try to add a postinstall.js:
```javascript
var fs = require('fs')
	, path = require('path')
	, dir = path.dirname(require.resolve('react-day-picker'));

fs.unlink(dir+'/.babelrc',function(err){
	if(err) console.error(err.message);
	console.log("Deleted .babelrc");
});
```
and change scripts section in npm to something like:
```json
"scripts": {
    "start": "node keystone.js",
     "postinstall": "node postinstall.js"
},
```
This way you can still work with the latest version until the issue is solved.

JedWatson:
I still can't replicate this, but `react-day-picker` is now bundled in the common packages and should be ignored by the dynamic browserify bundles. @megheaiulian and/or @jacargentina can you try the latest code in master and let me know if it's still an issue for you?

PR #1828, opened 2015-11-17
Ignore react-day-picker from babelify since it is already compiled

megheaiulian:
#1827


megheaiulian:
It seems ignore is checked after options in .babelrc so this fixes nothing.

PR #1830, opened 2015-11-17
UI Filter list only filterable items

snowkeeper:


JedWatson:
@snowkeeper thanks for the PR! this wasn't going to work as-is, other parts of the UI depend on that method (unrelated to filtering). I've implemented a new `getAvailableFilters` method on the CurrentListStore, you might be interested in reviewing the commit (and also the syntax)

PR #1838, opened 2015-11-18
Replace `function`'s with arrow functions

jonathanp:


JedWatson:
This is great, thanks!

I'm not 100% sold on inlining the `showCreateForm` and `wasSignedOut` functions, imo it's clearer if they're defined above even though they're only used in one place, but happy to go with it. These will disappear soon anyway when react-router is used.

Issue #1849, opened 2015-11-21
Select menu should have higher z-index? 


terenzeyuen:
![screen shot 2015-11-21 at 18 50 20](https://cloud.githubusercontent.com/assets/2553680/11319932/e94ec78a-9080-11e5-97db-95b555312433.png)

jossmac:
Resolved by #1850

PR #1850, opened 2015-11-21
Fix to issue select menu being overlapped

terenzeyuen:
https://github.com/keystonejs/keystone/issues/1849

screen shot 2015-11-21 at 18 50 20
screen shot 2015-11-21 at 18 53 06


JedWatson:
Thanks @terenzeyuen!

PR #1862, opened 2015-11-24
delete item fixes #1856

### LINKED ISSUE #1856: [0.4] Deleting from the item view goes to an empty page
image

snowkeeper:
This adds back the same code from 0.3.x.  Not sure if you were working on something new.

### DIFF
--- a/admin/server/routes/list.js
+++ b/admin/server/routes/list.js
@@ -46,6 +46,30 @@ module.exports = function(req, res) {
 			}
 		});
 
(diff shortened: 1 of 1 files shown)


JedWatson:
Planning to do this client-side, working to get entirely away from processing code in the routes so we can go entirely SPA.

The thing we need to do here is use the `DeleteItems` method on the `List` class, as per the `items` view.

snowkeeper:
Makes sense.  Are you working on it or would you like me to have a look?

JedWatson:
Thanks, just jumped in and did it :)

PR #1876, opened 2015-11-25
Markdown fields: added posibility of configure marked options, fixed #1798

### LINKED ISSUE #1798: Markdown Type: How to limit to safe tags only?
Is there a way to allow only certain tags to be used in a Markdown field?

I would like to allow users to use common tags like `a` but for example not `script`.

Thank you!

Edit: Looks like [this is where](https://github.com/keystonejs/keystone/blob/v0.3.x/fields/types/markdown/MarkdownType.js#L66) keystone transforms markdown input to html and I would like to enable the `sanitize` option from [marked](https://github.com/chjj/marked) as described in their readme.

jacargentina:
This adds the feature of adding `markedOptions` to any markdown field.

Im my case, i needed it so i could do this:

````javascript
  content: {
    type: Types.Markdown,
    markedOptions: {
      gfm: true,
      highlight: function(code, lang) {
        var result = highlight.highlight(lang, code);
        return '<pre class="hljs"><code>' + result.value + '</code></pre>';
      }
    },
    height: 600
  },
````

I'm using https://github.com/isagalaev/highlight.js to format my custom languages

@creynders I've head you! Hope this gets merged soon. Cheers.

### DIFF
--- a/fields/types/markdown/MarkdownType.js
+++ b/fields/types/markdown/MarkdownType.js
@@ -20,12 +20,13 @@ function markdown(list, path, options) {
 	options.nofilter = true;
 
 	this.toolbarOptions = options.toolbarOptions || {};
+	this.markedOptions = options.markedOptions || {};
 	this.height = options.height || 90;
 
 	// since wysiwyg option can be falsey this needs to use `in` instead of ||
 	this.wysiwyg = ('wysiwyg' in options) ? options.wysiwyg : true;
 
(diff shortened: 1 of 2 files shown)


jacargentina:
Commenting for cross-referencing #1798

jacargentina:
I should note, #1798 refers to v0.3; but i'm using v0.4

jacargentina:
@JedWatson please take a look

JedWatson:
Looks great @jacargentina! thanks :)

PR #1889, opened 2015-11-30
Move the confirmation dialog from `EditForm` to own reusable component

jonathanp:
I refactored the `EditForm` component a bit by moving out the confirmation dialog to an own component so we're able to reuse it in other places and have a better [SOC](https://en.wikipedia.org/wiki/Separation_of_concerns).

### DIFF
--- a/admin/client/components/ConfirmationDialog.js
+++ b/admin/client/components/ConfirmationDialog.js
@@ -1,0 +1,45 @@
+import React from 'react';
+import { Modal, ModalBody, ModalFooter, Button } from 'elemental';
+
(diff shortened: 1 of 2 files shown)


adnasa:
nice :+1:

adnasa:
Can you please add support where the `onConfirmation` button can be displayed as a *primary* button rather than always *danger*?

Not every time that an action taken that requires a confirmation is dangerous.

JedWatson:
:+1: nice, @jonathanp 

Also agree with @adnasa, although this can be a separate PR, maybe we introduce a `confirmationType` prop?

PR #1890, opened 2015-11-30
Removed redudant code

adnasa:
```
List.prototype.deleteItem = function (itemId, callback) {
	this.deleteItems([itemId], callback);
};
```

is a redundant implementation, since this declaration already existed on [L161](https://github.com/keystonejs/keystone/blob/master/admin/client/lib/List.js#L161)


JedWatson:
Oops! Thanks @adnasa :)

PR #1891, opened 2015-11-30
Extending the store with notify-error in attempt to follow the patter…

adnasa:
Extending the store with notify-error in attempt to follow the patterns implemented in store-prototype.

## Changes

- `deleteItem` calls `deleteItems` instead having to deal with a single item explicitly (taken from `admin/lib/List.js`

--- 

## onError

- The idea here is to try to continue on the pattern that is implemented with `store-prototype` and go on from there instead of having to register callbacks from the components that consumes this store.

---

## example

```js
componentDidMount () {
  CurrentListStore.addChangeListener(this.updateStateFromStore);
  CurrentListStore.onError(function(err) {
    // oh snap!
  });
}
```

---

There is one thing I'm not sure of: that is whether `onError` should return the index that the callback is registered on.

### DIFF
--- a/admin/client/stores/CurrentListStore.js
+++ b/admin/client/stores/CurrentListStore.js
@@ -11,6 +11,7 @@ let _location = null;
 let _ready = false;
 let _loading = false;
 let _items = {};
+let _onErrorHandlers = [];
 
 const _list = new List(Keystone.list);
 
@@ -135,6 +136,11 @@ const CurrentListStore = new Store({
 			columns: active.columns,
 			page: page
 		}, (err, items) => {
+
+			if (err) {
+				this.notifyError(err);
+			}
+
 			_loading = false;
 			// TODO: graceful error handling
 			if (items) {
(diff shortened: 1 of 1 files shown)


admin/client/stores/CurrentListStore.js
```diff
@@ -148,15 +154,16 @@ const CurrentListStore = new Store({
...
-			this.loadItems();
-		});
+		this.deleteItems([itemId]);
 	},
 	deleteItems (itemIds) {
 		_list.deleteItems(itemIds, (err, data) => {
-			// TODO: graceful error handling
-			this.loadItems();
+			if (err) {
+				this.notifyError(err.message || `Failed to delete item(s) of type ${_list.path}`);
```
JedWatson: use the `plural` util here for a nicer message

JedWatson:
Thanks for this @adnasa - we've ended up replacing the store with Redux, so this won't make it in, really appreciate it though.

Issue #1904, opened 2015-12-03
04: deleting item throws error


webteckie:
It happens both when deleting the item from the list view as well as from within the item view.

```
Uncaught ReferenceError: item is not defined
List.deleteItems @ item.js:3087
List.deleteItem @ item.js:3083
handleDelete @ item.js:665
ReactErrorUtils.invokeGuardedCallback @ packages.js:23475
executeDispatch @ packages.js:16248
executeDispatchesInOrder @ packages.js:16268
executeDispatchesAndRelease @ packages.js:15707
executeDispatchesAndReleaseTopLevel @ packages.js:15718
forEachAccumulated @ packages.js:30373
EventPluginHub.processEventQueue @ packages.js:15923
runEventQueueInBatch @ packages.js:23501
ReactEventEmitterMixin.handleTopLevel @ packages.js:23517
handleTopLevelWithoutPath @ packages.js:23615
handleTopLevelImpl @ packages.js:23595
Mixin.perform @ packages.js:29844
ReactDefaultBatchingStrategy.batchedUpdates @ packages.js:22224
batchedUpdates @ packages.js:27914
ReactEventListener.dispatchEvent @ packages.js:23726
```

![image](https://cloud.githubusercontent.com/assets/5872515/11550912/654acbda-9942-11e5-9896-3b624cad1501.png)

JedWatson:
Fixed by #1905

PR #1905, opened 2015-12-03
fix for the error when deleting items #1904

thomasjonas:
Fix for deleting one or multiple items for issue #1904.


JedWatson:
Right, thanks @thomasjonas!

morenoh149:
@thomasjonas does this PR close #1904 ?

ericelliott:
Are there unit tests for `.deleteItems()`?

thomasjonas:
@ericelliott I don't think so. At least I couldn't find any. I actually couldn't find any tests for /admin/client/lib/List.js and I'm also not sure on how to implement them.

PR #1909, opened 2015-12-04
Using es6 interpolation instead of string concatenation

adnasa:
`'(' + something + ')'` has become harder to read lately.. so...

More es6 interpolation.


JedWatson:
:grinning:

PR #1914, opened 2015-12-04
Add test to guidelines

ericelliott:
We really need to improve test coverage. Let's make sure the test requirement is in the guidelines. =)


ericelliott:
@JedWatson

JedWatson:
:+1:

With this in mind, I acknowledge it's important to expand the skeleton of the test suite so actually writing tests is possible; for many things it's not at the moment, unfortunately.

PR #1938, opened 2015-12-08
Use entrance/exit animations for the confirmation dialog

jonathanp:
Use the same animations as in the `CreateForm` modal


JedWatson:
Nice @jonathanp! :grinning:

PR #1947, opened 2015-12-09
Fix  LocalFiles click on button "remove" do not remove the file in React

Rmannn:
The state of items must be reset to update children components.

### DIFF
--- a/fields/types/localfiles/LocalFilesType.js
+++ b/fields/types/localfiles/LocalFilesType.js
@@ -116,13 +116,13 @@ localfiles.prototype.addToSchema = function() {
 		if (typeof element_id === 'undefined') {
 			value = values[0];
 		} else {
-			value = values.find(function (val) { return val._id === element_id; });
+			value = values.find(function (val) { return val._id == element_id; });
 		}
 
 		if (typeof value === 'undefined') {
 			return false;
 		}
-
+		
 		var filepaths = value.path,
 			filename = value.filename;
 
@@ -143,9 +143,10 @@ localfiles.prototype.addToSchema = function() {
 			item.set(field.path, []);
 		} else {
 			var values = item.get(field.path);
-			var value = values.find(function (val) { return val._id === element_id; });
+			var value = values.find(function (val) { return val._id == element_id; });
 			if (typeof value !== 'undefined') {
 				values.splice(values.indexOf(value), 1);
+				item.set(field.path, values);
 			}
 		}
 	};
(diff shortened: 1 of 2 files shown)


fields/types/localfiles/LocalFilesType.js
```diff
@@ -143,9 +143,10 @@ localfiles.prototype.addToSchema = function() {
 			item.set(field.path, []);
 		} else {
 			var values = item.get(field.path);
-			var value = values.find(function (val) { return val._id === element_id; });
+			var value = values.find(function (val) { return val._id == element_id; });
```
ericelliott: Why was `===` changed to `==` here?

fields/types/localfiles/LocalFilesType.js
```diff
@@ -170,7 +171,7 @@ localfiles.prototype.addToSchema = function() {
 		delete: function(element_id) {
 			if (exists(this, element_id)) {
 				var values = this.get(field.path);
-				var value = values.find(function (val) { return val._id === element_id; });
+				var value = values.find(function (val) { return val._id == element_id; });
```
ericelliott: `==` vs `===`?

Rmannn: If you use the === you are comparing the also the type of data.
Here, one is a string, and the other is an ObjectId ?
The condition will always fail in admin using the type comparator ===.

Rmannn: http://stackoverflow.com/questions/359494/does-it-matter-which-equals-operator-vs-i-use-in-javascript-comparisons

Rmannn: I did not take time to check what is the type of data, changing the == was a workaround, the function "exists" was returning false even if the file exists.

JedWatson: This is a common trap with mongoose. @Rmannn @ericelliott what do you think of wrapping both values in `String(id)` and using `===`? That way the conversion is explicit rather than implicit. Otherwise we should at least have a comment explaining the intentional use.

jonathanp: If what we want to accomplish is to compare values with type coercion I think `==` is better suited than `String(x) === String(y)` since that's exactly what the `==` operator is for :) Maybe with a comment as you said, just because one might think that `val._id` is a `String`.

Rmannn: I do not see the goal of a comment because this is how javascript is suppose to work with == and ===.
Maybe a global note to remember how works mongoose and how to compare object ids

morenoh149: the goal of a comment and/or more explicitness is for maintainability and clarity. Even if you do know how javascript's `==` operator works I'm not a fan of it and avoid it whenever possible.

Rmannn: Yes you're right.

PR #1951, opened 2015-12-10
Review eslint

morenoh149:
* remove nonexistant gulp task from package.json
* change various eslint rules
* enable recommended ruleset

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,26 +7,18 @@
 	"plugins": [
 		"react"
 	],
+	"ecmaFeatures": {
+		"jsx": true
+	},
+	"extends": "eslint:recommended",
 	"rules": {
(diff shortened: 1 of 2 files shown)


morenoh149:
@JedWatson I removed all of the unused rules and enabled the recommended ruleset. I feel like this is a good approach. As violations are fixed you can learn more about each rule and decide if any of them should be changed.

ericelliott:
:+1:

ericelliott:
I like the rules, but definitely feel that the errors need to be addressed before merging this change. Don't break the CI checks. =)

JedWatson:
Thanks for the start @morenoh149. I'll start reviewing the errors this weekend. Agreed @ericelliott we shouldn't merge this until the tests pass.

PR #1962, opened 2015-12-11
changed NumberField to allow input of negative numbers

pswoodworth:
The current regex will scrub the input of the '-' character, hence it's impossible to enter a negative number.


JedWatson:
Good call @pswoodworth, thanks!

PR #1980, opened 2015-12-16
Fix conflicts on skin.min.css

jacargentina:
The file contained conflicts which where not resolved appropiately
This for example causes the HR button icon not being correctly displayed.


jacargentina:
@JedWatson  Also regenerating the .css file should be done automatically on building, or what?

JedWatson:
Thanks @jacargentina! We don't actually have a process for generating that, I don't think :/

Should probably go into the `build` process. Interested in taking a look?

jacargentina:
@JedWatson Yes i can, can u point me which npm modules would be better for doing the **less preprocessing** and later **css minification**?

Issue #1977, opened 2015-12-16
[0.4] Date fields default to Date.now()


JedWatson:
There's a bug in our new `DateField` and `DatetimeType` components where they display the value of `Date.now()` when provided an `undefined` value.

JedWatson:
Fixed by #2015

PR #2015, opened 2015-12-25
Fix: Date fields default to Date.now()

yn5:
Proposed fix for https://github.com/keystonejs/keystone/issues/1977


JedWatson:
Neat fix, thanks!

Issue #2009, opened 2015-12-23
0.4: Types.Code height attribute not working


webteckie:
The attribute value is currently not being applied.

yn5:
The referenced pull request: https://github.com/keystonejs/keystone/pull/2016 tries to fix this by passing the height property on to CodeField's FormInput component similair to how it is done in the HtmlField component. Unfortunately CodeMirror does not seem to use the height set on the textarea provided to CodeMirror#fromTextArea.

webteckie:
@yn5 you may need to use the setSize api.  See http://codemirror.net/doc/manual.html#setSize

yn5:
Thanks @webteckie. Updated the pull request.

PR #2018, opened 2015-12-27
Change indicating width not to be changed when setting height of CodeMirror instance in CodeField to comply with its documentation

yn5:
Change indicating width not to be changed when setting height of CodeMirror instance in CodeField to comply with its documentation. As suggested as an improvement on https://github.com/keystonejs/keystone/pull/2016 in order to fix https://github.com/keystonejs/keystone/issues/2009.


JedWatson:
Thanks @yn5, and @webteckie for the guidance on this :)

PR #2033, opened 2015-12-30
Fix: DateTime type cannot be correctly rendered.

dracupid:


dracupid:
Node 4.0 fails, but 0.12 is ok. Re-run the test? It seems that the problem is not caused by this PR.

JedWatson:
@dracupid looks fine to me, thanks for the fix!

PR #2035, opened 2016-01-02
fix: number filter is totally broken.

dracupid:
The number filter is boken and `filter` argument in the API URL is always an empty object.
This is caused by:
1. onChange handler is not correclty set and called.
2. invalid mode values.


JedWatson:
Thanks @dracupid - I've been working through the filters updating them from "working UI" to "actually working", hadn't gotten to this one yet. The fix is much appreciated - if you would like to check the rest & bring any others up to scratch that would be great too :grinning:

dracupid:
I am using keystone to build a website of my lab, and I am glad to contribute to this awesome project. I will try to fix any bugs I found. Thank you for your response.

PR #2054, opened 2016-01-05
drag drop across pages

snowkeeper:
Adds drag drop back to UI with **react-dnd**.

Adds `perPage` option back to list view.

dnd1
dnd3
dnd2

### DIFF
--- a/admin/client/components/ItemsTable.js
+++ b/admin/client/components/ItemsTable.js
@@ -1,0 +1,79 @@
+import React from 'react';
+import classnames from 'classnames';
+
+import Columns from '../columns';
+import CurrentListStore from '../stores/CurrentListStore';
+import ListControl from './ListControl';
+import TableRow from './ItemsTableRow';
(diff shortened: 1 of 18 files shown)


JedWatson:
Fantastic @snowkeeper, thanks!

Do you think you could follow up with another one that uses the same styles for the pagination that the top of the list has? (ideally, share the Elemental Pagination component as well? @jossmac can probably help out if you want)

snowkeeper:
No problem, I can work on that soon.

gmmeyer:
This requires react-dnd and react-dnd-html5-backend to be included as dependencies, right? Trying to run this commit makes the admin UI unable to show lists for me until I installed it.

snowkeeper:
Looks like I did not push the new package.json...  I will submit that later today.

Sorry about that.

JedWatson:
@gmmeyer @snowkeeper I've just pushed some updates which add the packages and include them in the common client-side bundle. Should be sorted now.

PR #2057, opened 2016-01-06
Add Universal React routing support

ericelliott:
This update adds a `component routes` option to Keystone's config which allows users to pass `react-router` routes into Keystone. In response, Keystone will wire up the routes on the server.

The same route modules can be used for rendering on the client side, as well, which adds universal routing support to Keystone.

You can test this PR using the [`universal-keystone-boilerplate`](https://github.com/ericelliott/universal-keystone-boilerplate).

Clone, configure, `npm install` and run `npm test` to see it working.

You can also load `localhost:3000/teams` in the browser to see a route rendered by the router.


server/createComponentRouter.js
```diff
@@ -0,0 +1,29 @@
...
+
+var match = ReactRouter.match;
+var RoutingContext = ReactRouter.RoutingContext;
+
+module.exports = function createComponentRouter (routes) {
+  return function componentRouter (req, res, next) {
+    match({ routes: routes, location: req.url },
+        function (error, redirectLocation, renderProps) {
+
+      if (error) return res.status(500).send(error.message);
```
ericelliott: Is there a way to trigger a custom 500 error message from Keystone?

JedWatson: what do you mean?

JedWatson:
Looks great @ericelliott, one thing I'd suggest changing: maybe call the setting `react routes` rather than `component routes`, which will make the "React" part emphasised for people new to Keystone. It's still not perfect but the clearest I've been able to come up with... Other suggestions / arguments welcome.

server/createComponentRouter.js
JedWatson: Also are there missed tabs & spaces in this file? the indentation in the snippet above looks weird

JedWatson:
Maybe `universal react routes`? baah I hate naming settings

ericelliott:
`react routes` sounds good to me. I'll update. =)

server/createComponentRouter.js
ericelliott: > what do you mean?

I assume Keystone supports rendering custom error messages -- the line above will send a pretty generic error.

> Also are there mixed tabs & spaces in this file? the indentation in the snippet above looks weird

Maybe, but lint passes. I'll check it out. Is there a lint rule to cover that? And did `.editorconfig` fall asleep at the wheel?

ericelliott:
What's the process for updating the docs?

webteckie:
@ericelliott docs are in https://github.com/keystonejs/keystonejs-site

server/createComponentRouter.js
ericelliott: > Also are there mixed tabs & spaces in this file? the indentation in the snippet above looks weird

It appears there's no mixed spaces and tabs.. but you're right. I'll fix the indentation.

ericelliott:
Updated.

ericelliott:
Super excited about this. Just sayin'.

JedWatson:
me too :grinning:

PR #2058, opened 2016-01-06
Express is a factory

ericelliott:
Express is a factory. No need for `new` here.

See every express doc:

```js
var express = require('express');
var app = express();
```


JedWatson:
:+1:

PR #2061, opened 2016-01-06
compile stylus files with middleware

### LINKED ISSUE #2060: Compile Stylus with middleware
For the sake of consistency, I think that Stylus files should be compiled by the Keystone server, just like LESS and SASS files are being compiled at the moment. I already wrote a working bindStylusMiddleware.js based on the existing bindSassMiddleware.js, so I can make a PR that closes this issue.

In addition, this requires a small change to the generator, so that Stylus will be added as a dependency to the generated project.

jkallunki:
resolves #2060 by using Stylus middleware to compile .styl files

### DIFF
--- a/server/bindStylusMiddleware.js
+++ b/server/bindStylusMiddleware.js
@@ -1,0 +1,38 @@
+module.exports = function bindStylusMiddleware (keystone, app) {
+	// the stylus option can be a single path, or array of paths
+	// when set, we configure the stylus middleware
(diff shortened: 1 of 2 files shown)


JedWatson:
This looks good to me @jkallunki - thanks!

Would you mind updating the docs in the `keystone-site` repo as well to add the new `stylus` option?

jkallunki:
Ok, great! I'll check the docs a bit later @JedWatson

PR #2063, opened 2016-01-06
Custom Logging Middleware for v0.4

gmmeyer:
I would love to have the ability to define my own custom logging middleware. To be honest, I'd prefer to use Winston to Morgan, if I could.

This is a PR for v0.4 that does just this.

I allow for two different manners of defining this.

* One is through an option on the keystone init.
* One is through the normal middleware hook, so that people will be able to use their normal middleware workflow rather than putting it in a weird place.

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,7 +24,7 @@ var moduleRoot = (function (_rootPath) {
  * @api public
  */
 var Keystone = function () {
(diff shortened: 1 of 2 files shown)


JedWatson:
Looks great @gmmeyer, thanks :)

gmmeyer:
:)

ericelliott:
Great addition. I'll be using this, myself. Thanks!

`\( ﾟヮﾟ)/`

webteckie:
Docs?

ericelliott:
:+1: for docs

PR #2068, opened 2016-01-07
Drag drop styles

snowkeeper:
Mirrored pagination styling.

Placed drop on page target into dev mode.  Dropping on a page target requires a sane `sortOrder`, which requires additional methods when removing items and final order sanity checks.


JedWatson:
Looks good @snowkeeper, thanks :)

PR #2082, opened 2016-01-10
fix regression with admin path and sortable

snowkeeper:


ericelliott:
Is there a good way to test this regression?

JedWatson:
Good question @ericelliott - might be something for integration / UI tests? testing drag and drop programatically sounds like black magic to me, the `react-dnd` repo might hold some clues though

snowkeeper:
Once I realized I still had the hard code value I ran some greps like this to find others...
<kbd>grep --exclude-dir={node_modules} -rnw --include=*.js . -e "/keystone" >> ../grepList.txt</kbd>

ericelliott:
> might be something for integration / UI tests?

Yeah, I've been intrigued by a new functional test library that we might be able to put to good use on UI tests. I'll experiment with it in the Universal Keystone Boilerplate and then we'll consider pulling a more polished functional test suite into the Keystone repo. =)

morenoh149:
nightwatch vs the intern @.@

PR #2083, opened 2016-01-10
Add new sortable list items to the front of the list

snowkeeper:
Allows `sortable` to be set to `unshift` which adds each item as number **1** and increments the current entries.


JedWatson:
looks great, @snowkeeper.

JedWatson:
@snowkeeper just a quick reminder to watch the whitespace - see my updates referenced above

Would you mind updating the docs with the new option as well?

PR #2095, opened 2016-01-14
Refactor Pagination to allow optional expressions

alsoicode:
## Rationale
A user may want to paginate results that are from a full text search where an extra expression is used for sorting. Example:

    Product.paginate({
            page: req.query.page || 1,
            perPage: 1,
            filters: {
                $text: { $search: req.query.searchTerm || '' }
            },
            optionalExpression: {
                score: { $meta: 'textScore' }
            }
        })
        .sort({
            score: { $meta: 'textScore' }
        })
        .exec(function(error, results) {
            locals.results = results;
            next(error);
        });

## Problem
- The current implementation of [lib/list/paginate.js](https://github.com/keystonejs/keystone/blob/master/lib/list/paginate.js) does not allow for extra expressions to be passed to `model.find()`
- When extra metadata is added to the find statement, it interferes with the `query.count()` operation, causing it to fail with:
    
    ```Error: field selection and slice cannot be used with count```

## Solution
- Added `options.optionalExpression` value to `model.find()`. This allows an additional object to be passed to the find method. If `options.optionalExpression` is undefined, the `.find()` operation is unaffected.
- Added an additional query that does not include the `optionalExpression` object to perform the count of objects. This allows the count to be performed without additional parameters. See: https://github.com/aheckmann/mquery/issues/55 for more information.

## Testing / Linting
- All tests are passing via: `npm test`
- No jslint errors

### DIFF
--- a/lib/list/paginate.js
+++ b/lib/list/paginate.js
@@ -19,7 +19,8 @@ function paginate (options, callback) {
 
 	options = options || {};
 
-	var query = model.find(options.filters);
+	var query = model.find(options.filters, options.optionalExpression);
+	var countQuery = model.find(options.filters);
 
 	query._original_exec = query.exec;
 	query._original_sort = query.sort;
@@ -46,8 +47,8 @@ function paginate (options, callback) {
 	};
 
 	query.exec = function (callback) {
-		query.count(function (err, count) {
-			if (err) return callback(err);
+		countQuery.count(function(err, count) {
+			if (err) callback(err);
 
 			query.find().limit(resultsPerPage).skip(skip);
 


ericelliott:
Any way you could add some functional tests to ensure that this works correctly?

JedWatson:
Looks good to me @alsoicode, thanks for the excellent PR.

I'm happy to merge this as-is but am interested in your take on how we could create tests to cover it (and the rest of Pagination) as @ericelliott mentions.

webteckie:
Docs?  @JedWatson are ee deviating from all the discussions we had at one point on keeping the docs updated?  At the very minimal, an issue should be opened in the docs repo to track it.

alsoicode:
Thanks @JedWatson, always glad to help. @ericelliott absolutely. I'm not as experienced with testing Node applications as I am with Python. Should we just assume the tester has MongoDB running and set up a test project with models, etc? Or should we perhaps look at something like [Mockgoose](https://github.com/mccormicka/Mockgoose)?

I'm open to suggestions, more than happy to write tests and contribute to documentation.

webteckie:
@alsoicode you can open an issue at https://github.com/keystonejs/keystonejs-site/issues to track doc for this.  Thanks!

morenoh149:
@alsoicode can you mock out the db response with sinon? as that's already a dev-dependency we have in keystone

alsoicode:
@ericelliott and @morenoh149  I noticed there's a Mongoose connection in test/helpers/getMongooseConnection.js, so I've been using that to write out some functional tests and it's working well so far. Do you think that's an OK approach?

ericelliott:
@alsoicode I think considering what this does, it's OK to write integration tests for this that actually hit Mongo. Looking forward to reviewing your work on it. Thanks!

alsoicode:
Super. I thought the same since doing something like a `$text` search adds extra meta data to the result set. I'll do my best to get a new PR with these tests submitted in the next day or so.

Issue #2090, opened 2016-01-13
Feeding data into universal rendering components


ericelliott:
I've been looking at how to compile data from the server and make it available for the new universal React routing & rendering capability that was recently added to 0.4.x.

That data should obviously include user data, relevant lists, etc... but we don't want to pass it explicitly through the component view hierarchy. We haven't discussed the topic in-depth, but the community seems to be converging around Redux and it's rapidly expanding ecosystem of developer tools. There's a really [great tutorial course](https://egghead.io/series/getting-started-with-redux) by @gaearon. 

Currently, it appears we take a bit of a wild west approach to [bootstrapping the client page with data from the server](https://github.com/keystonejs/keystone/blob/c4c0e7d6631f8cf2197b5b41d4fe1c856cc8081e/lib/core/render.js), with no consistent user-extensible mechanism that I can see to determine what data is intended to be delivered to the client.

It also seems that there are bits and pieces of it being written by different page views (e.g., via `viewLocals` and arbitrary data being shoved directly into `render()` calls). Additionally, our current React admin session store doesn't appear to be designed for universal rendering and the Node request/response cycle. This stuff hasn't been a huge issue so far because users define their own views and can essentially stuff any data they want anywhere they want, but that approach won't work with baked in support for universal rendering.

Thankfully, both Express and React provide mechanisms that may be an effective path forward.

First, Express has a `res.locals` environment that is intended to be a safe way to gather data for _the current request/response cycle, only_. This is a good place to store things like authentication data, current user preferences, CSRF, "view locals", and so on. We should use it for its intended purpose, and create a special key that users of Keystone can easily extend with whatever custom data they need to generate on the server for delivery to the client.

Second, React has a `context` mechanism that seems to be a good candidate for this stuff. It's a way of implicitly passing data down through the entire React component hierarchy as opposed to explicitly passing everything from parent to children in `props`.

Receiving context is opt-in, and a React component must opt-in to the particular context keys that they're interested in. Additionally, Redux & React-Redux offer a simple way to pass store context into the app.

I'm thinking that we should create a special `res.locals.context` key that Keystone users can add to in their own middleware, and then we should ensure that it gets added to a Redux store and made available in the client bootstrap (currently we set a bunch of variables directly on a `Keystone` object on the client page... we can do something similar, or keep doing that).

I need a working mechanism to pass state into universal routes for both server and client rendering right now. I'm using it for language preferences and user data in the short-term. I'd love to hear your feedback.

[...]

PR #2101, opened 2016-01-15
Custom actions

### LINKED ISSUE #1206: Add option to generate custom actions in Admin UI
It would be handy to add some custom actions to Admin UI. Such actions that do not require any input and they just return a message and information whether the operation was succesful or not.

For example, for posts there could be a button Archive. This would be in the Admin UI in the view that shows the selected post.

A draft:

```javascript
var Post = new keystone.List('Post', {
    autokey: { path: 'slug', from: 'title', unique: true },
    map: { name: 'title' },
    defaultSort: '-createdAt',
    actions: {
       'Archive': archive
    }
});

var archive = function(next) {
    this.state = 'archived';
    this.save(function(err) {
        if (!err) {
            next(err, 'Post is archived');
        } else {
            next(err, 'Archiving failed');
        }
    });
}
```
Of course this example is not very useful, but I think that useful use cases exist.

adamscybot:
This pull request is to resolve #1206. Essentially it adds initial support for custom actions on item detail pages. I plan on extending this further to also support list pages. This lets you add extra buttons like this:

screen shot 2016-01-15 at 14 57 22

Usage:

``` javascript
Post.customAction('Archive', function(req, res, next) {
    var item = req.items[0];
    item.status = 'archived';
    item.save(function(err) {
        if (err) {
            throw new Error("Couldn't archive this post!");
        }

        next('Post was archived successfully!');
    })
})
```

You can optionally pass an options object that currently accepts `mobileText` for the button text on mobile. You can also override the unique slug assigned to the endpoint for the action with the `slug` option.

The item that the action was called on can be accessed via `req.items[0]`. I intend to go on to add custom actions to Lists, hence the array format. For lists, `req.items` will be an array of items that were "ticked" on the list page. This is going to require some UI work and probably support for things like batch delete to go with it for it to make sense. You will be able to limit it if you only want it on list view or you only want it on item view.

Errors (via throwing) and success messages (via next()) are pulled through to the UI as flash messages. Instead of utilising `Keystone.messages` bootstrapped into the template, I parse the asynchronous call response and add them dynamically with react. Please see #2090 for my reasoning behind this. This  work has also unlocked proper handling of messages of other asynchronous functions like delete which were marked as TODO. I will do a separate pull request for those. I hope this approach is suitable. It is my view that we should move to all error handling being parsed from responses to the calls and not from `Keystone.messages`. This will help down the line with true SPA support.

I'm happy with how it works on item views now, so if you're ok with it, it could be pulled in as WIP. I will then do the documentation.

One question I have is over positioning and styling. Is it ok as it is now? Or should we consolidate custom actions into a small dropdown button next to the save button. Just thinking of scenarios where someone has loads of custom actions.


webteckie:
@adamscybot this looks very good!  This will help my app as well!  I vote for them being regular buttons as you currently have it.  And I think they should inherit the same style as the Save button.  To minimize these taking extra real estate in the screen, perhaps the `reset changes` should just become a `Reset` button and `delete <list>` should become a `Delete` button both with a title tooltip with the extra info.

cc: @jossmac

JedWatson:
@adamscybot I **really** love what you've done here. I've been planning to add something like this in a slightly different way for 0.4 (specifically, exposing methods on a model but configuring the access to them client-side, not server-side through an Admin UI plugin API)

I've also been holding off on adding something like this while we explore all the possible ways of implementing custom actions for list items, so that we don't go down one path then have to back-track and break existing customisations.

_however_ I like your approach so much I think it might be worth going with regardless.

Positioning and styling are great for this variant; You might notice we've left room in the sidebar (could be a callout menu on mobile) which could also be used for custom components / actions. We may also want to make actions available at the top of the edit form.

One thing I've been thinking about that hasn't been taken into account here is the availability of custom actions; e.g. if a post is archived, it doesn't make sense to still have an "archive" button.

I'm going to think some more on this before merging, ping @creynders and @ericelliott as well for thoughts.

@adamscybot if you haven't seen it, you might also be interested in watching my nodevember talk from last year in which I outline some of the ways we're planning to open up Keystone for extensions, it's pretty relevant to this: https://www.youtube.com/watch?v=of2VM3Wwkxk

webteckie:
@adamscybot @JedWatson what about adding a o dependsOn config here?  With all that power @snowkeeper   added to dependsOn it shouldn't be a problem?

creynders:
Looks cool. Definitely need to be able to disable specific action buttons though. I'd go for a 'guarded' approach here though, i.e. call a handler per action/button type which returns a boolean to determine the `disabled` state (and not e.g. an object with booleans for each action type).

Then, I'm not fond of `callback` as the property name, since it's a bit of a misnomer IMO, wouldn't `action` be more appropriate? .

One more thing, I'm not entirely sure about the URL. I think I'd prefer `/api/<list>/<id>/actions/<action>`
Oh, and then I'd change the `delete` action URL too to `/api/<list>/<id>/actions/delete`

creynders:
@adamscybot wow, that's a truly great PR. From start to finish. Totally :+1: as far as I'm concerned.
Just one minor thing we need to consider: whether `save`'s a good property name. Since, if I'm not mistaken, the item is _not_ saved after the custom action is triggered (only before), right? The name might be confusing then. How 'bout `preSave`? Or it could be made more versatile, something like: 
```
{
  save: 'pre'
}
```
with possible values `'pre'`, `'post'` or `['pre', 'post']`?
Or even:
```js
{
  save: { pre: true, post: false}
}
```
What do you think?

webteckie:
@adamscybot looking good!  Can't wait for this :-)  As I was reading your comment the same impression as @creynders  about save crossed my mind.  Thus, I agree with his suggestions and prefer the very last one.  Thanks!

admin/client/components/EditForm.js
```diff
@@ -65,12 +68,35 @@ var EditForm = React.createClass({
 		list.deleteItem(data.id, err => {
 			if (err) {
 				console.error(`Problem deleting ${list.singular}: ${data.name}`);
-				// TODO: slow a flash message on form
+				// TODO: show a flash message on form
```
ericelliott: Is there an issue tracking this TODO, or is it just hiding here in the code? =)

ericelliott:
Is there some corresponding documentation for the new functionality and API endpoints?

ttsirkia:
Very nice job @adamscybot ! I had almost forgotten that I had created that issue. This implements exactly the feature that I was looking for. Now I have a few use cases for this: For example, I have a model for email validations. If there is a need to resend the validation message, this feature allows to implement that feature into the Admin UI.

morenoh149:
@adamscybot https://github.com/keystonejs/keystone/issues/2151

webteckie:
@JedWatson can we please take some action on this one?  I think this is a pretty useful feature!

VinayaSathyanarayana:
When is this getting merged?

creynders:
:100: times :+1: Let's get this merged!

webteckie:
:+1: :+1:

creynders:
@JedWatson could we get your blessing on this one? If so, @adamscybot could you fix the merge conflicts? Sorry for the long wait!

webteckie:
@creynders @adamscybot I discussed this one with @JedWatson and he thinks it is best to address it after 0.4 since otherwise it will make 0.4 harder to get out, which is his first keystonejs priority. Trust me I want this as bad as you all do:-)  I sympathize with Jed and to help get to this faster our efforts at this moment would be better spent helping with getting 0.4 out ASAP.  I'm pretty sure this PR will be amongst the first to be considered once 0.4 is published.  Now if Jed can change his mind about including this in now then that's something else :-)

mxstbr:
I'm not entirely up to speed here, why would we wait to get this in until the PR is out of date and hard to rebase instead of now? Is there any more work that'd need to be done for 0.4 to be released, how would this block that?

webteckie:
@mxstbr there's a few of us that really want this.  If you can convince @JedWatson to merge it that would be great.  I had a talk about it with him and there were concerns of merging in and how it would affect the remaining of the 0.4 work.  If that's no longer a concern then it should definitely be merged!

SharadKumar:
Hi @adamscybot - @JedWatson and team are preparing v0.4 beta. Are you plan to rebase since there has been quite a lot of changes since then and bring this up to date?

Appreciate all the effort mate.

VinayaSathyanarayana:
Any updates on this?

michaek:
This is a great feature addition. If there's any way I can help prep this to go into KeystoneJS, I would be happy to. It seems like @adamscybot is waiting on @JedWatson's answer on whether the internal API is stable enough to re-roll this. As 4.0 is now in beta, does that make it more likely to be stable?

VinayaSathyanarayana:
Any updates on this PR getting merged?

VinayaSathyanarayana:
Any updates on this PR getting merged?

VinayaSathyanarayana:
Any updates on this PR getting merged?

VinayaSathyanarayana:
Any updates on this PR getting merged?

ichiriac:
any news ? I also need this PR, exactly the same need

Twansparant:
This functionality would be great yes! +1

JedWatson:
This isn't going to land in v4, but is in scope for v5 (see keystonejs/keystone-5#1002)

Issue #1979, opened 2015-12-16
0.4:  CodeMirror support enhancement plan


webteckie:
I am proposing the following CodeMirror support enhancements.  I will work on this on and off as time allows.  If anyone wants to help please let me know so that we do not duplicate each other's work.

   - [ ] Refactor Types.Code to Types.CodeMirror (#1972)
    
   - [ ] Update Docs/README to reflect refactored field
    
   - [ ] Add support for CodeMirror addons
    
        Currently thinking of doing this via an attribute to the Types.CodeMirror field.  For example:
        
            addon: {
                js: ['display/fullscreen.js']
                css: ['display/fullscreen.css']
            }

            This would be preprocessed by the field onto something like:
            
                this.addonJS = ['display/fullscreen.js']
                this.addonCSS = ['display/fullscreen.css']
                
            Then it would be activated in the template doing something like:
            
                each css in addonCSS
                    link(rel="stylesheet", href="#{adminPath}/js/lib/codemirror/addon/}#{css}")

                each js in addonJS
                    script(src="#{adminPath}/js/lib/codemirror/addon/}#{js}")

   - [ ] Support user-generated CodeMirror bundles 
    
        CM is huge and most apps do not need all combinations. For example, we may have:
        
            codemirror.min.js  (default and includes the world)
            codemirror-myapp.min.js   (user override and includes just support for what the app really needs)
            
        Then need to figure out whether to configure bundles as a keystone config or as a field config.  I'm
        currently thinking as field config since it would be more flexible (e.g., one page may want just web
        and another the full support).  For example:
        
            content: {
                type: Types.CodeMirror,
                height: 400,
                bundle: codemirror-myapp.min.js,    // the field type would default it to "codemirror.min.js"
                config: {
                    mode: {name: "jade", alignCDATA: true},
                    lineNumbers: true
                }
            }

        Then script/components.jade would do something like:
        
            if list.fieldTypes.codemirror
                script(src="#{adminPath}/js/lib/codemirror/#{field.options.bundle}")

    
   - [ ] Document how-to-generate CodeMirror bundles

   - [ ] Add support for CodeMirror events
    
   - [ ] Add support for a CodeMirror toolbar

webteckie:
Closing this issue as  #2964 was the most critical for me.  If I (or anyone else) see value in implementing any of the others we can just PR the enhancement.

PR #2108, opened 2016-01-16
refactor(Types.Code): to Types.CodeMirror

webteckie:
@JedWatson this is a remake of #1972 into it's own branch.  Due to my infrequent PRs I thought the master branch would suffice but now have other PRs I would like to send your way.  I know you have reservations about this change and, if you ultimately decide not to merge it then all good.  But please see my explanation at the end of #1972 to see if it convinces you :-)  I will commit to updating all docs should it be merged!


mxstbr:
To continue the discussion from https://github.com/keystonejs/keystone/pull/1972#issuecomment-169320622:

I don't think we should do this. That we're using codemirror is an implementation detail of the code field. Instead of having `Types.CodeMirror`, `Types.Ace`, etc. I think having a config option called `editor` (or something) where you can then change the editor used and the specific configuration under an `<editorname>` configuration option is a better way to do this. It'd allow us to add unlimited different editors without polluting our fields, while still letting developers choose the editor they prefer.

webteckie:
@mxstbr i'm fine with your suggestion. Let's add it to #1979 and close this PR.

PR #2109, opened 2016-01-17
Pagination Functional Tests

alsoicode:
## Pagination Functional Tests

- Added simple "Post" model with a multi-key, weighted index.
- Added tests and assertions for list.paginate to ensure no existing functionality has been affected by the addition of the `optionalExpression` and `countQuery` modifications.
- Added tests and assertions for a paginated result set that includes an `optionalExpression` that adds metadata to the result.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,3 +27,10 @@ git:
 cache:
   directories:
     - node_modules
+
+addons:
+  apt:
+    sources:
+      - mongodb-3.0-precise
+    packages:
+      - mongodb-org-server
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ node_js:
 before_script:
   - npm run lint
   - sleep 15
-  - mongo test --eval 'db.addUser("travis", "test")'
 
 script:
   - npm run test-cov
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ node_js:
 before_script:
   - npm run lint
   - sleep 15
-  - mongo test --eval 'db.addUser('travis', 'test')'
+  - mongo test --eval 'db.addUser("travis", "test")'
 
 script:
   - npm run test-cov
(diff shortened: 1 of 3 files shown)


alsoicode:
I'm assuming the tests failed the Travis build because MongoDB isn't available in that environment. Any suggestions as to how we can work around that?

When I'm working on Django projects that have database-related tests, I use SQLite. Is there a MongoDB equivalent?

webteckie:
Maybe this helps:  https://docs.travis-ci.com/user/database-setup/#MongoDB

webteckie:
Although the tests at https://github.com/keystonejs/keystone/blob/master/fields/types/relationship/test/server.js seem to require mongodb up already.  At least, in my environment, if mongo is not running the tests fail.

alsoicode:
I think in the case with my tests, it might be a Mongo version issue, but I haven't quite nailed down the right config yet.

alsoicode:
Finally getting somewhere. I updated the version of MongoDB to 3.0, and tests are passing for the 4.0 environment but failing for 0.12. That's a pretty old version... thoughts anyone?

morenoh149:
afaik lots of things in keystone require a fairly recent version to run

alsoicode:
Considering that 5.4.1 is the most recent at the time I'm writing this, 0.12 seems ancient :)

ericelliott:
Are we supporting 0.12 for 0.4.x?

JedWatson:
@ericelliott @alsoicode we _were_ planning to, but the longer it gets the less inclined I am to worry about it.

Our tests frequently fail with *false negatives* on 0.12, and I'm not set up to easily replicate them so it's adding an unfortunate amount of overhead at the moment. Not sure there's really a lot of value when there is so much to do for 0.4 tbh.

AFAIK there is _nothing intentional_ in 0.4 that would break with node 0.12 but it needs review.

I'm going to merge this in the meantime so it doesn't fall behind, and open a new issue about 0.12.

JedWatson:
oh, and @alsoicode - thanks for these tests :grinning:

alsoicode:
@JedWatson You're very welcome.

PR #2131, opened 2016-01-21
Fix Confirmation Dialog

andrewlinfoot:
Confirmation dialogs for delete and reset never opened after the button was pressed.


JedWatson:
Right! Thanks, @andrewlinfoot :)

Issue #190, opened 2014-02-24
New field type: JSON


JedWatson:
Would store a JSON string in the database, and have a nice renderer / editor in the Admin UI.

Could use [pretty-stringify](https://github.com/JedWatson/html-stringify) for formatting the JSON, and [Code Mirror](http://codemirror.net) for editing.

Would validate JSON on save.

uptownhr:
problem with using "Types.Code" is that it will store the JSON as text. Meaning you will not be able to run queries against it.

morenoh149:
@uptownhr someone has to code it. There's probably no technical reason blocking this.

jbalde:
I created a field type Types.Json that store in mongo a JSON in raw format. I commit two pull requests to branch 0.3.x and master.

kadosh1000:
Hi I have published my npm package: 
https://github.com/kadosh1000/keystone-custom-fieldtypes

It allows you to create custom field types, and copy your custom fieldTypes from a selected dir into the keystone's fields dir and registers it.

mxstbr:
We're closing all feature requests to keep the issue tracker unpolluted. From now on submit feature requests on [productpains.com](https://productpains.com/post/keystonejs/new-field-type-json)!

ChristianRich:
This worked for me and allowed my to save an arbitrary object, only it's not visible in the Admin UI because it is unsupported. But you can query the fields just fine.

```
list.schema.add({
    model: mongoose.Schema.Types.Mixed
});
```

gemscng:
@mxstbr  @JedWatson  i have make a pull request on latest version base on . @jbalde 
 code.  https://github.com/keystonejs/keystone/pull/4855

victusfate:
since this has been around for 3 years, is there a suggested work around for existing data schema with Mixed types (responses from third parties etc)

trying out keystone v5 and was hoping to hit the ground running with a light cms around my data
```
    transcription_job: { type: JSON, schemaDoc: 'response of transcription job' }
```

no good atm

internetErik:
@victusfate 
Not a great solution, but you could just save JSON as plain test - perhaps in a Textarea type - and after reading it in from mongo in your controller run JSON.parse on it.

PR #2136, opened 2016-01-22
New field type: JSON #190

jbalde:
New field type JSON based on Code field

Define in model like:
```javascript
fieldname: { type: Types.Json }
```

Only accept valid JSON Object value. If try use a simple values like string, book, number keystone throws a error in this field.

Save the data in mongo in JSON format for make searchs.

Example of data:

```json
{
	"prueba4": [
		1,
		2,
		3
	],
	"prueba2": 1,
	"prueba": "valor",
	"prueba3": true
}
```

### DIFF
--- a/admin/client/fields.js
+++ b/admin/client/fields.js
@@ -12,6 +12,7 @@ module.exports = {
 	embedly:          require('../../fields/types/embedly/EmbedlyField'),
 	geopoint:         require('../../fields/types/geopoint/GeoPointField'),
 	html:             require('../../fields/types/html/HtmlField'),
+	json:             require('../../fields/types/json/JsonField'),
 	key:              require('../../fields/types/key/KeyField'),
 	localfile:  	  require('../../fields/types/localfile/LocalFileField'),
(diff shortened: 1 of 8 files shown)


mxstbr:
Thanks for doing this @jbalde, from a first glance this is looking quite good! I don't think it's worth cutting another 0.3 release for this, so I'm inclined to focus on this PR.

As per #2438, we're updating all field types to be completely and utterly tested. Could you rebase this onto the newest changes on `master`, rename the test file from `server.js` to `type.js` and add tests for all methods?

VinayaSathyanarayana:
Any inputs on when this will be merged?

stennie:
@VinayaSathyanarayana As per the previous comment from **mxstbr**, there is still work to do before this is ready to be merged:

> As per #2438, we're updating all field types to be completely and utterly tested. Could you rebase this onto the newest changes on master, rename the test file from server.js to type.js and add tests for all methods?

The PR won't progress unless someone is interested in finishing up the required changes. Currently there are no volunteers so this will remain the backlog.

If you (or anyone else) is keen to see this field type added, an updated PR would be welcome.

Regards,
Stennie

PR #2138, opened 2016-01-23
Fix: isUTC on DateField

andrewlinfoot:


JedWatson:
@andrewlinfoot, catching _all the things_ :grinning: 

Great work, thanks!

PR #2141, opened 2016-01-24
List.create item fixes

andrewlinfoot:
Fixed an error with errors while using createInline. Switched the List.createItem endpoint to send up form data instead of JSON so that it can support file uploads.


JedWatson:
Great!

PR #2142, opened 2016-01-24
Add: search to columns Popout

andrewlinfoot:
I implemented search on the columns popout just like it is on the filters popup. 
image


JedWatson:
:+1:

Issue #1714, opened 2015-10-07
Replace Field.prototype.getRequestHandler with processUpdateData method


JedWatson:
Field types currently have limited (i.e. defined by the `UpdateHandler`) support for async ways to update items in the form of the `Field.prototype.getRequestHandler` method.

This has a number of limitations.

I think any async process or transformation (e.g. storing uploaded files, autocompleting locations) should return a standard data structure that can then be validated by the `validateInput` method, as if it were provided in the initial data to update.

This will simplify the rules around how data is preprocessed, validated, and updated by fields.

So instead of a forked update path per field type of:

`getRequestHandler(item, req, paths, callback) -> process() || validateInput(data, required, item) -> updateItem(item, data)`

we'd now have:

`processUpdateData({ data, files }, callback) -> validateInput(data, required, item, callback) -> updateItem(item, data, callback)`

... which is much more consistent and flexible.

(validate and update processes will be async, see #1713)

The `processUpdateData` method would be optional and skipped if not defined on the field type.

I _think_ it should transform the `data` object by reference to allow for parallel async processing; it will be important not to introduce side effects by modifying other keys in the `data` argument.

Alternately we could return the data that should be passed to the `validateInput` and `updateItem` methods, but I think that might be unnecessary overhead as each field would need an implementation of `processUpdateData` that returns a new object containing the keys from `data` that should be passed on. @creynders @dcousens any feedback on that?

mahnunchik:
+1 to `processUpdateData({ data, files }, callback) -> validateInput(data, required, item, callback) -> updateItem(item, data, callback)`

andrewlinfoot:
@JedWatson like the new API. How/where do we validate the files for things like CloudinaryImage? Would that happen in the `processUpdateData` function?

andrewlinfoot:
@JedWatson Another question: why are we passing in `required` as a parameter to `validateInput(data, required, item, callback)` when the field has a property `this.required`?

andrewlinfoot:
I implemented a version of this in https://github.com/keystonejs/keystone/pull/2146. I changed the method name from `processUpdateData` to `processFormData` because I thought it was confusing to have both `updateItem` and `processUpdateData`. `processFormData` seemed more appropriate because the method is only needed if you use file uploads from a form because both `validateInput` and `updateItem` are async so you can do any non-file related data pre-processing in those two methods in the majority of cases.

[...]

Issue #1713, opened 2015-10-07
Make Field.protoype.updateItem and .validateInput async


JedWatson:
To support more flexible custom field type implementations in 0.4, we should make the field validation & update methods async. This would allow for external (e.g database or even web service) validation processes, which would be powerful.

Will require updates to all the current field types and tests.

mxstbr:
Done!

Issue #1705, opened 2015-10-05
New List / Item Update method


JedWatson:
To enable RESTful API support for the Admin UI (as well as custom API routes), we need a new List.prototype.updateItem method to replace List.prototype.getUpdateHandler.

This is a placeholder issue while I work out what the API should actually look like, but in particular it needs to allow for Fields to hook their own update and async logic in, rather than having it built into the core function as is the case with the UpdateHandler functionality.

Probably relates to #1563, how we handle files will be an important part of how the method works.

Once the single-item update is working, we need to make a call on continuing to support the `UpdateHandler` as it is, or whether it can be implemented as a wrapper for the new method. I'd like to avoid breaking the API for 0.3 > 0.4 if possible, but this is an area that really needs to be cleaned up.

JedWatson:
Status update: basic field updates work now (where `Field.prototype.updateItem` is sufficient); needs more testing around validation, and needs to support async update processes (e.g. file upload)

ericelliott:
:+1:

JedWatson:
Progress update / more info:

We're basically replacing

* https://github.com/keystonejs/keystone/blob/master/admin/server/routes/item.js and
* https://github.com/keystonejs/keystone/blob/master/lib/updateHandler.js

with

* https://github.com/keystonejs/keystone/blob/master/admin/server/api/item/update.js and
* https://github.com/keystonejs/keystone/blob/master/lib/list/updateItem.js

... and fully fleshed-out / async update methods on Field types as per #1713

Then we can complete #1714

ericelliott:
Thank you! I was just trying to dig into these details. =) You read my mind!

(∩｀-´)⊃━☆ﾟ.*･｡ﾟ

ericelliott:
Where are the API tests?

JedWatson:
There aren't any tests yet for the REST API (we should add some as part of this work), the field types have a `/test` folder under each `/fields/types/{type}` which directly tests the field type class

ericelliott:
Cool. I've been reading up a bit more on lists and field types in the docs. I'll play with it a bit more and try to soak up enough context to write good tests. =)

wmertens:
Note that the exact implementation is still being discussed at #3209

JedWatson:
This is done now, good to close!

PR #2146, opened 2016-01-24
Add: async creation of items

andrewlinfoot:
This is a WIP to tackle issues: https://github.com/keystonejs/keystone/issues/1705, https://github.com/keystonejs/keystone/issues/1713, and https://github.com/keystonejs/keystone/issues/1714. I'd love to get some feedback on it.

I just implemented async validation and left the `.updateItem` method alone. If this approach looks like the right direction, I'll make the `updateItem` methods async as well. From there, it should be easy to add the `.processUpdateData` api and handle file uploads.

A couple questions about the API in general:

1. Should the `.validateInput` api return more information than just `false` when there is an invalid input? I'm having trouble seeing how we would implement more granular error messages with the API as is.

2. Should the `files` be passed to the `.validateInput` method too or should file validation happen somewhere else?

3. Is it necessary to pass `required` to `.validateInput(data, required, item, callback)` when the field has the property `.required` and it can be accessed with `this.required` in the `.validateInput` method?

Overall, this new API seems way cleaner than the UpdateHandler approach.


andrewlinfoot:
I submitted a different PR that made all of the `updateItem` methods async. That is merged in and good to go (https://github.com/keystonejs/keystone/pull/2147).

I have some free time tonight and tomorrow and am planning on tackling the `.processUpdateData` API and polishing off the new update and create API endpoints. I would love to get some thoughts on this PR before I jump to deep into implementing the rest of the validation errors and adding support for file uploads. @JedWatson @morenoh149 @creynders @webteckie @ericelliott any thoughts?

andrewlinfoot:
Just pushed some updates that implement a `processFormInput` method and change the `validateInput` method to call its callback with an error object instead of just `true` or `false`. This seems like a pretty solid API and should allow for file uploads, async data updating, async data validation and fine grained error messages defined by the FieldTypes themselves. Assuming I can get some consensus around this API, I'm going to start implementing `processFormInput` methods for some of the types that depend on file uploads. @JedWatson thoughts?

andrewlinfoot:
Commit https://github.com/keystonejs/keystone/commit/a73d1315bb95c57a4743cf1b3a9648115b6d0e0f adds a minimal version of `processFormData` for CloudinaryImage types

andrewlinfoot:
Newest commits by Jed implement this API in a cleaner way. Closing the PR.

Issue #2140, opened 2016-01-24
Consolidate promises to es6


morenoh149:
Currently Keystone relies both on `bluebird` and `mpromise` it'd be great to drop these dependencies and use es6 Promises.

JedWatson:
Closed by #2149

morenoh149:
:100: @mldangelo dangelo

PR #2149, opened 2016-01-25
Convert promises to ES6 promises

mldangelo:
This PR addresses issue https://github.com/keystonejs/keystone/issues/2140.

MPromise and Bluebird dependencies have been removed.


JedWatson:
Neat, thanks @mldangelo - I didn't realise`Promise` was available in 0.12 but just checked and it is :grinning:

PR #2169, opened 2016-01-27
Fix: Async updating existing items

andrewlinfoot:
Async updates never called their callback when editing existing items.


andrewlinfoot:
Actually it was breaking when creating or saving items with noedit or password fields. I just didn't notice until I updated an existing item in one of my projects.

PR #2183, opened 2016-01-27
Fix tag to use the public id

### LINKED ISSUE #2182: Using the CloudinaryImage.tag() function, results in an incorrect src url
The schema method "tag" gets a field called "path"

https://github.com/keystonejs/keystone/blob/master/fields/types/cloudinaryimage/CloudinaryImageType.js#L184

```js
tag: function(options) {
			return exists(this) ? cloudinary.image(this.get(field.path), options) : '';
		},
```

that is an object-

```js
{
    public_id: '',
    version: 0,
    signature: '',
    format: '',
    resource_type: '',
    url: '',
    width: 0,
    height: 0,
    secure_url: '',
}
```

The result of using this function is an image tag that resembles -

```html
<img src="{cloudinaryName}/[object Object]" />
```

bringking:
This PR closes #2182 by passing only the public_id to the cloudinary.image function, and not an object (field.path)

### DIFF
--- a/fields/types/cloudinaryimage/CloudinaryImageType.js
+++ b/fields/types/cloudinaryimage/CloudinaryImageType.js
@@ -181,7 +181,7 @@ cloudinaryimage.prototype.addToSchema = function() {
 			return src(this, options);
 		},
 		tag: function(options) {
-			return exists(this) ? cloudinary.image(this.get(field.path), options) : '';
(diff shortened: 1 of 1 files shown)


morenoh149:
:+1: https://help.github.com/articles/closing-issues-via-commit-messages/#keywords-for-closing-issues

JedWatson:
Good call @bringking, thanks!

Issue #2162, opened 2016-01-26
Error on apply updates from file


w01fgang:
I see these error on first start when apply updates from `/updates` folder:
```
Applying update 1.0.0-users-companies...
TypeError: callback is not a function
    at nextTickCallbackWith0Args (node.js:433:9)
    at process._tickCallback (node.js:362:13)
```
I use [ketstone-test-project](https://github.com/keystonejs/keystone-test-project) 
I think it is because the updates file doesn't contain callback:
```
exports.create = {
	User: [
		{ 'name.first': 'Admin', 'name.last': 'User', email: '...', password: '...', isAdmin: true }
	]
};
```
Maybe after this PR #2147 I need to add callback, but don't know where.
I will try the long-hand version, short version doesn't applies.

samkuhn:
I see the same.. on a fresh Ubuntu server with the latest 0.4 version from master. Any luck fixing it w01fgang?

samkuhn:
I can confirm that a version from the 24th (2a68e2113e27e5b147bcd382ce1d354a9972c75f) creates the database user ok. So a later commit must have broken the update functionality.

Possibly PR #2147 as w01fgang suspects.. I would investigate but big deadline! :)

w01fgang:
This still works
```
'use strict'

let keystone = require('keystone')
let async = require('async')
let User = keystone.list('User')

let admins = [
	{ email: 'admin@email.com', password: 'changeOnStart', name: { first: 'Admin', last: 'User' } },
{ email: 'another@email.com', password: 'changeOnStart', name: { first: 'Another', last: 'Admin' } }
];

function createAdmin(admin, done) {

	var newAdmin = new User.model(admin);
	newAdmin.isAdmin = true;
	newAdmin.save(function(err) {
		if (err) {
			console.error("Error adding admin " + admin.email + " to the database:");
			console.error(err);
		} else {
			console.log("Added admin " + admin.email + " to the database.");
		}
		done(err);
	});

}
```

webteckie:
I can also confirm that this is broken very recently (last few days)!  This is a blocker for me as I have fairly complex updates files I constantly run.  If it was a breaking change somewhere please inform what it is that should be changed.

cc: @JedWatson

andrewlinfoot:
Yep, was a regression from https://github.com/keystonejs/keystone/pull/2147. Sorry about that. Check out https://github.com/keystonejs/keystone/pull/2196. Should have fixed this. Let me know if I missed anything.

[...]

PR #2196, opened 2016-01-29
Fix: Issue #2162

andrewlinfoot:
.updateItem wasn't updated to be async in keystone.createItems. This is
used by the shorthard update script syntax and therefore was only
throwing errors when applying updates with that syntax. See https://github.com/keystonejs/keystone/issues/2162

### DIFF
--- a/lib/core/createItems.js
+++ b/lib/core/createItems.js
@@ -103,29 +103,29 @@ function createItems(data, ops, callback) {
 						refs[list.key][data.__ref] = doc;
 					}
 
-					_.each(list.fields, function(field) {
+					async.each(list.fields, function(field, callback) {
 						// skip relationship fields on the first pass.
 						if (field.type !== 'relationship') {
-							field.updateItem(doc, data);
+							// TODO: Validate items?
+							field.updateItem(doc, data, callback);
 						}
-					});
-
-					if (options.verbose) {
-						var documentName = list.getDocumentName(doc);
-						writeLog('Creating item [' + itemsProcessed + ' of ' + totalItems + '] - ' + documentName);
-					}
-
-					doc.save(function(err) {
-						if (err) {
-							err.model = key;
-							err.data = data;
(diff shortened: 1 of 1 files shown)


webteckie:
@andrewlinfoot your changes were not enough for my scenarios.  I keep my own version of updates/createItems since I have some use cases, which the default modules do not cover.  But the code you change is the same as in my versions.  Here's what I had to do to mine version to get it to work:
```
...
                    async.each(list.fieldsArray, function(field, doneField) {
                        // skip relationship fields on the first pass.
                        if (field.type !== 'relationship') {
                            field.updateItem(doc, data, doneField);
                        } else {
                            doneField();
                        }
                    }, function(err) {
                        if (options.verbose) {
                            var documentName = list.getDocumentName(doc);
                            writeLog('Creating item [' + itemsProcessed + ' of ' + totalItems + '] - ' + documentName);
                        }

                        if ( !err ) {
                            doc.save(function (err) {
...
```

I believe if you update keystone's modules accordingly the problem will be solved for all others.  And for me that was not enough!  My app happens to call field.updateItem directly in other places and thus I had to change that too.  It may be worth noting some of these breaking calls in the Wiki's 0.3->0.4 breaking changes section.

JedWatson:
Thanks @andrewlinfoot!

@webteckie any chance you'd be able to add that info the the wiki?

webteckie:
@JedWatson you probably misunderstood my comment above.  I don't think this should have been merged, yet, as I suggested changes for @andrewlinfoot to do in createItems :-)  But if others are happy with the latest version of createItems then all good...I have my own version :-)  I'll try to find some time to update the wiki unless @andrewlinfoot takes that over.

JedWatson:
Sorry @webteckie - you're right, I did misunderstand. Going to merge your PR #2197 now - does that resolve it completely?

webteckie:
It should work for all now...hopefully :-)

Issue #2164, opened 2016-01-26
[0.3][0.4] error passed from update script disappears into the void


creynders:
A bug was introduced here https://github.com/keystonejs/keystone/blob/cdc8655fdb1beabbcb2c801e13ad316f92a5bf7f/lib/updates.js#L109

As you can see nothing happens when the `err` object is truthy. The same applies to https://github.com/keystonejs/keystone/blob/cdc8655fdb1beabbcb2c801e13ad316f92a5bf7f/lib/updates.js#L100
This means that if you pass an error from your update script, everything hangs and nothing is written to the out stream, keystone just sits there waiting.

@webteckie what do you think, is this as easily solved as I think, just by adding:

``` js
}else{
    done(err);
}
```

??

webteckie:
@creynders yeah, it seems like a bug not to call done(err).  However, if memory serves me well even if you did that the process would exit before the logs would be flushed.  And it may be one of those intermittent things.  The updates function (e.g., createItems) should be logging the error.  Anyways, feel free to fix, at least, not to hang the process :-)

JedWatson:
@webteckie fixed by #2197?

mxstbr:
_ping_ @webteckie is this fixed?

Noviny:
Going to close this as solved by #2197. Happy to reopen if it's not.

PR #2197, opened 2016-01-30
Fixed issues with updates and createItems.

webteckie:
These may be needed to completely address #2162 and #2164

### DIFF
--- a/lib/updates.js
+++ b/lib/updates.js
@@ -26,7 +26,7 @@ exports.apply = function(callback) {
 
 	var updatesPath = keystone.getPath('updates', 'updates');
 
-	// logError is used to log errors before the process exits since it is more synchronous than console.error.  Using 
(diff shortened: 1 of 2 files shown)


JedWatson:
Thanks @webteckie - let me know if you think this area needs more work.

lib/updates.js
```diff
@@ -101,6 +101,8 @@ exports.apply = function(callback) {
 							if (update.__commit__ !== false) {
 								new Update({ key: file }).save();
 							}
+						} else {
+							done(err);
```
creynders: This doesn't seem right either, since `done` is called on L108 anyhow, i.e. if an error occurs it will call `done` for a second time...

webteckie:
@JedWatson can you please comment on @creynders's comment?  I can remove L104-105 (if it was the wrong thing to do) or L108, otherwise.  Can you please explain background updates?

Issue #1011, opened 2015-02-06
dependsOn and required: true


morenoh149:
Let's suppose you're building a blog with keystonejs, a la https://github.com/JedWatson/keystone-demo/blob/master/models/posts.js#L14
Let's say your client has the requirement `All published posts should have a publishedDate`. Keystone has a convenience facility [dependsOn](http://keystonejs.com/docs/database/#fields) See `conditional Fields`. @webteckie proposed that combining `dependsOn` with `required: true` would properly enforce the above requirement. It does not. Instead the `require` forces all posts to then have a `publishedDate` - when most posts start out as drafts.

I was able to solve this with a mongoose pre-save hook https://gist.github.com/morenoh149/463fa1e717e67a6df228#file-post-js-L69

@keystonejs/contributors is this a bug? should keystone Lists support this usecase?

webteckie:
The short version of @morenoh149 use case is this:
```
	state: { type: Types.Select, options: 'draft, published, archived', default: 'draft' },
	publishedDate: { type: Types.Date, dependsOn: {state: 'published'}, required: true },
```

That seems logical and it should work as expected, IMO.  That is, the dependsOn attribute condition should take precedence over the required attribute setting.  If dependsOn===false then required should not be enforced.

JohnnyEstilles:
@morenoh149, like you, I was likewise skeptical when I first read about this feature. It felt counter-intuitive to conditionally `require` a field. Especially when tryng to make sense of it in terms of the *blog post state/publish date* use-case @webteckie presents above which, as you mention, can easily be enforced using pre-save hooks.

While I believe this is a very useful feature, I think the use of the  *blog post state/publish date* paradigm doesn't quite make the case for it.There are many other use-cases in which the concept of a conditionally required field is not only called for, but essential.

Take this scenario, for example: a field to capture a person's marital status, along with a conditional field for the name of that person's spouse, which is only required if the person is married.

```js
List.add({
    maritalStatus: { type: Types.Select, options: 'single, married, divorced, widowed', default: 'single' },
    spouse: { type: Types.Name, dependsOn: {maritalStatus: 'married'}, required: true }
});
```

Or maybe a scenario in which there you have a non-exhaustive list, represented by a `Types.Select` field which includes an `other` option to cover any options not listed in the select field, accompanied by a `String` field that is only required with `other` is selected.

```js
List.add({
[...]

PR #2200, opened 2016-01-31
Field require will respect dependsOn

snowkeeper:
Includes tests.

If a Field that has both `dependsOn` and `require: true` set, we first check for `evalDependsOn()`.
If `evalDependsOn` returns `true`, we will respect the `required` option. 
If `evalDependsOn` returns `false` we will ignore the `required` option.


webteckie:
Cool stuff...left some nitpick comments :-)

fields/types/Type.js
```diff
@@ -64,12 +64,25 @@ function Field(list, path, options) {
...
 		process.exit(1);
 	}
-
+	
+	// if dependsOn and required, set required to a function for validation
+	if (this.options.dependsOn && this.options.required === true) {
+		var opts = this.options;
+		this.options.required = function() {
+			// `this` refers to the validating document
+			debug('validate dependsOn required', evalDependsOn(opts.dependsOn, this.toObject()));
```
ericelliott: Is this needed?

JedWatson:
@webteckie I'm working on the linter now actually. Since they stopped enabling rules by default in ESLint, we've actually had some massive blind spots. Will be much more robust soon :)

fields/types/Type.js
snowkeeper: Without it working in the Admin UI would produce a different result that working on a List directly.

JedWatson:
This looks like a really good enhancement, @snowkeeper, except for one concern -

I am pretty sure Mongoose will validate `required` paths regardless of whether Keystone pre-validates them or not, so this would still actually fail but with mongoose error messages instead of Keystone's higher-level ones. May not be true of all field types (e.g. I don't think we specify `required` on complex field sub-paths, which may actually be something we should do...)

This needs to be looked into further; another option may be to _exclude_ the `required` option from the Mongoose schema if `dependsOn` is also set, and live with the limitation that validation will then only be done through Keystone's APIs (e.g. `updateHandler`)

snowkeeper:
Do the changes to fields/types/Type.js not address your concern?  

By setting `required` to a function we can control the validation.  The test file covers this unless I did something wrong.

webteckie:
According to http://mongoosejs.com/docs/api.html#schematype_SchemaType-required it seems like @snowkeeper's  changes in Types.js should handle @JedWatson's concerns.  They seem to disable the mongoose required validation on that field type if evalDependsOn is falsy.

ericelliott:
:+1:

snowkeeper:
Here is the mongoose [issue](https://github.com/Automattic/mongoose/issues/941#issuecomment-61155319) that talks about using `required` as a function.  

The conversation above the linked comment starting [here](https://github.com/Automattic/mongoose/issues/941#issuecomment-38994781) covers our exact use case.

snowkeeper:
We can actually extend this to so that the Admin UI respects  user supplied required functions in the schema.  

In `updateHandler.js` we can remove the `dependsOn` check and replace it with a function check
```
var isFunc = _.isFunction(field.required);
if (!isFunc || (isFunc && field.required.call(data))) {
	message = options.requiredMessages[field.path] || field.options.requiredMessage || field.label + ' is required';
	addValidationError(field.path, message);
	invalidated = true;	
}
```
Our `dependsOn` required function would then need to accept either a mongoose doc or plain Object.

I have these changes done if you want me to push them along.

dodiaraculus17:
**Sorry about Advertising**

I need contributors for my project. If you would like to help the project is located here.
https://github.com/poKe-Websites/poKe-CMS

Thank You :)

webteckie:
Can we get this merged if there's no other issues/concerns with this?

JedWatson:
@snowkeeper and @webteckie thanks for clarifying - I didn't dig deep enough in my original review before asking the question. Looks like a great solution :grinning:

webteckie:
@JedWatson thanks for merging @snowkeeper awesome work here!  @snowkeeper may want to add a bit of clarification to the docs about this.  @JedWatson you can close #1011 as a result of this merge.

Issue #1797, opened 2015-11-11
Review .eslintrc


JedWatson:
I've noticed a lot of things that should be considered warnings or errors are passing our linter. The configuration needs to be reviewed, may be out of date given updates to ESLint (and the React plugin)

morenoh149:
https://github.com/keystonejs/keystone/pull/1781 similar

morenoh149:
I took some time to review it https://github.com/keystonejs/keystone/blob/master/.eslintrc One thing I'll note is if a rule is not being used (value `0`) then it should be omitted from the ruleset altogether.

My suggestion is to review all the recommended rules [here](http://eslint.org/docs/rules/) and see which of those you like.

JedWatson:
I've made some major updates and fixed a lot of issues in the codebase, including pulling a lot of rules from `standard`. Also updated some old `jslint` flags in the code.

.eslint is much closer to "good", I've also got a wishlist of rules that I want to add next but will require code refactoring to pass:

```
* `"brace-style": 2,`
* `"dot-notation": 2,`
* `"eqeqeq": [2, "allow-null"],`
* `"guard-for-in": 1,`
* `"indent": [2, "tab"],`
* `"key-spacing": 2,`
* `"new-cap": 2,`
* `"no-undef": 2,`
* `"no-unused-vars": [2, { "vars": "all", "args": "none" }],`
* `"no-unneeded-ternary": [2, { "defaultAssignment": false }],`
* `"one-var": [2, { "initialized": "never" }],`
* `"operator-linebreak": [2, "before"],`
* `"prefer-const": 2,`
* `"space-in-parens": 2,`
* `"spaced-comment": 2,`
* `"yoda": 2,`
```

These can only be added once the codebase is fixed and they pass, otherwise we'll break our tests on PRs and commits.

If anybody has a chance to take any of these on and PR them shout out here and let us know, otherwise I'll work through the rest asap so we can close this.

Feedback on the rules I've gone with is also welcome!

karlbright:
- [ ] (dot-notation)[http://eslint.org/docs/rules/dot-notation]
- [ ] (eqeqeq)[http://eslint.org/docs/rules/eqeqeq]
- [ ] (guard-for-in)[http://eslint.org/docs/rules/guard-for-in]
- [ ] (indent)[http://eslint.org/docs/rules/indent]
- [ ] (key-spacing)[http://eslint.org/docs/rules/key-spacing]
[...]

PR #2230, opened 2016-02-05
Added space-in-parens rule to eslint

LennyBoyatzis:
As per issue #1797

### DIFF
--- a/admin/client/views/signin.js
+++ b/admin/client/views/signin.js
@@ -123,7 +123,7 @@ var SigninView = React.createClass({
 						<FormInput type="password" name="password" onChange={this.handleInputChange} value={this.state.password} ref="password" />
 					</FormField>
 					<Button disabled={this.state.animating} type="primary" submit>Sign In</Button>
-					{/*<Button disabled={this.state.animating} type="link-text">Forgot Password?</Button>*/}
+					{/* <Button disabled={this.state.animating} type="link-text">Forgot Password?</Button>*/}
 				</Form>
 			</div>
 		);
--- a/fields/types/Type.js
+++ b/fields/types/Type.js
@@ -1,4 +1,4 @@
-/*!
+/* !
  * Module dependencies.
  */
 var _ = require('underscore');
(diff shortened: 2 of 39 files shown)


JedWatson:
Thanks @LennyBoyatzis - a few things to address with this PR.

This is actually the `"spaced-comment"` rule :P

Comments starting with `/*!` are actually special syntax and shouldn't be caught by this rule - I am pretty sure there's a config option to exclude that character from requiring a space, we should specify it.

I also noticed a few instances (I'll comment directly) where there's a space at the start of a multiline comment but not the end - it should be consistent, there may be a way to configure the rule to catch those as well.

Finally I fixed the indentation in .eslintrc which will cause conflicts, would you mind rebasing and fixing these issues? Thanks!

fields/types/Type.js
```diff
@@ -1,4 +1,4 @@
-/*!
+/* !
```
JedWatson: This is special syntax, we don't want to introduce a space here

fields/types/azurefile/AzureFileType.js
```diff
@@ -144,7 +144,7 @@ azurefile.prototype.addToSchema = function () {
 		reset: function () {
 			try {
 				azure.createBlobService().deleteBlob(this.get(paths.container), this.get(paths.filename), function () {});
-			} catch (e) {}//eslint-disable-line no-empty
+			} catch (e) {}// eslint-disable-line no-empty
```
JedWatson: there should be a space _before_ the comment as well. Not sure which rule should catch that though, possibly a different one?

admin/client/views/signin.js
```diff
@@ -123,7 +123,7 @@ var SigninView = React.createClass({
 						<FormInput type="password" name="password" onChange={this.handleInputChange} value={this.state.password} ref="password" />
 					</FormField>
 					<Button disabled={this.state.animating} type="primary" submit>Sign In</Button>
-					{/*<Button disabled={this.state.animating} type="link-text">Forgot Password?</Button>*/}
+					{/* <Button disabled={this.state.animating} type="link-text">Forgot Password?</Button>*/}
```
JedWatson: if we need a space at the start, we should need a space a the end too for consistency

PR #2259, opened 2016-02-10
hook up from redirection

### LINKED ISSUE #1884: [0.4] Implement signin "from" redirect
For instance, after successful sign in from this URL http://keystone4-preview.herokuapp.com/keystone/signin?from=/keystone/contacts the user should be taken to the `contacts` list

erg0dic:
The "from" field is already provided when rendering the jade view, so I just pushed it down into react.

Closes #1884


JedWatson:
Thank you @erg0dic!

PR #2293, opened 2016-02-16
Added eslint rule No unneeded ternary

LennyBoyatzis:
As per issue Review .eslintrc #1797

### DIFF
--- a/admin/server/api/list.js
+++ b/admin/server/api/list.js
@@ -40,8 +40,8 @@ module.exports = function (req, res) {
 				if (!field) return sendError('invalid field provided');
 
 				_.each(req.query.filters, function (value, key) {
-					query.where(key).equals(value ? value : null);
-					count.where(key).equals(value ? value : null);
+					query.where(key).equals(value);
+					count.where(key).equals(value);
 				});
 			}
 			count.exec(function (err, total) {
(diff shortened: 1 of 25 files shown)


admin/server/api/list.js
```diff
@@ -40,8 +40,8 @@ module.exports = function (req, res) {
 				if (!field) return sendError('invalid field provided');
 
 				_.each(req.query.filters, function (value, key) {
-					query.where(key).equals(value ? value : null);
-					count.where(key).equals(value ? value : null);
+					query.where(key).equals(value);
```
JedWatson: These aren't unnecessary, `null` vs `false` has an important impact on the query. Note that and include // ignore statements for these lines.

fields/types/Type.js
```diff
@@ -113,7 +113,7 @@ Field.prototype.getOptions = function () {
 		if (this.getProperties) {
 			Object.assign(this.__options, this.getProperties());
 		}
-		this.__options.hasFilterMethod = this.addFilterToQuery ? true : false;
+		this.__options.hasFilterMethod = this.addFilterToQuery;
```
JedWatson: This gets serialised, so the true/false value has importance

JedWatson:
Thanks for the work on this @LennyBoyatzis but after reviewing, a lot of these are important and I'm concerned including the rule in our linter config will discourage valid use of ternary statements to reduce conditions to simple values.

On that basis I'm rolling back my recommendation we add this rule, and will update #1797 accordingly.

PR #2297, opened 2016-02-16
[0.3] move browserify to dependencies

creynders:
When running keystone v0.3.16 with `--production` it hangs on `Error: Cannot find module 'browserify'`, i.e. I think it should be moved to the `dependencies`


creynders:
Aaargh, it was a hanging npm-shrinkwrap.json which messed up everything, leading to this error. Sorryz!

PR #2305, opened 2016-02-18
Fix URL linking in column view

### LINKED ISSUE #2255: URL field on admin UI turns into physical path when it's external
This isn't a big deal but when you save an external URL using the URL field type the Admin UI strips it of its protocol. This is fine for the text inside the anchor but not for the actual href attribute as it turns into a physical address and ends up like this **http://server/keystone/theUrl** rather than what the actual URL should be **http://theUrl**

I think this is just that this field doesn't have a case for it specifically. Something like this on the mixins/columns jade could be the solution:

``` jade
else if col.type == 'url'
		+column_link(col.field.format(item), item.get(col.path), true)
```

wmertens:
Fixes #2255

### DIFF
--- a/templates/mixins/columns.jade
+++ b/templates/mixins/columns.jade
@@ -33,7 +33,7 @@ mixin column(list, col, item)
 		else if col.type == 'boolean'
 			+column_boolean(value)
 		else if col.type == 'url'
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @wmertens!

PR #2308, opened 2016-02-19
Update importer.js

flyfy1:


morenoh149:
I would describe this as a grammar edit (for the issue name). But lgtm.

JedWatson:
Thanks for the fix @flyfy1

@webteckie you're right, we could do that & expose it the same way in Keystone to avoid breaking changes

PR #2316, opened 2016-02-21
RelationshipFilterFix

w01fgang:
fix for error: «can’t access to props», because ‘this’ is the anonymous
function of ‘_.each’


JedWatson:
:+1: thanks @w01fgang!

Issue #2322, opened 2016-02-23
Promise firing 'fullsetup' & not 'open' when using replica set


trentmillar:
I'm using 'mongo replica set' & not 'mongo' to connect to MongoDB. I found that .on('open') in  [openDatabaseConnection.js](https://github.com/keystonejs/keystone/blob/master/lib/core/openDatabaseConnection.js#L51) needs to run however the event [fullsetup](http://mongoosejs.com/docs/api.html#connection-js) is called instead.

I'd really appreciate a quick fix - I'd do a pull request but I think you'll see what I mean and push a fix.

Thanks
Trent

trentmillar:
hum, maybe i was a bit premature...

I may have a different issue,  [open](http://mongodb.github.io/node-mongodb-native/2.0/api/ReplSet.html#event:open) should work too

morenoh149:
feel free to clean that up via PR

trentmillar:
Actually I have another proposal.

there is no need for `mongo replica set` it indirectly limits what keystone allows to the mongo native driver.

instead why not add a new option `mongo options`, which would be a big help to me (and I'm sure others) who need to set `keepalive`, `sockettimout`, et.al.

how,,, get rid of the [if ... else](https://github.com/keystonejs/keystone/blob/master/lib/core/openDatabaseConnection.js#L33) and just modify the else to look like,

`keystone.mongoose.connect(keystone.get('mongo'), keystone.get('mongo options'));`

the only change would be for those using `mongo replica set` would get a depreciation warning in the console jargon. So I guess keep the **if** clause logic until a major release, then kill it.

Thoughts?

morenoh149:
sounds reasonable. 4.0 would be a breaking change anyway so that's a good opportunity to get this in.

trentmillar:
It would also be a good idea to add `mongo options` to the Keystone site's documentation - I could do this another day.

webteckie:
@trentmillar if something is missing in the docs can you please, at least, create an issue in the keystone-site project to track it?  tnx

trentmillar:
@webteckie sure, I'll add that shortly

trentmillar:
@webteckie [https://github.com/keystonejs/keystonejs-site/issues/74](https://github.com/keystonejs/keystonejs-site/issues/74)

webteckie:
:+1:

JedWatson:
Closed by #2334

PR #2334, opened 2016-02-25
adding 'mongo options' option and deprecating 'mongo replica set' option

trentmillar:
#2322 

- added depreciation warning when using `mongo replica set` 
- added `mongo options` option to replace this and provide access for passing options down to the native driver.


creynders:
@trentmillar cool, could you fix the linting errors?

trentmillar:
I see. Thanks.

JedWatson:
:+1: thanks @trentmillar!

PR #2347, opened 2016-03-01
Allow for deletion of LocalFiles

### LINKED ISSUE #2340: Localfiles type unable to remove data from db and image file from local directory
I have using Localfiles type to upload multiple files, it upload the files successfully and when later I tried to delete the files through the admin UI it did not remove from the data from the table and local directory. I do not get any errors when removing files. Here is the code snippet from the Model.

```
localImages: {
	type: Types.LocalFiles,
	dest: './public/images',
	prefix: '/images',
	allowedTypes: ['image/jpeg', 'image/png'],
	filename: function (item, file) {
		return item.key + '-' + file.originalname;
	},
	format: function (item, file) {
		return '<img src="' + file.href + '" style="max-width:100px" />';
	}
}
```

jacobj:
Fixes #2340

### DIFF
--- a/fields/types/localfiles/LocalFilesType.js
+++ b/fields/types/localfiles/LocalFilesType.js
@@ -117,7 +117,7 @@ localfiles.prototype.addToSchema = function() {
 		if (typeof element_id === 'undefined') {
 			value = values[0];
 		} else {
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @jacobj!

Issue #1934, opened 2015-12-07
babel-6 .babelrc build errors due to unknown options


Cspeisman:
```
keystone/admin/src/fields.js
keystone/admin/src/views/home.js
keystone/admin/src/views/list.js
keystone/admin/src/views/item.js
```
All encounter build errors due to me having a `.babelrc` file in my project root directory with configurations for babel6

I get error messages that look like this:
```
error building fields.js:
[BABEL] /Users/cspeisman/Sites/project_inception/node_modules/keystone/admin/src/fields.js: Unknown option: /Users/cspeisman/Sites/project_inception/.babelrc.presets while parsing file: /Users/cspeisman/Sites/project_inception/node_modules/keystone/admin/src/fields.js
```

ericelliott:
can you show the `.babelrc` file contents?

Cspeisman:
@ericelliott 
```
{
  "presets": ["es2015"]
}
'''

**Comment by ericelliott** [comment]
And what version of Keystone is your project using?

**Comment by Cspeisman** [comment]
I've seen the issue on both 3.15 and 3.16

**Comment by ericelliott** [comment]
What does your package.json look like?

**Comment by Cspeisman** [comment]
@ericelliott sorry about the delay. Here is my package.json file: 

```
{
  "name": "project_inception",
  "version": "1.0.0",
  "description": "internal DD playbook",
  "main": "index.js",
  "scripts": {
    "start": "node ./keystone.js"
  },
  "author": "Corey Speisman <cspeisman@gmail.com> (http://github.com/cspeisman)",
  "license": "ISC",
  "dependencies": {
    "keystone": "^0.3.16"
  },
  "devDependencies": {
    "babel-cli": "^6.3.15",
    "babel-preset-es2015": "^6.3.13"
  }
}
```

ericelliott:
Looks like you're trying to use the 6.x Babel API with a version of keystone that isn't compatible with it. Two ways forward:

* Use the (experimental) 0.4.x Keystone, or..
* Use the 5.x.x Babel line (including all Babel utils, as well)

Cspeisman:
:+1: Thanks @ericelliott 
Is it worth opening a ticket for 6.x Babel support in 0.3.x Keystone or keep it as is?

ericelliott:
I think that's a question for @JedWatson. IMO, our priority should be to release the next stable version, instead.

Which means we should decide the scope of what's going into it and start working toward a finished milestone ASAP. =)

[...]

Issue #1865, opened 2015-11-24
JSX transform fails when keystone is a dependency of a global package


erg0dic:
```
{ "keystone": "0.3.15" }
```
I have a project my-project that has keystone as a dependency. my-project is packaged and pushed to a private npm registry. For deployment, I do an npm i my-project against the private npm registry. I have a "bin" entry in package.json (https://docs.npmjs.com/files/package.json#bin) that symlinks the entry point to my-project. 

When running my-project via that symlink, I experience JSX parse errors, the same symptoms as #1411 & #1514.

It is not clear to me whether this is an issue with keystone or something upstream (babel? browserify?). I am more than happy to continue investigating but I need some direction.

erg0dic:
I have published an npm module that demonstrates this issue.

```
npm install -g keystone-issue-1865
node $(npm root -g)/keystone-issue-1865/keystone.js
```

I get the following output:
<img width="1179" alt="screen shot 2015-12-03 at 10 50 06 pm" src="https://cloud.githubusercontent.com/assets/5550520/11581463/3f4b2af0-9a10-11e5-9c2b-565fb053898c.png">

ericelliott:
Thanks for publishing the module. =) Confirmed.

erg0dic:
I believe this is due to browserify ignoring files under node_modules/
https://github.com/babel/babelify#why-arent-files-in-node_modules-being-transformed

If I add babelify to the list of browserify transforms in package.json (as suggested in the above link), I no longer see these errors.

PR shortly

JedWatson:
Closed by #1913

Issue #1514, opened 2015-07-07
Error building files on fresh install


grepped:
I'm getting these errors with a fresh install of keystone with yeoman. What could be causing them?

>2015-07-07 14:07:42 error building views/home.js:
>Parsing file /usr/local/lib/node_modules/keystone/admin/src/views/home.js: Unexpected token (11:4)
>2015-07-07 14:07:43 error building views/item.js:
>Parsing file /usr/local/lib/node_modules/keystone/admin/src/views/item.js: Unexpected token (46:9)
>2015-07-07 14:07:43 error building views/list.js:
>Parsing file /usr/local/lib/node_modules/keystone/admin/src/views/list.js: Unexpected token (25:4)
>2015-07-07 14:07:66 error building fields.js:

morenoh149:
can you paste your home and token.js

grepped:
Here is my home.js file http://pastebin.com/PEes5TpD

I'm unable to locate token.js, can you specify the path?

morenoh149:
sorry meant item.js and list.js

grepped:
/usr/local/lib/node_modules/keystone/admin/src/views/item.js
http://pastebin.com/W1xherTv

/usr/local/lib/node_modules/keystone/admin/src/views/list.js
http://pastebin.com/Hu6ub8ER

tomasbillborn:
I got these errors all of a sudden and could fix them by

https://github.com/keystonejs/keystone/issues/1411

if you update to creynders answer 

var bundles = {
    fields: browserify('./fields.js', 'FieldTypes'),
    home: browserify('./views/home.js'),
    item: browserify('./views/item.js'),
    list: browserify('./views/list.js')
};

if it is not obvious its the missing ./ in the path that creates the problem.
I think its just / in keystone atm.

grepped:
Just prepended './' to all of the bundle urls in /node_modules/keystone/admin/app/static.js. I'm getting the same issue.

http://pastebin.com/d6pmAGH1

grepped:
Is this the same as creynders answer? The error I'm getting is "Parsing file : Unexpected token", the error message in the issue you linked is "Cannot find module".
Still haven't gotten it to work, tried both the './' and the '../src/' paths

tomasbillborn:
It solves the error with the paths for the items above being incorrect, exactly what the token issue is I can't help you with. I got multiple errors trying to go with the yeoman set up. So I used the tutorial under getting instead.

[...]

Issue #1411, opened 2015-05-18
cannot find module 'views/home.js' ...


freder:
after switching from `0.3.8` to `0.3.9` I get the following messages:

```
------------------------------------------------
KeystoneJS Started:
project is ready on port 3000
------------------------------------------------

2015-05-18 15:05 error building views/home.js:
Cannot find module 'views/home.js' from '/Volumes/Macintosh HD 2/Projects/project/platform/node_modules/keystone/admin/src'
2015-05-18 15:05 error building views/item.js:
Cannot find module 'views/item.js' from '/Volumes/Macintosh HD 2/Projects/project/platform/node_modules/keystone/admin/src'
2015-05-18 15:05 error building views/list.js:
Cannot find module 'views/list.js' from '/Volumes/Macintosh HD 2/Projects/project/platform/node_modules/keystone/admin/src'
```

poksme:
I get the same messages when trying to access the admin.

JedWatson:
Can you see if the file actually exists? I've just tried a fresh install from npm (via the `yo keystone` generator) and I can't reproduce the issue.

Also try `rm -rf node_modules && npm install` - this may fix it for you

poksme:
Yes the 3 files exists. I tried removing the node_modules directory and run npm install again but it did not fix it.

BerkeleyTrue:
@JedWatson running into the same issue in 0.3.9 on fresh install from npm. The files do exist in the directory.

BerkeleyTrue:
issue disappears when reverting back to 0.3.8

JedWatson:
What platform are you folks using? Because I would love to fix this, but can't until I can reproduce it.

Otherwise could someone who _can_ reproduce it please debug it a bit further?

BerkeleyTrue:
OSX 10.10.3
io.js 1.6.4

freder:
OS X 10.10.3
node 0.12.2

freder:
well, in `admin/app/static.js` changing

```javascript
var bundles = {
	fields: browserify('fields.js', 'FieldTypes'),
	home: browserify('views/home.js'),
	item: browserify('views/item.js'),
	list: browserify('views/list.js')
};
```

to

```javascript
var bundles = {
	fields: browserify('fields.js', 'FieldTypes'),
	home: browserify('../src/views/home.js'),
	item: browserify('../src/views/item.js'),
	list: browserify('../src/views/list.js')
};
```

did it for me.

[...]

PR #2349, opened 2016-03-02
Upgrade Babel to V6

mekanics:
this should fix some error building:
- views/home.js
- views/item.js
- views/list.js
- fields.js:

I think it's related and fixes the following issues: 
#1934
#1865
#1514
#1411
and maybe more... 

would be nice if @JedWatson could make a new version v0.3.17 with this fix


morenoh149:
:100:

mikestopcontinues:
I would very much appreciate this! 😇

JedWatson:
Thanks @mekanics - on that now!

We _still_ need the `object-assign` plugin for v6, I'll add that back in after merging.

mekanics:
🍻

PR #2361, opened 2016-03-03
Migrate underscore to lodash (fixes #2287)

### LINKED ISSUE #2243: Underscore error on LocalFilesField.js
I use `Types.LocalFiles` and when I select file in admin area there is an error:
`Uncaught TypeError: _underscore2.default.contains is not a function`
This is an Underscore error.
Actually `_.contains` has an alias `_.includes` and is work.
Should I try to implement this function without `underscore`? Or I can use `lodash`? Or simply change `_.contains` with `_.includes`?

### LINKED ISSUE #2287: Use lodash instead of underscore?
I've seen a lot of benchmarks that shows underscore is slower than lodash. Can we use lodash instead?

jacargentina:
This also fixes #2243


admin/server/app/createStaticRouter.js
```diff
@@ -48,14 +48,14 @@ module.exports = function createStaticRouter (keystone) {
 	};
 
 	/* Configure router */
-	router.use('/styles', less(path.resolve(__dirname + '../../../public/styles'), lessOptions));
-	router.use('/styles/fonts', express.static(path.resolve(__dirname + '../../../public/js/lib/tinymce/skins/keystone/fonts')));
+	router.use('/styles', less(path.resolve(__dirname + '/../../public/styles'), lessOptions));
```
JedWatson: Is there a reason these path changes are included in this PR?

JedWatson:
So tests are failing because something is looking for underscore... and I'm not sure why those paths have changed, but I'm really keen to merge this before changing anything else (otherwise the conflicts will be a nightmare) so I'm going to do that and work on it locally to make sure everything's good. Ping me on slack when you're around.

Also, thanks! :smile:

victoriafrench:
found it and running it through travis on my fork... send over pr in a sec

victoriafrench:
@JedWatson @jacargentina passed travis, sending over

JedWatson:
Thanks @victoriafrench - really appreciate you picking this up and running with it!

Issue #2291, opened 2016-02-15
[.04] Admin UI e2e test status


webteckie:
This issue covers the current status of the ui/functional e2e tests coverage.  Need lots of help here for anyone that's interested!  We are using [nightwatch](http://nightwatchjs.org/guide) to handle e2e testing. All Admin UI e2e tests are under `test/e2eTests/adminUI`  These are broken into groups and group related tests within those groups.  Both the groups and tests have a sequence number in them to control the order of execution by nightwatch.  To run the tests locally use `npm run test-e2e`.  Make sure you have done a successful `npm install` for the keystone directory first!  If you pick a group/test please mention it here to avoid duplication.  Ideally we should focus both on positive and negative testing but at this point positive testing is more important to quickly detect regressions!

   - [X] group001Login
   - [ ] group002Home
   - [ ] group003List
   - [ ] group004Item
   group005Fields
       - [ ] azurefile
       - [ ] boolean
       - [ ] cloudinaryimage
       - [ ] cloudinaryimages
       - [ ] code
       - [ ] color
       - [ ] date
       - [ ] datearray
       - [ ] datetime
       - [ ] email
       - [ ] embedly
       - [ ] geopoint
       - [ ] html
       - [ ] key
       - [ ] localfile
       - [ ] localfiles
       - [ ] location
       - [ ] markdown
       - [ ] money
       - [ ] name
       - [ ] number
       - [ ] numberarray
       - [ ] password
       - [ ] relationship
       - [ ] s3file
       - [ ] select
       - [ ] text
       - [ ] textarea
       - [ ] textarray
       - [ ] text
       - [ ] url

   - [ ] group006Relationships

jstockwin:
@webteckie Mark me down for some more simple ones i.e. `textarea` and `url` for now. 

Out of interest, and maybe for future reference, how many checks need to happen on the more complex types? For example, markdown should display with buttons to make text bold and add links etc. Do we check these are there, and working, and then maybe also somehow check the output html is correct, or do we simply just add some text in the field, without many extra checks.

webteckie:
@jstockwin alright will do!  You can add as much testing as you need.  Currently we've been adding simple testing just to get the basic stuff covered.  But we should really try to cover as much of it as possible.

jstockwin:
@webteckie #2709 should be okay to merge now. I'll probably do datetime later, so put me down for a few more. Uhh html, markdown and password? 

[...]

PR #2366, opened 2016-03-04
Build failure fix

victoriafrench:
fix for #2287 #2361 
lodash does not have a `where` function, switched to `filter`. 

associate #2291 
Group5 Failure: Fixed test case where the selector for the Admin Name Field should have been looking at the 11th child not the 12th.


victoriafrench:
@JedWatson this should fix travis. (see slack #general). My instance of travis isn't running sauce and I hadn't locally run e2e tests. Found the failures in there and fixed. npm run tests-all gets green on my local.

JedWatson:
Awesome. Thanks @victoriafrench!

PR #2376, opened 2016-03-05
Added eslint rule New cap

LennyBoyatzis:
Progress commit

As per Review .eslintrc #1797

### DIFF
--- a/admin/client/components/Checkbox.js
+++ b/admin/client/components/Checkbox.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import blacklist from 'blacklist';
 import classnames from 'classnames';
-import Color from 'color';
+import colorConverter from 'color';
 import E from '../constants';
 
 var Checkbox = React.createClass({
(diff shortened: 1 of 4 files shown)


admin/client/components/Checkbox.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import blacklist from 'blacklist';
 import classnames from 'classnames';
-import Color from 'color';
+import colorConverter from 'color';
```
JedWatson: couldn't this be `color`?

LennyBoyatzis: There is already a variable called color which stores information about font color. It was either change this variable or the name of the color package. See code snippet

`		let checkedColor = colorConverter('#3999fc');

		let background = (checked && !readonly) ? checkedColor.hexString() : 'white';
		let borderColor = (checked && !readonly) ? 'rgba(0,0,0,0.15) rgba(0,0,0,0.1) rgba(0,0,0,0.05)' : 'rgba(0,0,0,0.3) rgba(0,0,0,0.2) rgba(0,0,0,0.15)';
		let boxShadow = (checked && !readonly) ? '0 1px 0 rgba(255,255,255,0.33)' : 'inset 0 1px 0 rgba(0,0,0,0.06)';
		let color = (checked && !readonly) ? 'white' : '#bbb';
		let textShadow = (checked && !readonly) ? '0 1px 0 rgba(0,0,0,0.2)' : null;
`

mxstbr:
What's the blocker here?

JedWatson:
Discussed w/ @LennyBoyatzis in person, we're going to bail on adding this eslint rule. I've come to think it's as much misdirection as clarification.

PR #2408, opened 2016-03-11
Fix #2394 - tighten redirection destination sanitation

### LINKED ISSUE #2394: [0.3] Open redirects as in keystone/signin?from=ftp://example.com considered an attack vector
I got notification today that one of the keystone sites I'm developing is considered having malicious content, of the flagged URL's two contained a signin redirect and the other two were the results of the redirect. When checking Google's [Identify the vulnerability page](https://support.google.com/webmasters/answer/3013681) it says:

>Open redirects are coded with the intention for the URL structure to allow the addition of another URL so users can reach a useful file or webpage on the site. For example:

>http://example.com/page.php?url=http://example.com/good-file.pdf
>Hackers can abuse open redirects by adding their spammy or malware page to the site’s open redirect, similar to this:
>http://example.com/page.php?url=<malware-attack-site>

The solution to this would be

>whether your code can prohibit off-domain redirects, or if you can sign the redirect so that only those with properly hashed URLs and the cryptographic signature can be redirected.

We need to fix this. What would be the best approach? Sifting off-domain redirects or?

@JedWatson @dcousens @jossmac

creynders:
As discussed. It falls back to `/keystone` if the redirection destination contains a host, protocol or auth part and removes query strings.

### DIFF
--- a/admin/routes/views/signin.js
+++ b/admin/routes/views/signin.js
@@ -1,5 +1,6 @@
 var keystone = require('../../../');
 var session = require('../../../lib/session');
+var url = require('url');
 
 exports = module.exports = function(req, res) {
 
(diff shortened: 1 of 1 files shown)


JedWatson:
Cool, thanks @creynders!

PR #2411, opened 2016-03-12
#2405: align the first list view column header with its value.

### LINKED ISSUE #2405: [0.4] First column alignment in list view off
I think the first column header should be aligned with the column value.  This leaves real estate in case we want to (and we should) add a header check box to delete multiple items in a page.

image

webteckie:
Fix #2405  Now the alignment looks like this:

image

### DIFF
--- a/admin/client/components/ItemsTable.js
+++ b/admin/client/components/ItemsTable.js
@@ -28,16 +28,21 @@ const ItemsTable = React.createClass({
 		return <colgroup>{cols}</colgroup>;
 	},
 	renderHeaders () {
-		var cells = this.props.columns.map((col, i) => {
-			// span first col for controls when present
-			var span = 1;
-			if (!i) {
-				if (this.props.list.sortable) span++;
-				if (!this.props.list.nodelete) span++;
-			}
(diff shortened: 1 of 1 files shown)


JedWatson:
Nice @webteckie. I like it :)

JedWatson:
Oops - this has broken our e2e tests, mind fixing that up?

```
----------------------------------------------------
TEST FAILURE: 1 assertions failed, 162 passed (6m 36s)
 ✖ group003List/uiTest001ListView
   - List view must have a name column header
     Expected element <.Table > thead:nth-child(2) > tr:nth-child(1) > th:nth-child(1)> text to equal: "Name" - Expected "equal 'Name'" but got: ""
   SKIPPED:
   - List view must have an email column header
   - List view must have an Is Admin column header
   - List view items must a delete icon
   - List view user item must have a name value
   - List view user item must have a value in the email column
   - List view user item must have a value in the Is Admin column
```

webteckie:
Crap...I thought I ran it...I must have done it in the wrong directory :-) Will take care of that!

Issue #2229, opened 2016-02-05
[0.4] Searching does not reset paging in list view


alobodig:
When searching a list, if the list was initially viewing a page, the UI tries to display that page of the search results. This causes the message "No __ found" when the result set actually does contain items. 

![video](http://g.recordit.co/Sbv8ucpOZA.gif)

snowkeeper:
Looks like the internal page number needs to be reset to 1.

mxstbr:
Fixed with #2415

danielgormly:
I believe this is a problem again with 4.0.0!

PR #2415, opened 2016-03-14
Fix issue #2229: [0.4] Searching does not reset paging in list view

suryagh:
Fix for issue #2229

**Problem** 
Issue was that the the query was skipping the `results` record even if the total `records count` was less than the `skip count`. 

**Solution** 
Modified the query to `skip`, only if the `records count` is greater than `skip count`. Also, had to replace `async.series` with `async.waterfall` so that the query count can be passed to the `next` function.

**Behaviour with this fix:**

keystone-2


JedWatson:
Nice fix @suryagh, thanks!

PR #2416, opened 2016-03-14
Make sure that unit tests run with a new db.

webteckie:
This is more of a problem with local environments as sometimes stale data in the database causes tests to fail resulting in long troubleshooting sessions!


JedWatson:
@webteckie nice - I was thinking about how to do this just yesterday. You read my mind :)

PR #2421, opened 2016-03-15
Change lodash .contains to .includes

mxstbr:
lodash.contains was deprecated in v4 and replaced with .includes, but we were still using .contains in the cloudinaryimages field which made live preview borked.


JedWatson:
Thanks @mxstbr!

Issue #1255, opened 2015-03-25
Theming the Admin UI


davideberlein:
Hello,

@JedWatson in issue #442 you mentioned thinking about how the admin ui can be themed.
Have you made any progress on this topic?
I think the bootstrap approach having a variables.less would be an easy and functional solution.
The user could pass a path to his own variables.less file in which he can overwrite any variables as keystone.init() option.
Since keystone is based on bootstrap thisnvariables.less file propably would consist of 90% bootstrap varibales.

What do you think?

Best regards

David

ctaepper:
here is what i do:

- create custom less file and import default keystone.less into it adding my custom styles and overwriting keystone bootstrap variables
```
@import "../../node_modules/keystone/public/styles/keystone";
// change keystone navbar background
@navbar-keystone-bg: green;
```
- gulp task to convert that file to css
- add this route to express config to overwrite the keystone default stylesheet with your own
```javascript
app.get('/keystone/styles/keystone.min.css', function (req, res) {
    res.sendfile('./dist/styles/keystone.min.css');
});
```

it is obviously some form of workaroud, but it pretty much gets the job done

davideberlein:
Thanks for the feedback @ctaepper!
Looks like a doable workaround.. however having a proper integration would even be better..
If I have the time I might look into the keystone source and check if I can find a good solution

mxstbr:
Closing as a duplicate of #220!

Issue #2242, opened 2016-02-07
Allow for configurable Admin UI custom styling


webteckie:
This would allow apps to easily custom style the Admin UI templates.  These custom styles can be configured at init time.  The Admin UI client would then append this styles to the templates.  For example:
```
'adminui custom styles': {
     signin: custom-style-url,
     home: custom-style-url,
     list: custom-style-url,
     item: custom-style-url,
}
```

To go with this, it may be nice to also allow the list and fields to accept a custom css class.  For example:
```
var MyList = new keystone.List('MyList', {
    cssclass: 'mylist-custom-class'
});
MyList.add({
    myfield: {type: XXX, cssclass: 'mylist-name-css-class'}
});
```

webteckie:
I've addressed this via #2462.

PR #2426, opened 2016-03-15
[v0.3] Change the color of the navbar in the admin

dam1:
Very simple setting, but useful to know which env you are working on.

in keystone config

`keystone.init({'admin color': '#009933', ... `

related #1255

### DIFF
--- a/lib/core/render.js
+++ b/lib/core/render.js
@@ -91,7 +91,8 @@ function render(req, res, view, ext) {
 			skin: keystone.get('wysiwyg skin') || 'keystone',
 			menubar: keystone.get('wysiwyg menubar'),
 			importcss: keystone.get('wysiwyg importcss') || ''
-		}
+		},
+		admincolor: keystone.get('admin color')
 	};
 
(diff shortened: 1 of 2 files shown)


creynders:
:+1:

webteckie:
Wouldn't it better to support #2242 so that other customization is possible?

creynders:
@webteckie sure, but this is a PR and the other seems to involve a lot more work. But obviously it would be great if you could come up with a backwards compatible solution, i.e. small change here makes it compatible with what you have in mind in #2242

webteckie:
@creynders I may be missing something here but at dont see the value of this small tweak at the expense of another keystone config.  Am I missing something?

mxstbr:
Agreed @webteckie, I don't think this is necessary. Lets wait for #2242 instead of adding a global config var.

mxstbr:
We decided to go with #2462, which allows you to change everything.

Thanks so much for contributing, we really appreciate it!

PR #2437, opened 2016-03-17
Update and test select field

mxstbr:
Update and comprehensively test the select field based on the provided specs

### DIFF
--- a/fields/types/select/SelectType.js
+++ b/fields/types/select/SelectType.js
@@ -144,16 +144,17 @@ select.prototype.validateInput = function (data, callback) {
 select.prototype.validateRequiredInput = function (item, data, callback) {
 	var value = this.getValueFromData(data);
 	var result = false;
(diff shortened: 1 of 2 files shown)


fields/types/select/SelectType.js
```diff
@@ -143,7 +143,19 @@ select.prototype.validateInput = function (data, callback) {
...
-	var result = value !== undefined || (value === undefined && item.get(this.path)) ? true : false;
+	var result = false;
+	if (typeof value === 'string' && this.numeric) {
+		value = utils.number(value);
+	}
+	if (value === undefined) {
+		if (item.get(this.path)) {
+			result = true;
+		}
+	} else if (typeof value === 'string' || typeof value === 'number') {
```
JedWatson: I'm guessing we're not looking in `this.map` because if the value made it this far, that has already been checked, and we just need to ensure a value is provided?

In which case, is there value in converting to a number at all? (ln. 148)

PR #2445, opened 2016-03-17
Fields update

mxstbr:
Small updates across the board.

### DIFF
--- a/fields/types/date/test/server.js
+++ b/fields/types/date/test/server.js
@@ -40,10 +40,9 @@ exports.testFieldType = function (List) {
 			});
 		});
 
-		// TODO This shouldn't fail
-		it('should validate dates', function (done) {
-			List.fields.date.validateInput({ date: new Date() }, function (result) {
-				demand(result).be(true);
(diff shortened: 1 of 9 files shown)


fields/types/date/test/server.js
```diff
@@ -8,36 +9,106 @@ exports.initList = function (List) {
...
+	describe('validateInput', function () {
+		it('should validate date strings', function (done) {
+			List.fields.date.validateInput({ date: '2015-01-01' }, function (result) {
+				demand(result).be(true);
+				done();
+			});
+		});
+
+		it('should invalidate unformatted dates', function (done) {
+			List.fields.date.validateInput({ date: new Date(2015, 1, 1) }, function (result) {
```
mxstbr: The reason this test fails is because the date is a string in the wrong format (something like "Thu 28th March 2016 blabla") which moment doesn't want because we have a different default format string. Is this intended behaviour?

mxstbr:
Updated, but unsure about the date field updateItem tests.

PR #2453, opened 2016-03-18
Update date field

mxstbr:


fields/types/date/DateType.js
```diff
@@ -93,6 +93,8 @@ date.prototype.moment = function (item) {
  */
 date.prototype.parse = function (value, format) {
 	var m = this.isUTC ? moment.utc : moment;
+	// TODO Check should maybe be if (typeof value === 'string')
+	// use the parseFormatString. Ever relevant?
 	if (typeof value === 'number' || value instanceof Date) {
```
mxstbr: This should maybe be

```JS
if (typeof value === 'string') {
  return m(value, format || this.parseFormatString);
} else {
  return m(value);
}
```

instead of the other way around – does that ever matter?

fields/types/date/test/type.js
```diff
@@ -67,14 +67,14 @@ exports.testFieldType = function (List) {
...
-		it('should clear the value when passed undefined', function (done) {
+		it('should not clear the value when a value exists and passed undefined', function (done) {
 			var testItem = new List.model({
 				date: '2015-01-01',
 			});
 			List.fields.date.updateItem(testItem, {
 				date: undefined,
 			}, function () {
-				demand(testItem.date).be(null);
+				demand(testItem.date).not.be.null();
```
mxstbr: This test was a bit weird, to stay consistent with the other fields I've now changed it and fixed the implementation – was this intended behaviour?

JedWatson: Good catch, that was not the correct behaviour.

fields/types/date/DateType.js
JedWatson: :shrug:

Issue #2401, opened 2016-03-10
Date resets in Datetime field on admin panel


w01fgang:
Description
-------------------
If I change the time after the date, the date field lost changes.
It does because the `DateInput` component has his own state and don't pass changes to parent component `DatetimeField`, so props and state is different.
![datetime](https://cloud.githubusercontent.com/assets/8686751/13679068/2606ecbc-e6f2-11e5-98d7-0be4730943c2.png)
When I change the time, parent component rerenders and pass old `props`.
Maybe `DateInput` must be stateless?

mxstbr:
Good catch, I can replicate this – we'll take a look!

mxstbr:
Here's a gif replication:

![datetime-bug](https://cloud.githubusercontent.com/assets/7525670/13869128/43ba4a72-ed25-11e5-8310-4ade40a70933.gif)

geloescht:
Can't replicate in v0.4, is it fixed? @mxstbr

mxstbr:
Yeah, I fixed that while I was at it, thanks for reminding me!

PR #2454, opened 2016-03-18
Datetime field update

mxstbr:
This is **not a one-click-merge**, I'm not sure if this is the right way to go here. The `updateItem` and `parse` functions looked exactly the same as the (old) `date` ones, so I reused them – for the `parse` function, I think that's fine. (Merge #2453 before merging this)

There's #2401, which leads me to believe that the `updateItem` was simply copy & pasted from `date`, without regards that this made up of two fields not only one. **I'll do a functional test** to make sure there are no regressions there.

EDIT: The bug still exists with this change, lets fix it with this PR too.

### DIFF
--- a/fields/types/datetime/DatetimeType.js
+++ b/fields/types/datetime/DatetimeType.js
@@ -80,7 +80,23 @@ datetime.prototype.inputIsValid = function (data, required, item) {
 /**
  * Updates the value for this field in the item from a data object
  */
-datetime.prototype.updateItem = DateType.prototype.updateItem;
+datetime.prototype.updateItem = function (item, data, callback) {
+	if (!(this.path in data)) {
+		return process.nextTick(callback);
+	}
+	// var newValue = this.parse(data[this.path]);
+	var value = this.getInputFromData(data);
+	console.log('update', value);
+	if (typeof value === 'object') {
+		if (typeof value.date === 'string' || value.first === null) {
+			item.set(this.path.date, value.date);
+		}
+		if (typeof value.time === 'string' || value.time === null) {
+			item.set(this.path.time, value.time);
+		}
+	}
+	process.nextTick(callback);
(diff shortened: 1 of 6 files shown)


mxstbr:
WIP: Made the `DayInput` a controlled component as per discussion with @jed. It now calls `this.props.onChange` when a day is selected. I also made it work better with the keyboard (as discussed with @twalve), check this out:

datepicker

**Todo**

- [ ] Save value on form submit (I broke that)
- [ ] Test that it works with a custom format string
- [ ] Test that the datetime field work again (ref #2401)
- [ ] *Maybe fix timezone bug (ref #1345)*

mxstbr:
Finally fought through it, this is ready for review!

fields/types/datetime/DatetimeType.js
```diff
@@ -84,16 +81,22 @@ datetime.prototype.inputIsValid = function (data, required, item) {
...
-		if (!oldValue || !newValue.isSame(oldValue)) {
-			item.set(this.path, newValue.toDate());
+	// Get the values from the data
+	var value = this.getInputFromData(data);
+	if (value !== undefined) {
+		if (value !== null && value !== '') {
+			// If the value is not null, empty string or undefined, parse it
+			var newValue = this.parse(value, this.formatString);
+			// If it's valid and not the same as the last value, save it
+			if (newValue.isValid()) {
```
mxstbr: This should possibly not be in the updateItem, since the validation should already be done by the point it reaches this in `validateInput`/`validateRequiredInput`?

mxstbr:
Moved validation out of `updateItem` and into a custom `validateInput` for `Datetime`, which makes more sense imo. This should now finally be finished for real and ready to merge!

JedWatson:
Looks good @mxstbr

PR #2465, opened 2016-03-19
Fixed issue #2427 - wysiwyg cloudinary image upload https

### LINKED ISSUE #2427: wysiwyg cloudinary image upload https
Wondering if there is a configuration option for making the inserted image source be https when inserting a cloudinary image using the wysiwyg option (enabled by the 'wysiwyg cloudinary images': true option) for the html data type .

I've tried doing

keystone.set('cloudinary secure', true);

but the inserted url is still http.

Would be super duper great to have this because otherwise would have to train users to view source and add https to every image they upload--something easily forgotten. :)

suryagh:
Fixed #2427. 

Issue: 
- Set the `keystone.set('cloudinary secure', true);` option and start the application.
- Upload an image to the cloudinary using the _wwysiwyg editor_.
- Click the `source code` button of the  _wwysiwyg editor_.
- The cloudinary image url used is `http` instead of `https`.

Solution: 

- The `cloudinary uploader` module does return both the `secure` and `non-secure` urls of the uploaded image. Modified the code to return `secure` url to the client if `cloudinary secure` option is set to `true`

### DIFF
--- a/admin/server/api/cloudinary.js
+++ b/admin/server/api/cloudinary.js
@@ -15,7 +15,7 @@ module.exports = {
 					if (result.error) {
 						res.send({ error: { message: result.error.message } });
 					} else {
-						res.send({ image: { url: result.url } });
+						res.send({ image: { url: (keystone.get('cloudinary secure') === true) ? result.secure_url : result.url } });
 					}
 				};
 


JedWatson:
Looks good, thanks @suryagh!

PR #2472, opened 2016-03-21
Add some general doc about fields

mxstbr:


JedWatson:
Good start @mxstbr :+1:

We should follow up with filtering as well.

Issue #1776, opened 2015-11-08
Implementing Filter UI Components for Field Types in 0.4


JedWatson:
Currently WIP, completion of each field type also means checking the back-end implementation of filtering in the FieldType.

- [x] boolean
- [x] code
- [x] color
- [x] date 
- [x] datearray (#2544)
- [x] datetime
- [x] email
- [x] geopoint (#2546)
- [x] html
- [x] key
- [x] location
- [x] markdown
- [x] money (#2480)
- [x] name (#2545)
- [x] number
- [x] numberarray (#2511)
- [x] password
- [x] relationship
- [x] select
- [x] text
- [x] textarea
- [x] textarray (#2510)
- [x] url

----

- [ ] *azurefile*
- [ ] *cloudinaryimage*
- [ ] *cloudinaryimages*
- [ ] *embedly*
- [ ] *localfile*
- [ ] *localfiles*
- [ ] *s3file*

*Not critical for initial release*

mxstbr:
Testing the already done `addFilterToQuery` methods now!

mxstbr:
Have to fix text filtering. The text fields `addFilterToQuery` checks the `filter.mode` to be `startsWith`, but the admin interface sets it to `beginsWith` so that filter doesn't work.

mxstbr:
Also need to improve array filtering, it works for now but only filters for "at least one element in the array does x". It should also be able to filter for "all elements in the array do x" or "y amount of elements in the array do x".

mxstbr:
All done!

PR #2480, opened 2016-03-22
addFilterToQuery tests

mxstbr:
~~**WIP, do not merge yet**~~

Adding tests for all existing `addFilterToQuery` methods, ref #1776.

- [x] boolean
- [x] code
- [x] color
- [x] date 
- [x] email
- [x] html
- [x] key
- [x] money
- [x] number
- [x] location
- [x] password
- [x] relationship
- [x] select
- [x] text
- [x] textarea
- [x] url

*(EDIT: 62 tests in one PR hell yea)*

### DIFF
--- a/fields/types/number/NumberType.js
+++ b/fields/types/number/NumberType.js
@@ -51,7 +51,7 @@ number.prototype.addFilterToQuery = function (filter, query) {
 	query = query || {};
 	if (filter.mode === 'equals' && !filter.value) {
 		query[this.path] = filter.inverted ? { $nin: ['', 0, null] } : { $in: ['', 0, null] };
-		return;
+		return query;
 	}
 	if (filter.mode === 'between') {
 		var min = utils.number(filter.value.min);
@@ -59,7 +59,7 @@ number.prototype.addFilterToQuery = function (filter, query) {
 		if (!isNaN(min) && !isNaN(max)) {
 			query[this.path] = filter.inverted ? { $gte: max, $lte: min } : { $gte: min, $lte: max };
 		}
-		return;
+		return query;
 	}
 	var value = utils.number(filter.value);
 	if (!isNaN(value)) {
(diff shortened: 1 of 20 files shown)


fields/types/text/TextType.js
```diff
@@ -37,7 +37,7 @@ text.prototype.addFilterToQuery = function (filter, query) {
 	query = query || {};
 	if (filter.mode === 'exactly' && !filter.value) {
 		query[this.path] = filter.inverted ? { $nin: ['', null] } : { $in: ['', null] };
-		return;
+		return query;
```
mxstbr: @JedWatson I'm not sure but I think this was omitted on accident, changing `query` but then not returning it seems like a bug.

I added tests for this, but if this is the intended behaviour I'll happily revoke this change and update the tests accordingly.

mxstbr:
Just noticed the `number` field also has an `addFilterToQuery` method – will do that too before we merge!

fields/types/number/NumberType.js
```diff
@@ -51,7 +51,7 @@ number.prototype.addFilterToQuery = function (filter, query) {
 	query = query || {};
 	if (filter.mode === 'equals' && !filter.value) {
 		query[this.path] = filter.inverted ? { $nin: ['', 0, null] } : { $in: ['', 0, null] };
-		return;
+		return query;
```
mxstbr: Same question as above for the text field.

fields/types/number/NumberType.js
```diff
@@ -59,7 +59,7 @@ number.prototype.addFilterToQuery = function (filter, query) {
 		if (!isNaN(min) && !isNaN(max)) {
 			query[this.path] = filter.inverted ? { $gte: max, $lte: min } : { $gte: min, $lte: max };
 		}
-		return;
+		return query;
```
mxstbr: Same question as above for the text field.

mxstbr:
- Money field now uses number field `addFilterToQuery´ (tested this in the admin ui and it works)

JedWatson:
For the record: **so** stoked with this PR

PR #2490, opened 2016-03-22
Password type validation method updates

mxstbr:
~~**WIP DO NOT MERGE YET**~~

Updates to the password field validation methods. I'm not sure how `validateRequiredInput` should behave, and if it behaves correctly at the moment – any comments @JedWatson?

### DIFF
--- a/fields/types/password/PasswordType.js
+++ b/fields/types/password/PasswordType.js
@@ -123,9 +123,16 @@ password.prototype.validateInput = function (data, callback) {
 	var result = true;
 	var confirmValue = this.getValueFromData(data, '_confirm');
 	var passwordValue = this.getValueFromData(data);
-	if (passwordValue !== undefined && confirmValue !== undefined && passwordValue !== confirmValue) {
+	if (passwordValue !== undefined
+		&& confirmValue !== undefined
+		&& passwordValue !== confirmValue) {
 		result = false;
 		detail = 'passwords must match';
+	}
+	if ((passwordValue === undefined || confirmValue === undefined)
+		&& passwordValue !== confirmValue) {
+		result = false;
+		detail = 'both fields need to be filled';
 	}
 	// TODO: we could support a password complexity option (or regexp) here
 	utils.defer(callback, result, detail);
--- a/fields/types/password/PasswordType.js
+++ b/fields/types/password/PasswordType.js
@@ -123,16 +123,10 @@ password.prototype.validateInput = function (data, callback) {
 	var result = true;
 	var confirmValue = this.getValueFromData(data, '_confirm');
 	var passwordValue = this.getValueFromData(data);
-	if (passwordValue !== undefined
-		&& confirmValue !== undefined
+	if (confirmValue !== undefined
 		&& passwordValue !== confirmValue) {
 		result = false;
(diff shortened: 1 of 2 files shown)


fields/types/password/PasswordType.js
```diff
@@ -123,10 +123,17 @@ password.prototype.validateInput = function (data, callback) {
...
 	var confirmValue = this.getValueFromData(data, '_confirm');
 	var passwordValue = this.getValueFromData(data);
-	if (passwordValue !== undefined && confirmValue !== undefined && passwordValue !== confirmValue) {
+	if (passwordValue !== undefined
+		&& confirmValue !== undefined
+		&& passwordValue !== confirmValue) {
 		result = false;
 		detail = 'passwords must match';
 	}
+	if ((passwordValue === undefined || confirmValue === undefined)
```
JedWatson: This isn't necessary - if the `confirmValue` isn't provided, we don't validate it server-side. Password fields may be updated programatically or after client-side verification,

JedWatson:
I worked on `validateRequiredInput` for Password fields recently (last week? the week before? recently.) and it's working as expected

fields/types/password/test/type.js
```diff
@@ -8,6 +8,376 @@ exports.initList = function (List) {
...
+			var testItem = new List.model({
+				password: 'asdf',
+			});
+			List.fields.password.updateItem(testItem, {
+				password: null,
+			}, function () {
+				demand(testItem.password).be.null();
+				done();
+			});
+		});
```
mxstbr: Should this be the case?

fields/types/password/test/type.js
```diff
@@ -8,6 +8,376 @@ exports.initList = function (List) {
...
+			var testItem = new List.model({
+				password: 'asdf',
+			});
+			List.fields.password.updateItem(testItem, {
+				password: '',
+			}, function () {
+				demand(testItem.password).be('');
+				done();
+			});
+		});
```
mxstbr: Should this be the case?

fields/types/password/test/type.js
JedWatson: Yep, that should clear the stored value. We should actually query and assert the stored value after a save to ensure we're not saving a hash of `null` (if that's even a thing)

fields/types/password/test/type.js
JedWatson: Yes, again with an assertion that the stored value is cleared as well.

fields/types/password/PasswordType.js
mxstbr: Alright. Do we check that a `password` was passed? Or if that's undefined but the `confirm` exists should it still pass?

JedWatson: Good catch. If the `password` is undefined but a `confirmValue` was passed, I think we should fail validation as if they were both provided and didn't match.

mxstbr: Alright, what about `password` is `null` or `""` and `confirmValue` is valid?

mxstbr:
Updated, should now hit on all points. I don't yet check the stored value, but we'll do that in another PR

PR #2491, opened 2016-03-23
Allow for a better way of selecting tabs and sub-menu nav items

webteckie:
Adds a `data-list` attribute to the home tabs as well as to the nav sub-menus.  This is to simplify the e2e selector configuration for dealing with these elements.


mxstbr:
LGTM!

PR #2503, opened 2016-03-23
Display categories in filter list

mxstbr:
Previously in the admin interface, we'd show category headers in the "Columns" list but not in the "Filters" list. This PR fixes that:

<div  align="center">
Filter list with category headers
</div>


JedWatson:
Good catch @mxstbr

PR #2507, opened 2016-03-24
Factor defining prototype getters out into its own utils file

mxstbr:
Cleaner code and avoids excessive code repetition


JedWatson:
Looks good to me @mxstbr!

PR #2512, opened 2016-03-24
Fix text filtering

mxstbr:
**Please functionally review.** I'm not 100% this isn't just my browser cache still having an old version but I think this fixes the "Begins with…" filter for the text and markdown fields.


JedWatson:
Good catch, this _is_ inconsistent with the React Text Filter component

JedWatson:
Just FYI I found this in the `name` field as well. Updated.

PR #2518, opened 2016-03-25
Immutable filtering and searching methods

mxstbr:
Refactored filtering and searching methods to make query immutable.

This is much cleaner than what we previously had, where `addFilterToQuery` and `addSearchToQuery` would mutate the query somewhere deep inside the fields.

### DIFF
--- a/admin/server/api/list/download.js
+++ b/admin/server/api/list/download.js
@@ -1,6 +1,7 @@
 var baby = require('babyparse');
 var keystone = require('../../../../');
 var moment = require('moment');
+var assign = require('object-assign');
 
 module.exports = function (req, res) {
 	var format = req.params.format.split('.')[1]; // json or csv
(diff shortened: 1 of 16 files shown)


admin/server/api/list/download.js
```diff
@@ -11,10 +12,10 @@ module.exports = function (req, res) {
 		catch (e) { } // eslint-disable-line no-empty
 	}
 	if (typeof filters === 'object') {
-		req.list.addFiltersToQuery(filters, where);
+		where = assign({}, req.list.addFiltersToQuery(filters));
 	}
 	if (req.query.search) {
-		req.list.addSearchToQuery(req.query.search, where);
+		where = assign({}, req.list.addSearchToQuery(req.query.search));
```
JedWatson: It's a valid use case to provide _both filters and a search_, so these lines should be

```
assign(where, req.list.addSearchToQuery(req.query.search));
```

mxstbr: Oh yeah, good catch, my bad. Will fix!

mxstbr:
Fixed and ready to merge!

JedWatson:
Looks good!

PR #2523, opened 2016-03-26
Updated Current Version

tmvanetten:
updated current version to 0.3.17 to match latest release in changelog.


mxstbr:
Good catch, thanks very much!

Issue #916, opened 2015-01-17
LocalFile Field Type Format not working


prachaquant:
```
image: { type: Types.LocalFile,dest: '/home/veda/ks/public/img', prefix: '/img',
            format: function(item, file){
                return '<img src="/img/'+file.path.replace('/home/ks/admin/public/img','')+'/'+file.filename+'" style="max-width: 200px"\>';
            }
    },
```

![screen shot 2015-01-17 at 11 20 32](https://cloud.githubusercontent.com/assets/7953643/5787665/13fbc34a-9e3b-11e4-83ae-397b66f23834.png)

It show an image on keystone version 0.2.36 but  when I update to keystone 0.2.40 it show file name instead.

morenoh149:
Does the file store correctly? It may still be 'working' but we might have changed the display in the adminUI. Can anyone else chime in?

JedWatson:
I can confirm this is (for now) not working in the new React fields, we may need a new strategy to support it. AFAIK the feature was used entirely to support image previews, so maybe we should just make that an option.

dtuyenle:
Would be nice if you could put a notification on the keystone documentation. Thanks

morenoh149:
@dtuyenle a notification of what?

JedWatson:
I'm bumping this to an 0.3.x release, and will include a note that if you're dependent on this you should wait a little bit before upgrading from 0.2.x

adamscybot:
Just ran into this. +1. Agree, file preview option would suffice, though at the same time, I think a longer-term feature might be able to enable a custom format() function on every field.

tomasztunik:
Would be great to have an option for localfile and localfiles fields to set if it's image and render preview if it is, maybe even add option for videos, sounds in the long run but images would be highest priority I believe.

and regarding format as @adamscybot mentions having option to format every field would be great way to do simple adjustments without having to hack or create custom fields for keystone for many less complex use cases. For example being able to display image preview in place of name in the relationship field could be great usability helper that could be easily handled with that.

JedWatson:
See #1222

I would love to add functionality like this to the common `File` field - let's get that merged and go from there.

prachaquant:
:+1: I like the idea.

prachaquant:
If change to common file field will that change data schema on mongodb? I'm working on project that depend on Localfile field.

mariohmol:
Same with S3File #2215

maxkoryukov:
How it works now????

[...]

PR #2528, opened 2016-03-27
made localfiles previews properly show up in admin ui column

krksgbr:
I came across an issue when I wanted to have  preview images of uploaded localfiles in the admin ui column view. The localfiles format method was not properly called in templates/mixins/columns.jade, so the previews did not properly show up in the admin ui column. The format method expects an index of the file, which was never supplied to the method, so it always defaulted to returning just the number of files. This commit fixes the issue. possibly related to #916.


JedWatson:
Thanks @krksgbr!

PR #2544, opened 2016-03-30
Datearray filtering

mxstbr:
Implemented datearray filter, including the new "presence" option.

### DIFF
--- a/fields/types/datearray/DateArrayType.js
+++ b/fields/types/datearray/DateArrayType.js
@@ -120,7 +120,7 @@ datearray.prototype.addFilterToQuery = function (filter) {
 	var dateTypeAddFilterToQuery = DateType.prototype.addFilterToQuery.bind(this);
 	var query = dateTypeAddFilterToQuery(filter);
 	if (query !== {} && query[this.path]) {
-		query[this.path] = addPresenceToQuery(filter.presence, query[this.path]);
+		query[this.path] = addPresenceToQuery(filter.presence || 'some', query[this.path]);
 	}
 	return query;
 };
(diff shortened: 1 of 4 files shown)


fields/types/datearray/DateArrayType.js
```diff
@@ -105,6 +107,25 @@ datearray.prototype.validateRequiredInput = function (item, data, callback) {
...
+ *
+ * @param {Object} filter 			   		The data from the frontend
+ * @param {String} filter.mode  	   		The filter mode, either one of "on",
+ *                                     		"after", "before" or "between"
+ * @param {String} [filter.presence='some'] The presence mode, either on of
+ *                                          "none" and "some". Default: 'some'
+ * @param {String|Object} filter.value 		The value that is filtered for
+ */
+datearray.prototype.addFilterToQuery = function (filter) {
+	var dateTypeAddFilterToQuery = DateType.prototype.addFilterToQuery.bind(this);
```
mxstbr: I used the `DateType` `addFilterToQuery`, simply adding the presence at the end which works beautifully.

If you go back two or three commits I had the whole thing copy & pasted and added the presence at the end, which didn't feel very DRY. If you want me to revert this I'll gladly do, but I think this is much nicer!

fields/types/datearray/DateArrayType.js
```diff
@@ -105,6 +107,25 @@ datearray.prototype.validateRequiredInput = function (item, data, callback) {
...
+	var query = dateTypeAddFilterToQuery(filter);
+	if (query !== {} && query[this.path]) {
```
JedWatson: This will always be true

PR #2550, opened 2016-03-31
Add geopoint filtering tests

mxstbr:


fields/types/geopoint/test/filters.js
```diff
@@ -0,0 +1,162 @@
...
+				COORDINATES.nyc,
+				COORDINATES.sydneyCenter,
+				COORDINATES.sydneySuburb,
+				COORDINATES.vienna,
+			]);
+			done();
+		});
+	});
+
+	it('should not filter anything if lon is not specified', function (done) {
```
JedWatson: picky, but longitude is abbreviated to `lng`, not `lon`

mxstbr: Lets get this landed, and when I refactor the field filtering to use constants I'll change this too!

PR #2552, opened 2016-03-31
Make boolean field type render "for" attribute

### LINKED ISSUE #2551: boolean field type does not render the "for" attribute
All other field types render a `for` attribute whose value is the field name but the boolean field type doesn't.  For example, compare the text field type below for a `name` field vs the boolean type below for a `fieldA` field:

image

This makes it a bit difficult to use a non-positional selector to select the field type rendering.

cc: @joerter

joerter:
This fixes #2551. I'm not sure if this is the best way to fix as I noticed that the Boolean field uses the base Field.js a little differently than other fields.

### DIFF
--- a/fields/types/boolean/BooleanField.js
+++ b/fields/types/boolean/BooleanField.js
@@ -40,7 +40,7 @@ module.exports = Field.create({
 
 	renderUI () {
 		return (
-			<FormField offsetAbsentLabel={this.props.indent} className="field-type-boolean">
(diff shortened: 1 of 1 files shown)


webteckie:
LGTM

mxstbr:
LGTM too, thanks so much @joerter!

JedWatson:
:+1: Thanks @joerter!

webteckie:
@joerter now that this has landed I will leave it up to you to  make the edits to the boolean test selectors...thanks :-)

PR #2562, opened 2016-04-01
AzureFile uploads seem broken since v0.3.13

andreugrimalt:
A quick fix is to set the AzureFileType options before calling super_.call()


JedWatson:
Thanks @andreugrimalt!

PR #2565, opened 2016-04-02
Add e2e tests for Number field

joerter:
@webteckie here are the tests for the number field. I also took took out the nth-child selectors for the Boolean field in adminUI.


test/e2e/adminUI/adminUI.js
```diff
@@ -151,14 +151,14 @@ module.exports = {
...
 						},
 						fieldB: {
-							button: '.EditForm-container .field-type-boolean:nth-child(6) label button',
-							label: '.EditForm-container .field-type-boolean:nth-child(6) label span',
-							value: '.EditForm-container .field-type-boolean:nth-child(6) label input[name="fieldB"]'
+							button: '.EditForm-container .field-type-boolean[for="fieldB"] button',
+							label: '.EditForm-container .field-type-boolean[for="fieldB"] span',
+							value: '.EditForm-container .field-type-boolean[for="fieldB"] input[name="fieldB"]'
 						}
 					}
```
webteckie: minor comment here...@JedWatson want tests to eventually pass the same linting rules applied to keystone core, one of which is comma dangling.  Please add a trailing comma to to all JSON lines...see others for example...tnx

test/e2e/adminUI/adminUI.js
```diff
@@ -294,9 +294,9 @@ module.exports = {
 							value: '.Modal-dialog .field-type-text[for="name"] input[name="name"]',
 						},
 						fieldA: {
-							button: '.Modal-dialog .field-type-boolean button',
-							label: '.Modal-dialog .field-type-boolean span',
-							value: '.Modal-dialog .field-type-boolean input[name="fieldA"]',
+							button: '.Modal-dialog .field-type-boolean[for="fieldA"] button',
+							label: '.Modal-dialog .field-type-boolean[for="fieldA"] span',
+							value: '.Modal-dialog .field-type-boolean[for="fieldA"] input[name="fieldA"]',
 						}
```
webteckie: ditto

test/e2e/adminUI/group005Fields/uiTest006NumberField.js
```diff
@@ -0,0 +1,41 @@
...
+		browser.expect.element(adminUI.cssSelector.initialModalView.fieldType.number.number.name.label)
+			.text.to.equal('Name');
+		browser.expect.element(adminUI.cssSelector.initialModalView.fieldType.number.number.name.value)
+			.to.be.visible;
+
+		browser.expect.element(adminUI.cssSelector.initialModalView.fieldType.number.number.fieldA.label)
+			.to.be.visible;
+		browser.expect.element(adminUI.cssSelector.initialModalView.fieldType.number.number.fieldA.label)
+			.text.to.equal('Field A');
+	}
```
webteckie: can we also check here for the existence of the value field?

webteckie:
@joerter love it!  Just some minor comments :-)  tnx

joerter:
@webteckie awesome, I think I found all the instances of dangling commas let me know if there's any I missed.

test/e2e/adminUI/adminUI.js
```diff
@@ -251,8 +251,8 @@ module.exports = {
 						fieldB: {
 							label: '.EditForm-container .field-type-number[for="fieldB"] label',
 							value: '.EditForm-container .field-type-number[for="fieldB"] input[name="fieldB"]'
```
webteckie: add one here

test/e2e/adminUI/adminUI.js
```diff
@@ -164,8 +164,8 @@ module.exports = {
 							button: '.EditForm-container .field-type-boolean[for="fieldB"] button',
 							label: '.EditForm-container .field-type-boolean[for="fieldB"] span',
 							value: '.EditForm-container .field-type-boolean[for="fieldB"] input[name="fieldB"]'
```
webteckie: add one here

webteckie:
Missed a couple but one of them was my bad :-)  We should probably run lint on this at some point to cover all the basis...but for now eyeballing it is fine :-)  Also I had added a couple more comments in the original review...see above.

joerter:
@webteckie No problem, got em!

webteckie:
How about the comment to check for the presence of the number field A value in the initial modal?  Does it make sense to do so in the ui test?

test/e2e/adminUI/group005Fields/uiTest006NumberField.js
```diff
@@ -0,0 +1,43 @@
...
+		browser.expect.element(adminUI.cssSelector.initialModalView.fieldType.number.number.name.label)
+			.text.to.equal('Name');
+		browser.expect.element(adminUI.cssSelector.initialModalView.fieldType.number.number.name.value)
+			.to.be.visible;
+
+		browser.expect.element(adminUI.cssSelector.initialModalView.fieldType.number.number.fieldA.label)
+			.to.be.visible;
+		browser.expect.element(adminUI.cssSelector.initialModalView.fieldType.number.number.fieldA.label)
+			.text.to.equal('Field A');
+		browser.expect.element(adminUI.cssSelector.initialModalView.fieldType.number.number.fieldA.value)
```
joerter: Do you mean this check here?

test/e2e/adminUI/group005Fields/uiTest006NumberField.js
```diff
@@ -37,5 +37,7 @@ module.exports = {
 			.to.be.visible;
 		browser.expect.element(adminUI.cssSelector.initialModalView.fieldType.number.number.fieldA.label)
 			.text.to.equal('Field A');
-	}
+		browser.expect.element(adminUI.cssSelector.initialModalView.fieldType.number.number.fieldA.value)
```
joerter: Do you mean this check here?

joerter: Do you mean this check here?

webteckie:
ah, cool, did realize you had done it...merged!  which one you want to work on next?  :-)  tnx

joerter:
No problem! Let's see... how about text? Any unique cases I should test for with that?

webteckie:
No that I can think of...I think it just a string.  It's yours :-)  btw, the first pass we are just doing basic tests.  Another pass will test other field options as well as more in depth coverage of filtering, etc.  btw, I've also been thinking of restructuring adminUI.js since it's becoming a large module.  Somehow try to isolate field selector config into its own per field file.  Mulling it over...if you have any suggestions let's hear them :-)

joerter:
Yeah I was looking at the nightwatch documentation today and was wondering what you think of page objects? If we had separate page objects for the different views we have now under `cssSelector` it might make things easier because we could add in commands to do things like log in / out, click on plus icons for fields, etc.

webteckie:
Yeah, let's refactor using that pattern before we get deeper.  You want to take a stab at it?  If not, let me know and I will.   Thanks!

webteckie:
Actually we can both attack this in parallel.   We just got to leave adminUI.js alone for now until we're done creating all the page objects and refactoring the tests to use those page objects instead of adminUI.js.  But let's just do the signin tests first so that we set the pattern and file structure down.

joerter:
Sounds good! Going in parallel that way makes a lot of sense. I'll take a look tomorrow.

webteckie:
@joerter when you get a chance head over to the slack design channel :-)

PR #2608, opened 2016-04-11
Begin migration to webpack

cameronroe:
I think it might make sense to start transitioning over to webpack as you begin to introduce more React/Redux code and remove gulp.


package.json
```diff
@@ -140,8 +152,8 @@
     "url": "https://github.com/keystonejs/keystone/issues"
   },
   "scripts": {
-    "build": "NODE_ENV=production gulp build-packages",
-    "build-dev": "gulp build-packages",
+    "build": "NODE_ENV=production webpack",
```
wmertens: these additionally need --bail, otherwise webpack won't return error on build failures

webpack.config.babel.js
```diff
@@ -0,0 +1,75 @@
...
+
+function createProdConfig(config) {
+	let plugins = [
+		new webpack.optimize.UglifyJsPlugin({
+		    compress: {
+		        warnings: false
+		    },
+		    mangle: true,
+		    output: {
+		    	comments: false
```
wmertens: Here's two more options I like:

    screw_ie8: true,
    semicolons: false

Without semicolons you get the same filesize but editors can handle the file much better because the lines are short. And of course, IE8 is not supported by React either.

webpack.config.babel.js
```diff
@@ -0,0 +1,75 @@
...
+		path: path.resolve(__dirname, './admin/public/js')
+	},
+	module: {
+		loaders: [
+			{ test: /\.jsx?$/, loader: 'babel', exclude: /node_modules/ }
+		]
+	},
+	plugins: [
+		new webpack.DefinePlugin({
+			ENV: process.env.NODE_ENV
```
wmertens: Is this correct? I was under the impression it had to be `{process:{env:{NODE_ENV: JSON.stringify(process.env.NODE_ENV)}}}`

wmertens:
@cameronroe could you also add a config for the admin UI with all fields included? Then we can optimize that with react-proxy loader so only the used fields are downloaded by the client and we'll have a static admin ui build!

cameronroe:
I've added the admin fields and filters.

Regarding `react-proxy`, I think it might be better to start with using the require prefix for the loader instead of making it global. Could be good to just get webpack in first then we can refine the components for better hot-loading etc. testing with prefixes first, then going to a global loader. 

Thoughts @wmertens ?

wmertens:
I'm confused, shouldn't `admin` and `fields` be in the entries for them to be built?

The end goal is to have pre-built packages for everything, with async loading of fields so the ui is not 6MB of code.

That way, the browserify dependency is completely removed, the keystone startup super fast and the admin UI perfectly sized.

The admin UI is currently created by browserify and served here https://github.com/keystonejs/keystone/blob/master/admin/server/app/createStaticRouter.js#L72-L74, we should replace that too.

cameronroe:
Oops, sorry about that.

cameronroe:
Any other thoughts on this one @wmertens ?

wmertens:
Well, I gave it a shot and this branch doesn't work at all, I can't compile the client and fields so I commented those out, and when I renamed the built file to `packages.js`, the client doesn't start because it can't find `react`. Much more tweaking needed methinks.

wmertens:
I used this branch as a basis for https://github.com/keystonejs/keystone/tree/webpackify - I was able to run `node test/e2e/server.js --notest` with it :) (after building of course)

Now to make it load the fields lazily…

wmertens:
closing this in favor of #3360, comments welcome there

PR #2612, opened 2016-04-12
Added support for heroku's mLab connection string

MichaelZaporozhets:
Updated to reflect the current version


JedWatson:
Thanks @MichaelZaporozhets!

PR #2613, opened 2016-04-12
SSL: Support SNI and PFX

wmertens:
* Support SNI: Provide a synchronous fn on 'ssl sni' that, given a hostname, returns a
  suitable object to give to tls.createSecureContext()
* Support 'ssl pfx' files and 'ssl passphrase'. PFX can be encrypted, which provides some security at rest.
* Allow providing the certificate data directly via 'ssl .* data'
* If no default certificates are provided, use the sni function to get them

@JedWatson @mxstbr thoughts on this? I think this makes it easy to have multiple servers on the same host…

I still need to test with memoization and will squash update.

### DIFF
--- a/server/startSecureServer.js
+++ b/server/startSecureServer.js
@@ -10,6 +10,7 @@
  */
 
 var https = require('https');
+var tls = require('tls');
 var fs = require('fs');
 
 module.exports = function (keystone, app, created, callback) {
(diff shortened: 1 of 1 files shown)


wmertens:
Decided that memoizing should be done by the user if necessary - NodeJS runs the SNICallback a few times but not any more after that. I didn't want to second-guess NodeJS.

server/startSecureServer.js
```diff
@@ -29,27 +31,65 @@ module.exports = function (keystone, app, created, callback) {
...
+			var localCtx = sniFunc('localhost');
+			if (localCtx) {
+				for (var prop in localCtx) {
+					if (localCtx.hasOwnProperty(prop)) {
+						options[prop] = localCtx[prop];
+					}
+				}
+			}
+		}
+		if ((!options.key || !options.cert) && !options.pfx) {
```
webteckie: Isn't the this check (L72) the same as the check at L60?

wmertens: Yes but the options might have changed thanks to the call from sniFunc…

This is an error case, so I think code clarity is more important than not repeating the test if there is no sniFunc.

wmertens:
@JedWatson ?

JedWatson:
Sorry for the delay @wmertens - looks good to me! Thanks.

PR #2566, opened 2016-04-02
[WIP] Admin Interface Refactor

mxstbr:
WIP DO NOT MERGE

<br />

**TODO**

- [x] Switch to the API instead of form submissions (avoid page reloads)
  - [x] Figure out which parts do not yet use the API
    - [x] Item updates via API
    - [x] Item creation via API
      - [x] Fix file and image uploading on save
        - [x] Create legacy API points for creating and updating
        - [x] Make creating and updating use the legacy endpoints
    - [x] Autocreates via API
- [x] Clean up the react part
  - [x] Architecture update
  - [x] Integrate react-router
    - [x] Fix download choosing columns
  - [x] Integrate Redux
    - [x] ListView
      - [x] Remove drag-n-drop-across-pages for now, reimplement later on
      - [x] Get rid of CurrentListStore, instead use Redux actions
        - [x] Filter selection via actions
        - [x] Column selection via actions
        - [x] Search via actions
        - [x] Pagination via actions
        - [x] Row Alerts via Redux actions
      - [x] Sync query parameters with state (react-router-redux?)
      - [x] Fix filtering
      - [x] Fix endless loop when switching between lists
      - [x] Fix empty to non-empty list switch not loading any items
      - [x] Fix redirection after item creation
      - [x] Fix linking to query parameters
      - [x] Deleting an item
    - [x] ItemView
      - [x] Fix redirection after item creation from item (`Item.js:53`)
      - [x] Resetting the form
      - [x] Deleting an item
    - [x] HomeView
- [x] **Fix flash messages**
  - [x] Home
    - [x] General network errors
  - [x] Item
    - [x] General network errors
    - [x] Validation errors
    - [x] Item creation duplicate values database error message
  - [x] List
    - [x] General network errors
  - [x] Form errors
    - [x] CreateForm
    - [x] EditForm
    - [x] Signin form
- [x] Remove custom stores
  - [x] SessionStore
    - [x] Sign in via API on Signin view
    - [x] Delete SessionStore file
  - [x] CurrentListStore
    - [x] Integrate Redux
    - [x] Delete CurrentListStore file
  - [x] Get rid of `store-prototype` package

----

Legend:
- *italic*: Do after this is merged
- **bold**: Important to do before merging

PR #2616, opened 2016-04-13
allow prefix + to specify ascending sorting of a column (fixes #2022)

### LINKED ISSUE #2022: 0.4: defaultSort does not seem to work
Not sure if it's just my environment (latest chrome, windows 10) but adding `defaultSort: "name"` to a list does not seem to enable sorting on that list field.

I've added keystonejs/keystonejs-site#54 just in case it's a documentation issue.

Warning messages in console:
image

Used like this (according to doc example):
```
var MyList = new keystone.List('MyList', {
    defaultSort: '+name'
});
MyList.add({
    name: {
        type: String,
    },
    ...
});
```

geloescht:
Conflicts with #2566 , but it's a pretty simple patch.

### DIFF
--- a/admin/client/lib/List.js
+++ b/admin/client/lib/List.js
@@ -118,6 +118,9 @@ List.prototype.expandSort = function (input) {
 			invert = true;
 			path = path.substr(1);
 		}
+		if (path.charAt(0) === '+') {
+			path = path.substr(1);
+		}
 		const field = this.fields[path];
 		if (!field) {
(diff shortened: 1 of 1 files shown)


mxstbr:
This does not yet conflict with #2566, I haven't changed lib/List.js. Would you mind submitting this PR against the single-page-adminui branch, then I'll merge it into that WIP PR! Thanks so much!

PR #2641, opened 2016-04-16
Fixes for LocalFilesField

micahbule:
- Added a temporary ID field for new file uploads, as current code assumes there is an exising _id field
- Refactored removeItem() so as to follow React's immutable principle
- Removed 'key' field from propTypes as it's causing a warning

NOTE: I'm not sure if how I implemented a temporary ID field is correct. If you have any other suggestions, I can try to implement it.

### DIFF
--- a/fields/types/localfiles/LocalFilesField.js
+++ b/fields/types/localfiles/LocalFilesField.js
@@ -17,7 +17,6 @@ var LocalFilesFieldItem = React.createClass({
 		deleted: React.PropTypes.bool,
 		filename: React.PropTypes.string,
 		isQueued: React.PropTypes.bool,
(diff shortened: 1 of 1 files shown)


mxstbr:
Thanks so much for taking the time to submit this PR, we really appreciate it!

We've been thinking about revamping all storage fields, so I'm not 100% sure we want to merge this right now. Let's wait for @JedWatson to chime in!

JedWatson:
I'm happy for this to go in as is - thanks @micahbule, much appreciated!

re: the incremented ID, I usually prefix a string to numeric keys (numeric keys used to cause issues, that may not be true anymore) but otherwise this is the same technique I've used myself in a few places :)

PR #2645, opened 2016-04-18
[0.3.x] update mongo-connect using compatible mode for node 0.12. According…

gemscng:
… to github doc

https://github.com/kcbanner/connect-mongo

### DIFF
--- a/lib/core/mount.js
+++ b/lib/core/mount.js
@@ -129,13 +129,13 @@ function mount(mountPath, parentApp, events) {
 		// Perform any session store specific configuration or exit on an unsupported session store
 
 		if (sessionStore === 'mongo') {
-			sessionStore = 'connect-mongo';
+			sessionStore = 'connect-mongo/es5';
 		} else if (sessionStore === 'redis') {
 			sessionStore = 'connect-redis';
 		}
 
 		switch (sessionStore) {
-			case 'connect-mongo':
+			case 'connect-mongo/es5':
 				debug('using mongo for session store');
 				_.defaults(sessionStoreOptions, {
 					collection: 'app_sessions',


mxstbr:
We're not going to release another 0.3.x version, so I'm inclined to close this Pull Request as it won't reach any endusers. @JedWatson what do you think?

creynders:
@mxstbr until v0.4 _is_ released I definitely wouldn't assume no more v0.3.x releases will happen. People will need some time to transition and in the mean time some releases _may_ happen.

JedWatson:
This is a kind-of challenging one (I mean, I'm not sure what to do)

I think I made a mistake with the session store option; you should pass it an instance of the package you want to use (or something similar) rather than an option that requires a package to be installed. It's really weird as it is, and we have no control over the version.

I'm not sure that _always_ requiring the es5 version is a good idea but we need to expose this in some way. At the moment, if users install the latest version there's no way for it to work with keystone and node 0.12, which is bad and totally worth a new 0.3 patch release even if just for this fix.

Unfortunately fixing it correctly would require a breaking change; which is the challenging thing here. We can't do that, for obvious reasons.

Maybe the best thing to do here is just merge the PR and cut a new version of 0.3

@creynders your thoughts?

gemscng:
@JedWatson A better version but not the correct version can be checking the node version before require different package.   Since 0.3.x is only for supporting current user and not really going to iterate it in the best way, i think temporary fix  is fine. We should not fix this way in 0.4, so i haven't submit pull request in master branch.

creynders:
@Gemscng is right, let's check the node version and require the correct version accordingly.

JedWatson:
@mxstbr let's get that into the new 0.3 version you're about to release

JedWatson:
**edit** I've realised there are some issues with this implementation, will add it myself to the `0.3.x` branch and `master`. Thanks for the PR, @gemscng :)

Issue #2500, opened 2016-03-23
Add browserify CLI npm script and remove gulp


JedWatson:
We really don't need gulp just to browserify + uglify our common packages for the Admin UI.

If someone can take our last remaining gulp task `build-packages` and rewrite it as an npm script that would be great. There are two tasks: build for production (with `NODE_ENV=production`, strips warnings etc. from React, and minifies the build) and build for dev (with source maps, debug flags, etc) that need to be replicated.

cameronjroe:
Thoughts on using webpack for the build? It has a number of benefits in regards to bundling/optimization.

JedWatson:
@cameronjroe I'm definitely open to that but we'd need to replace the custom browserify middleware we have too, which is probably worth another issue to discuss. It currently handles some things I'm not sure how to do effectively with Webpack (possibly due to inexperience) including the ability to build from generated JS code at runtime (which we'll need for supporting custom field types).

Seems simpler to use the same system for the prebuilt packages and the dynamic admin bundle so until then we'd stick with Browserify (or not? reasons?)

Ultimately a webpack setup would be great, hot reloading for dev alone would be worth it.

cameronjroe:
Cool, yeah webpack has has a number of [loaders](https://webpack.github.io/docs/list-of-loaders.html) that help for bundling. There's also [webpack-dev-middleware](https://github.com/webpack/webpack-dev-middleware) that could be used in node for dev purposes. If it's just client-side code being bundled though, you probably would want to serve that up statically no? What I do see out there is plenty of isomorphic [webpack configurations](https://github.com/iam4x/isomorphic-flux-boilerplate/blob/master/webpack/prod.config.js) that help with the static builds. It's not too bad once you see how the configuration works.

JedWatson:
It's just client-side code being bundled at the moment, I don't see isomorphic/universal rendering being a great advantage for the Admin UI. At least, it's not a near-term target.

I've used webpack for other projects but to summarise the current requirements for Keystone that I haven't worked out with it yet:

- Common bundles (React, Elemental, etc) get pre bundled (and minified, etc) to keep launch time fast - this file gets committed into the repo and rebuilt when npm packages are updated, which is kind of unfortunate but ultimately better for users
- The rest of the Admin UI JS needs to be bundled at project runtime from a dynamic template so only Field types actually used by Lists, and including custom field types from npm, and plugins down the track, are included
[...]

PR #2669, opened 2016-04-22
Add browserify CLI npm script and remove gulp

suryagh:
Removed `gulp-*` #2500 

New intermediate file `packages.bundle.js` is used for following two reasons:

1.  `uglify` does not take the input file as a `stream` (so could not pipe), instead it has to be passed in as a `file name`
2.  The bundle file can  be referenced by the `.map` file.


mxstbr:
This looks good to me!

JedWatson:
Awesome! Thanks @suryagh, really appreciate it!

PR #2676, opened 2016-04-23
Comment for signin?signedout version added

divyabiyani:
As suggested and explained by @mxstbr ,comment explaining the working for Line 11 of  https://github.com/keystonejs/keystone/blob/master/admin/server/routes/signout.js is added.


mxstbr:
One tiny nitpick, can we move that above the ´if (...)`? Otherwise it's perfect, thanks so much!

mxstbr:
Thanks so much!

divyabiyani:
@mxstbr Thank you so much for explaining it.

PR #2686, opened 2016-04-26
[e2e testing] Add ui and ux tests for text fields

jstockwin:
Add testing for text fields, as set out in #2291 

@webteckie - Was attracted by the 'easy for beginners' tag on the issue, so am new to all this. Thought it would be interesting to see how the testing worked. I think I've added the testing for the Text field successfully. Happy to try a few others if this is correct, however please check I haven't missed anything important! 

All e2e tests (including the new ones) pass locally.

### DIFF
--- a/test/e2e/adminUI/adminUI.js
+++ b/test/e2e/adminUI/adminUI.js
@@ -70,11 +70,6 @@ module.exports = {
 			labelForSelectsTabUnderDashboardFieldsSubheading: '.dashboard-group__list[data-list-path="selects"] .dashboard-group__list-label',
 			plusIconLinkForSelectsTabUnderDashboardFieldsSubheading: '.dashboard-group__list[data-list-path="selects"] a.dashboard-group__list-create.octicon.octicon-plus',
 			itemCountForSelectsTabUnderDashboardFieldsSubheading: '.dashboard-group__list[data-list-path="selects"] .dashboard-group__list-count',
-			// Text List Tab
-			textsTabUnderDashboardFieldsSubheading: '.dashboard-group__list[data-list-path="texts"]',
-			labelForTextsTabUnderDashboardFieldsSubheading: '.dashboard-group__list[data-list-path="texts"] .dashboard-group__list-label',
-			plusIconLinkForTextsTabUnderDashboardFieldsSubheading: '.dashboard-group__list[data-list-path="texts"] a.dashboard-group__list-create.octicon.octicon-plus',
-			itemCountForTextsTabUnderDashboardFieldsSubheading: '.dashboard-group__list[data-list-path="texts"] .dashboard-group__list-count',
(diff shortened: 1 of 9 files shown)


morenoh149:
'easy for beginners' +1

webteckie:
@jstockwin great to have you interested in e2e! `easy for beginners` is a relative term...e2e can become a bit of a pain at times!  However, because there are patterns established, it should be relatively easy to apply those patterns.

Anyways, a few weeks ago your changes would have been accepted :-)  However, currently we are undergoing a refactoring effort to use the concept of Nightwatch [page objects](http://nightwatchjs.org/guide#page-objects).  I've already implemented the `name` field so you can use that implementation as guide for implementing new fields.  The following are the main pieces:

- Need to define the test lists (you already did that and looks good)

- Need to add the list to the test server (you already did that and looks good)

- Need to define the page object for the field (see https://github.com/keystonejs/keystone/blob/master/test/e2e/adminUI/pages/fieldTypes/name.js for the one for the `name` field).  I actually already had done it for text since needed it for simple string fields, which all lists that we test with have a string/text `name` field (see https://github.com/keystonejs/keystone/blob/master/test/e2e/adminUI/pages/fieldTypes/text.js)

- Need to define a list for testing the field; however, unlike before this is now done via a Page Object (see https://github.com/keystonejs/keystone/blob/master/test/e2e/adminUI/pages/lists/name.js for the one for the `name` field)

- Need to implement the actual tests (for the `name` field example, see https://github.com/keystonejs/keystone/tree/master/test/e2e/adminUI/tests/group005Fields/name).  But you already got most of it.  Just refactor a little bit to use the page objects and you should be good!  NOTE:  You should not have to use adminUI.js anymore in your tests!

- There's some extra testing that need to be done for fields off the home screen testing...@joerter is refactoring that right now and when it's done then a pattern for doing that should also be established.

Hope this is clear :-)

Thanks

PS:  if anyone else wants to join in adding e2e testing for fields, you can follow the above steps!

jstockwin:
@webteckie thanks for the detailed reply, much appreciated! Please see latest commit. I have

- Undone my changes to adminUI.js

- Added a `textsFieldsSubment` entry in pages/app.js

- Added `textList` to both initialForm.js and item.js.

- Added a text list in pages/lists

- Refactored the test code to use the page objects.

I have not done anything regarding your last point about testing for fields off the home screen, but other than adding this when the pattern is established, I think that should be everything? All checks pass locally.

webteckie:
@jstockwin this is awesome!  I think you got the pattern down!  Rock on :-)  Please let me know what field you want to work on so that others working on it do not collide on the same field.  Thanks!

webteckie:
And, btw, if you see ways to improve while we are still early on field testing let us know!

PR #2688, opened 2016-04-26
Update PasswordType.js to allow for inherited users

autoboxer:
Fix an issue where a user model that inherits from another model will have the bcrypt hash function run twice, causing an incorrect password hash to be saved.

### DIFF
--- a/fields/types/password/PasswordType.js
+++ b/fields/types/password/PasswordType.js
@@ -60,20 +60,28 @@ password.prototype.addToSchema = function () {
 			return next();
 		}
 		var item = this;
-		bcrypt.genSalt(field.workFactor, function (err, salt) {
-			if (err) {
-				return next(err);
-			}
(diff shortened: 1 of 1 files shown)


JedWatson:
This is a good solution, thanks for figuring out what's going on @autoboxer 

We need to take the field path into account or multiple password fields in a model would conflict with each other (edge case, sure, but still unexpected and would cause major issues if it's triggered)

I'm happy to merge this and make that change, it's probably easier to code than explain. Will tag this issue when I commit so you can see how it worked out.

mxstbr:
Jed, this wasn't meant to be merged which is why it's tagged "in progress". Can you revert please? We were going to pair-code on tests.

JedWatson:
Sorry @mxstbr! Didn't notice you'd assigned yourself to it. Will do.

PR #2694, opened 2016-04-27
Fixed rendering of DateArrayField

igorrafael:
Lists with DateArrays were not rendering because of a simple error. I do not know how to provide the tests that would fail in order to conform with the contributing guidelines.

Since this is a very simple fix and it is clear that this.moment does not exist I hope that there's no need for the usual contribution bureaucracy.

If it is really needed I would appreciate some pointers on how to create such tests.

### DIFF
--- a/fields/types/datearray/DateArrayField.js
+++ b/fields/types/datearray/DateArrayField.js
@@ -31,7 +31,7 @@ module.exports = Field.create({
 	},
 
 	formatValue (value) {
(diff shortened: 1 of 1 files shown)


mxstbr:
Looks good to me, thanks very much for catching and fixing that!

JedWatson:
Same, thanks @igorrafael!

PR #2701, opened 2016-04-28
Error Code for Bcrypt Error changed

divyabiyani:
It is a fix for https://github.com/keystonejs/keystone/blob/master/admin/server/api/session/signin.js.

The error status of bcrypt error and Database error was same because of which the offline page that we are planning to display when database error will occur will be displayed even when password error occurs.

### DIFF
--- a/admin/server/api/session/signin.js
+++ b/admin/server/api/session/signin.js
@@ -21,7 +21,7 @@ function signin (req, res) {
 							});
 						});
 					} else if (err) {
-						return res.status(500).json({ error: 'bcrypt error', detail: err });
+						return res.status(400).json({ error: 'bcrypt error', detail: err });
 					} else {
 						return res.json({ error: 'invalid details' });
 					}


JedWatson:
@divyabiyani that code branch should be _extremely rare_, it only happens when for some reason the `bcrypt` module returns an error (which is semantically similar to the database driver returning an error).

So it's not a password error so much as a server error, in which case `500` is actually the correct status code.

Status 400 means "The server cannot or will not process the request due to an apparent client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing)" (see https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) which is not the case in this condition.

If you want to differentiate between the two states, I suggest doing that based on the JSON returned, and not the status code.

Does that make sense?

divyabiyani:
@JedWatson thanks for telling me this.
Okay,I will display the offline page based on the JSON returned.
Also,I wanted to ask the theme for the offline page and also the content to be displayed in the offline page

JedWatson:
@divyabiyani not sure what you mean by

> I wanted to ask the theme for the offline page and also the content to be displayed in the offline page

?

Going to close this since it seems to be resolved, thanks :)

PR #2706, opened 2016-04-28
test(e2e): Add testing for color field

jstockwin:
cc @webteckie 
For #2291 

I couldn't find any identifying marks on the swatch that would allow me to actually click on a color properly. They don't have a class name or id or anything, and just have their styles set. I could potentially select them based on their style, but that seemed a bit silly, especially given how long the style string actually is. 

I check the color picker etc is visible, but when it comes to actually testing the fields, hex codes are just added into the input as normal. Completely understand if this is not sufficient, but may need some pointers if not...

All tests pass locally.

### DIFF
--- a/test/e2e/adminUI/tests/group005Fields/color/uxTestColorField.js
+++ b/test/e2e/adminUI/tests/group005Fields/color/uxTestColorField.js
@@ -1,0 +1,83 @@
+module.exports = {
+	before: function (browser) {
+		browser.app = browser.page.app();
+		browser.signinPage = browser.page.signin();
+		browser.listPage = browser.page.list();
+		browser.itemPage = browser.page.item();
+		browser.initialFormPage = browser.page.initialForm();
+
+		browser.app.navigate();
+		browser.app.waitForElementVisible('@signinScreen');
+
+		browser.signinPage.signin();
+		browser.app.waitForElementVisible('@homeScreen');
+	},
+	after: function (browser) {
+		browser.app.signout();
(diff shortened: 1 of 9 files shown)


test/e2e/adminUI/tests/group005Fields/color/uxTestColorField.js
```diff
@@ -0,0 +1,83 @@
...
+			.waitForElementVisible('@listScreen');
+
+		browser.listPage
+			.click('@createFirstItemButton');
+
+		browser.app
+			.waitForElementVisible('@initialFormScreen');
+
+		browser.initialFormPage.section.form.section.colorList.section.name
+			.fillInput({value: 'Name Field Test 1'});
```
webteckie: Now that I look at these inputs more...we should either call it something like "Color Field Test 1" or simply "Test 1"...most likely all field tests have been doing the same :-)

webteckie:
@jstockwin I think for a basic test this is fine.  We can come back and revisit later with more sophistication :-)

webteckie:
@jstockwin one thing I noticed with the color picker is that you can also give it inputs like `green`, `blue`, etc.  Perhaps at some point enhance test to input/verify those.  

At some point we should be adding data accessors to the field page objects to enable extra cross-list field data checking and some other assertions in the tests.  For example, for color:
```
getColorValue: function() {
	this
		.getValue('@value', function (result) {
		return result.value;
	});
}
```

PR #2709, opened 2016-04-28
testing(e2e): Added date field type for testing

jstockwin:
cc @webteckie 

In line with work on #2291 

Similar to the color field, I check that the "Today" button is there, but don't actually use this to populate the field. I thought about making the tests click it, and then comparing the value of the field with the system date or something like that, but this seemed a little over the top. Therefore, fields are populated by inserting a string of text in the correct form, and not using the interactive tools that (should) be available. 

I feel we should also check that the popup appears correctly. When looking at the test server site in Chrome, I could see at the bottom a div with the popup, which has classes like `DayPicker`, and `DayPicker-Month`. However, when I included a check for these, the tests failed. I've just looked at the site in Firefox, and I can't find the corresponding divs. Further, if I click on the arrow keys in the popup (to change date) nothing seems to happen. 

I've got a site live that's written in Keystone, so I had a quick look at that in firefox, and it works there, but differently to how it was working in Chrome. There are divs called `pika` rather than `DayPicker`. Presumably, this is due to using the browsers in-build capability? 

Anyway, I feel like I am going on a bit. In short, there are some basic checks implemented here, but none of it uses the popup day picker, or even checks that it is there. The day picker seems browser dependent, so is perhaps hard to test anyway, especially as we'd have to somehow get the real date to we know which month should be popping up. As far as I'm concerned, the test server's date picker doesn't actually work on Firefox, as I cannot change the month with the arrow keys (or any other way) so am restricted to the current month. However, it works fine in a live site of mine, so I'm a bit reluctant to open an issue as I think I might be missing something...


webteckie:
Please resolve conflicts. I will merge after you've done so as it looks good to me.  Also please open an issue as it is a legit bug that it doesn't work properly in firefox!

jstockwin:
That should be conflicts resolved.

PR #2710, opened 2016-04-28
Cloudinary media

WelltokCaleb:
Add new CloudinaryMedia type that supports both images and videos. Adds render of video thumbnail and preview.
TODO: update Lightbox to display videos for a better experience....some day.


waqas19921:
Looks good.
is this is added in v.0.3 also?

Issue #2711, opened 2016-04-29
Test environment server's date field doesn't work. 


jstockwin:
Something is slightly off with the test server's date popup field on Firefox. It works fine on Chrome, but in FF I am unable to click the arrow keys to change the month, resulting in me only being able to choose dates from the current month. 

I've got a live site built on the npm version of keystone-generator, and it's not an issue here. 

To reproduce, run `node test/e2e/server.js --notest` in your keystone root directory. You can sign in with username 'test@test.e2e' and password 'test'. Once PR #2709 is merged, there will already be a model, "Dates", which will contain a date field for you to test on.

jstockwin:
@mxstbr @yn5 Should this issue now be closed?

mxstbr:
Yeah, thanks I missed it! :+1:

PR #2713, opened 2016-04-29
Fix date field month selection on Firefox

yn5:
Fix for https://github.com/keystonejs/keystone/issues/2711.

Problem discussed here: https://github.com/facebook/react/issues/2011.


mxstbr:
LGTM, very good catch! Thanks so much!

JedWatson:
Great, thanks @yn5!

Issue #2664, opened 2016-04-21
Cannot delete last item from a TextArray


AndreasSvendby:
I'm using Keystone v0.4.0 (from master branch) and I've added a field for tags as TextArray. If I add tags, I can delete all but the last tag. After removing it and saving, it shows up again after the refresh.

Vincent-Loi:
I'm having trouble with this too and any relationship fields. Any solutions?

suryagh:
I will take a look into this

suryagh:
@AndreasSvendby @AndreasSvendby  Is this the issue in question?

![issue](https://cloud.githubusercontent.com/assets/5177920/14765200/260803e2-09a5-11e6-8786-a997fad29875.gif)

Vincent-Loi:
@suryagh Yes, that's exactly our issue

AndreasSvendby:
@suryagh Yes, except when I try other fields (as the Author-field here) are saved correctly. Only the tag fields that are not.

suryagh:
@Vincent-Loi @AndreasSvendby what are the versions of keystone you are running?

Vincent-Loi:
@suryagh I'm using v0.4. If it helps, I found out that it worked sometime before 4th April, not sure which commit though

AndreasSvendby:
@suryagh I'm also using v0.4.0 (updated from master branch a couple of days ago)

mxstbr:
This is a duplicate of #1655, so I'll close it for now! Thanks for reporting this, we'll look into it!

Issue #1655, opened 2015-09-08
Text array field bug(s)


jossmac:
Possibly related:
- cannot remove more than one item per save
- cannot remove the last item

dtiemann83:
Any progress on this? Able to duplicate on v0.3.15

jimmygeorgethomas:
Cannot remove the last item in v0.3.18

mxstbr:
Will be fixed in 0.4: #2726!

PR #2721, opened 2016-04-30
modified relationship-type's update-item method behavior

suryagh:
#2664 

1.  `updateItem()` method will now `delete` the `field` if the `path` not present in `data` object. This should fix the issue where -  the last item cannot be removed - because the `data` object will not contain the `empty` item.

2.  Fixed the breaking unit test.

**Note: please review this carefully before merging.**

### DIFF
--- a/fields/types/relationship/RelationshipType.js
+++ b/fields/types/relationship/RelationshipType.js
@@ -213,11 +213,9 @@ relationship.prototype.inputIsValid = function (data, required, item) {
  * Updates the value for this field in the item from a data object.
  * Only updates the value if it has changed.
  * Treats an empty string as a null value.
+ * If data object does not contain the path field, then delete the field. 
  */
 relationship.prototype.updateItem = function (item, data, callback) {
-	if (!(this.path in data)) {
-		return process.nextTick(callback);
-	}
 	if (item.populated(this.path)) {
 		throw new Error('fieldTypes.relationship.updateItem() Error - You cannot update populated relationships.');
 	}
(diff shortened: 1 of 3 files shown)


mxstbr:
Hey Surya, this looks awesome! #1655 and #2664 are related to the TextArray specifically, so I wouldn't close them by merging this PR.

One question, can we add unit tests for the new behaviour in `fields/types/relationship/tests/type.js`?

suryagh:
@mxstbr I see your point. I have unlinked this `pr` from `1665`. And will also add test cases as you suggested.

Btw - the title of #2664 looks misleading. This `pr` should address the issue referenced by #2664 , I had uploaded a video to #2664 , just to make sure I was on the same page with @AndreasSvendby who reported #2664 . Please let me know what you think.

mxstbr:
Ah yes, you're right, I should've read the whole issue! :+1: 

How about the unit tests?

suryagh:
@mxstbr updated the tests.

mxstbr:
Awesome, thanks! This LGTM!

JedWatson:
This and #2726 are breaking changes...

Previously, **only** the `Boolean` field type would have its value updated when `undefined` is passed, which is a concession to HTML forms not submitting anything when a checkbox isn't checked. Other field types require an explicit `null` value to be provided in order to clear out the value, and passing `undefined` skips the update entirely.

Basically, when exposed as a JSON API (or used internally) this could lead to some really confusing behaviour; you'd usually expect `undefined` to mean "don't change this field", and `null` to mean "clear this field". It means we can never really safely autodetect which fields to update in an item by their presence, and worse we have inconsistent behaviour between different field types.

The alternative way to resolve this is to leave the back-end processing logic as is, and change the way values are submitted from the UI.

I'm happy to merge this for now, because fundamentally there is a bug that needs to be fixed preventing data from being saved, but the whole spec / behaviour here needs some more discussion.

PR #2729, opened 2016-05-01
test(e2e): Refactor home screen tests

joerter:
@webteckie: With this refactor, new tabs will just need to be added to the array in home.js in order to be tested on the home screen. Let me know what you think!


mxstbr:
There's some conflicts here, would you mind pulling in the newest master?

test/e2e/adminUI/pages/home.js
```diff
@@ -1,95 +1,38 @@
+var dashboardGroup = require('./dashboardGroup');
+
+var accessdashboardGroup = dashboardGroup({
+	groupName: 'Access',
+	tabs: [
+		{name: 'users', items: '1 Item'},
```
webteckie: why need to specify the item count here?  what if the test changes it?  for sure I missed something :-)

webteckie:
@joerter seems like you found an easy pattern for groups/tabs...like it!  Had a minor nitpick about why need to specify the item count in the group tab definition.  Otherwise, please rebase onto master to resolve conflicts and should be ready to go!  @jstockwin may have to catch up all the test list he added with this pattern once it's merged :-)

jstockwin:
This looks good! Happy to have a go at something similar for the field tests, although some fields have different tests to others, e.g. where there are selectors. Could try and implement the standard tests that happen for all of them like this, and then add additional files for the extra tests on only the fields that need it.

webteckie:
I wasn't suggesting refactoring the fields test pattern but certainly if there's a better and simpler way let's hear it :-)

webteckie:
What I was suggesting is that we need to add tabs for all the test lists for all the fields you added @jstockwin

jstockwin:
@webteckie ah okay, will do. 

I just feel as though there is a lot of code repeated for each test. Every field is exactly the same in structure, and most of the ui/ux tests use the same label and values. We could potentially have a uiTestAllFields which has an array of fields to loop over and do the basic checks on. Would mean less repeated code, but it potentially makes it a bit confusing as e.g. color would be checked once in the 'all' checks, and then we'd need to check the color picker was there separately. Perhaps it's better as it is - all the color checks are in one place.

webteckie:
For this problem domain, I take simpler over efficient anytime :=)   That being said if you can still achieve simplicity and eliminate some of the repetition and still try to keep a sense of single responsibility the by all means give it a try and we can review and decide.

jstockwin:
@webteckie Well for example, every single (refactored) uiTest__Field.js is identical - they have before and after functions to open the webpage, and then a test function which calls `verifyUI` on `fieldA` and `name`. Why not have a uiTestAllFields.js, which initialises an array of field types, loops over them and calls verifyUI on each one? 

Having said this, I can see it's probably more complicated than this. Is such a for loop possible in these files..? I just tried a `for (var field in fields)` and it immediately errors `Unexpected token var`.

jstockwin:
@webteckie As a very quick example, see [this](https://github.com/jstockwin/keystone/commit/b94f8fa116afaf55dbc29d2b8cdba4e679f813e2).

All ui tests run as they did previously, but now we just have to update a value in an array. Not saying this is a very good implementation, but just a proof of concept. I kind of wanted the loop outside of the `'All field should show correctly in the initial modal': function (browser) {...` to add each one as it's own function, but it didn't seem to like that.

joerter:
@webteckie: The item count is there because Users has an initial item count of 1 where every other list has 0 items to start. For that reason, I decided to make it configurable. I think I'll rename that key to initialItemCount to be more clear.

joerter:
Okay, rebased and resolved conflicts!

test/e2e/adminUI/pages/dashboardTab.js
```diff
@@ -0,0 +1,25 @@
...
+		commands: [{
+			verifyUI: function() {
+				this
+					.expect.element('@label').text.to.equal(utils.titlecase(config.name));
+				this
+					.expect.element('@plusIconLink').to.be.visible;
+				this
+					.expect.element('@plusIconLink').to.have.attribute('class').which.contains('dashboard-group__list-create octicon octicon-plus');
+				this
+					.expect.element('@itemCount').text.to.equal(config.items);
```
webteckie: please `return this` in case we need to chain other operations.

test/e2e/adminUI/tests/group002Home/uxTestHomeView.js
```diff
@@ -2,107 +2,121 @@ var adminUI = require('../../adminUI');
...
-			.pause(browser.globals.defaultPauseTimeout)
-			.url(adminUI.url)
-			.waitForElementVisible(adminUI.cssSelector.homeView.id)
-			.pause(browser.globals.defaultPauseTimeout)
-			.click(adminUI.cssSelector.allView.fieldListsMenu)
-			.waitForElementVisible(adminUI.cssSelector.listView.id)
-			.pause(browser.globals.defaultPauseTimeout);
+	'Home view should allow clicking a nav menu item such as Access and Fields to show the list of items': function (browser) {
+		browser.app
+			.navigate()
```
webteckie: you already do the app.navigate and waiting for homeScreen in the before function...no need to do it here?

test/e2e/adminUI/tests/group002Home/uxTestHomeView.js
```diff
@@ -2,107 +2,121 @@ var adminUI = require('../../adminUI');
...
-		browser
-			.url(adminUI.url)
-			.waitForElementVisible(adminUI.cssSelector.homeView.id)
-			.pause(browser.globals.defaultPauseTimeout)
-			.click(adminUI.cssSelector.homeView.usersTabUnderDashboardAccessSubheading)
-			.waitForElementVisible(adminUI.cssSelector.listView.id)
-			.pause(browser.globals.defaultPauseTimeout);
+		browser.app
+			.navigate()
+			.waitForElementVisible('@homeScreen');
```
webteckie: ditto

test/e2e/adminUI/tests/group002Home/uxTestHomeView.js
```diff
@@ -2,107 +2,121 @@ var adminUI = require('../../adminUI');
...
-		browser
-			.url(adminUI.url)
-			.waitForElementVisible(adminUI.cssSelector.homeView.id)
-			.pause(browser.globals.defaultPauseTimeout)
-			.click(adminUI.cssSelector.homeView.plusIconLinkForUsersTabUnderDashboardAccessSubheading)
-			.waitForElementVisible(adminUI.cssSelector.initialModalView.id)
-			.pause(browser.globals.defaultPauseTimeout);
+		browser.app
+			.navigate()
+			.waitForElementVisible('@homeScreen');
```
webteckie: ditto--the before function runs before each one of these, right?  Or does it run before the test file starts executing?

test/e2e/adminUI/tests/group002Home/uxTestHomeView.js
```diff
@@ -2,107 +2,121 @@ var adminUI = require('../../adminUI');
...
-			.pause(browser.globals.defaultPauseTimeout)
-			.click(adminUI.cssSelector.initialModalView.buttonCreate)
-			.waitForElementVisible(adminUI.cssSelector.itemView.id)
-			.url(adminUI.url);
-
-		browser.expect.element(adminUI.cssSelector.homeView.itemCountForNamesTabUnderDashboardFieldsSubheading)
-			.text.to.equal('1 Item');
+		browser.app
+			.navigate()
+			.waitForElementVisible('@homeScreen');
```
webteckie: ditto

webteckie:
@joerter sorry, this time did the review from my desktop (prior tried to do in my cell) and this time had some more comments...and, yeah, initialItemCount I think is more clear :-)  I will merge this as-is and then you can do the updates on a follow up.  Thanks!

PR #2732, opened 2016-05-01
testing(e2e): Added html field testing

jstockwin:
cc @webteckie 

In line with work on #2291


mxstbr:
Awesome!

webteckie:
@jstockwin curious, does the html field also take actual html text with html tags as input?  Anyways, this looks good...merging.

jstockwin:
@webteckie yes, but it's just a textarea with no validation. Just like urls don't verify either.

PR #2736, opened 2016-05-02
test(e2e): Use common functions for field tests

jstockwin:
cc @webteckie 
#2291 

Replaces #2735

I've made the changes you suggested here, and am also allowing different fields to be in each test. Say `url` was the only type to also have a `fieldC`, perhaps because we want to check it accepts both `www.example.com` and `example.com` type addresses. Then the function call in `uxTestUrlField` would change to include a `fieldC` input, and provided the `urlList` has this field, everything will just work.  

Let me know if you have any questions. A possible future improvement would be to split `assertEditFormUX` into `fillEditFormUX` and `assertEditFormUX`. This would allow us to `assertInitialFormUX(... name: ... , fieldA: ... )`, then `fillEditFormUX(.... fieldB: ...)`, but then `assertEditFormUX(... name: ..., fieldA: ..., fieldB: ...)` which has the advantage of checking the data inputted in the initial form are not lost when entering additional info into the edit form.

### DIFF
--- a/test/e2e/adminUI/tests/group005Fields/commonFieldTestUtils.js
+++ b/test/e2e/adminUI/tests/group005Fields/commonFieldTestUtils.js
@@ -38,10 +38,10 @@ module.exports = {
 			});
 		}
 	},
-	assertInitialFormUX: function(config){
+	openInitialFormUX: function(config) {
 		var list = config.listName.toLowerCase() + 'List';
 		var listSubmenu = '@' + list + 'Submenu';
-		return function (browser) {
+		return function (browser){
 			browser.app
 				.click('@fieldListsMenu')
 				.waitForElementVisible('@listScreen')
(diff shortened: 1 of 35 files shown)


jstockwin:
Strange problem, when running all the group005 tests in one go, I hit the error given below in the middle of the `textarea` testing, however if I run the `textarea` tests on their own, this doesn't happen. 

Maybe it was just a one off thing, will try and run them again. Strange for it to get that far and then error, especially as the test completes on it's own?

```
Running:  Textarea field can be filled via the edit form
POST /keystone/textareas/57271e20b8b30d30d6f9c87e 302 5.254 ms
GET /keystone/textareas/57271e20b8b30d30d6f9c87e 200 5.123 ms
GET /keystone/api/textareas/57271e20b8b30d30d6f9c87e?drilldown=true 200 2.834 ms
 ? Element <#item-view> was visible after 109 milliseconds.
 ? Expected element <.Alert > span:nth-child(1)> text to equal: "Your changes have been saved." - condition was met in 105ms
 ? Element <Section[name=form],Section[name=textareaList],Section[name=fieldB],Element[name=@value]> was visible after 268 milliseconds.
 ? Passed [equal]: success == success
 ? Passed [equal]: Some text for field B == Some text for field B

OK. 5 assertions passed. (2.297s)
 ? Element <.primary-navbar [data-section-label="octicon-sign-out"]> was visible after 107 milliseconds.
GET /keystone/signout 302 0.946 ms
GET /keystone/signin?signedout 200 1.896 ms
 ? Element <#signin-view> was visible after 104 milliseconds.

SyntaxError: Unexpected token )
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:387:25)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at new Module (D:\Documents\Development\GitHub\keystone\node_modules\nightwatch\lib\runner\module.js:7:23)
    at new TestSuite (D:\Documents\Development\GitHub\keystone\node_modules\nightwatch\lib\runner\testsuite.js:17:17)
    at runTestModule (D:\Documents\Development\GitHub\keystone\node_modules\nightwatch\lib\runner\run.js:38:19)
    at runNextModule (D:\Documents\Development\GitHub\keystone\node_modules\nightwatch\lib\runner\run.js:191:23)
    at null._onTimeout (D:\Documents\Development\GitHub\keystone\node_modules\nightwatch\lib\runner\run.js:199:15)
    at Timer.listOnTimeout (timers.js:92:15)
```

test/e2e/adminUI/tests/group005Fields/url/uiTestUrlField.js
```diff
@@ -1,46 +1,8 @@
...
-		browser.initialFormPage.section.form
-			.click('@cancelButton');
-
-		browser.app
-			.waitForElementVisible('@listScreen');
-	},
+module.exports = {
+	before: fieldTests.before,
+	after: fieldTests.after,
+	'Url field should be visible in initial modal': fieldTests.assertInitialFormUI({fieldName: {fieldName: 'Url', fields: ['name', 'fieldA']}),
```
webteckie: You're missing a `}` in this line :-)

webteckie:
@jstockwin see comment above as to why you're seeing the error :-)

test/e2e/adminUI/tests/group005Fields/url/uiTestUrlField.js
jstockwin: Or rather, there's an extra `{fieldName: `, but good spot - thanks!

jstockwin:
@webteckie thanks. Pushed a fix - I amended the commit to keep the history clean, hope that's not messed things up too much.

test/e2e/adminUI/tests/group005Fields/commonFieldTestUtils.js
```diff
@@ -0,0 +1,114 @@
...
+
+		browser.signinPage.signin();
+		browser.app.waitForElementVisible('@homeScreen');
+	},
+	after: function (browser) {
+		browser.app.signout();
+		browser.end();
+	},
+	assertInitialFormUI: function (config) {
+		var nameLowercase = config.fieldName.toLowerCase();
```
webteckie: I would prefer this to be:
```
		var listName = config.fieldName.toLowerCase() + 'List';
```

and then use it like:
```
browser.initialFormPage.section.form.section[listName].section[field].verifyUI();
```

```

jstockwin: @webteckie Pushed a commit to do this.

jstockwin:
@webteckie See latest commit, which adds more granular functions. 

I think this will really come into its own when the functions are moved from `commonFieldTestFunctions.js` into commands of `initialForm.js` and `item.js`.

jstockwin:
Closing in favour of #2740

Issue #2715, opened 2016-04-29
No 'for' property set in the datetime html


jstockwin:
Causes testing to fail in #2714 

Currently, with two fields, the html is
```html
<div class="FormField field-type-datetime">...</div>
<div class="FormField field-type-datetime">...</div>
```
whereas it needs to be 
```html
<div class="FormField field-type-datetime" for="fieldA">...</div>
<div class="FormField field-type-datetime" for="fieldB">...</div>
```

Without this, when checking for the fieldB stuff, the testing goes into the first div it matches, which is for fieldA, and then everything fails. 

I was thinking the PR would fail the CI checks, but presumably CI is not currently checking datetime. If this PR is merged as is, all future checks will fail.

mxstbr:
That sounds reasonable. Is that a blocker for #2714, and does it only happen for the datetime field?

jstockwin:
It only happens for the datetime field, but the current testing practise adds a model with a name and two datetime fields, fieldA and fieldB. FieldA is initial, and fieldB is not. 

This means on the view where fieldB is checked, there should be two datetime fields on the same page, thus causing an error due to not having a for field. 

So yes it's a blocker for #2714, but it also only happens for the datetime field. I could change the datetime testing to not include fieldB tests for now, but that's inconsistent with the other tests. 

I hope that's clear

mxstbr:
That's a perfectly clear explanation, thanks! The problem is that I don't really want to tinker with the admin interface itself right now, since #2566 is still WIP, will be merged soon and won't have those updates.

What do you think we should do? Can this wait?

jstockwin:
This is all for work on #2291. I'd probably defer to @webteckie for a definitive answer on where to go, but personally I would do the following:

I could make a new PR which only concerns itself with FieldA for now. I'd say that's a good temporary solution, as it still does some testing on the datetime field, while circumventing this issue. Then when this issue is closed, we can add fieldB back into the testing.

jstockwin:
@mxstbr what's the best way of doing this temp fix? Should I comment out the lines relating to fieldB, and comment that they are pending this issue being resolved, or should I remove the lines altogether to keep the code clean?

mxstbr:
Comment them out for now, it'd be weird to then have to duplicate the work afterwards. :+1:

webteckie:
[...]

PR #2751, opened 2016-05-03
test(e2e): move utils functions to page object commands

jstockwin:
cc: @webteckie 
For #2291 

All functions currently in `commonFieldTestUtils` have been moved to their respective page objects as commands, aside from the `before` and `after` functions. This gives a much nicer pattern in the test files themselves, and seems more in keeping with using page objects properly. 

All fields have been refactored to use this new pattern, with the exception of Boolean, Email and Number, as these are still on the old pattern which doesn't use page objects at all. 

Further, the `assertUI` functions for code and select don't actually check which field you want them to carry out the assertion on, so the `Code/Select can be filled via the edit form` checks are currently commented out as they fail. 

This same check for Datetime is also commented out pending a fix for #2715. 

Currently haven't managed to run through the entire group of tests without firefox popping up and asking to be started in safe mode half way through, which seems to kill it. Will keep trying, and let you know when I get there, but most of them have run and nothing has failed yet.


jstockwin:
@webteckie with the latest commit all tests pass locally :tada: 
`OK. 628 total assertions passed. (8m 17s)`

PR #2763, opened 2016-05-05
modified x-array items behavior to be consistent with #2721

suryagh:
Modified x-array items behavior to be consistent with #2721

**Note: this is a breaking change**. Probably should revisit this once we have the re-factoring of the `admin UI` merged.

Also, I did not notice the original issue in question with: `CloudinaryImages` and `LocalFiles`.

### DIFF
--- a/fields/types/datearray/DateArrayType.js
+++ b/fields/types/datearray/DateArrayType.js
@@ -182,24 +182,22 @@ datearray.prototype.updateItem = function (item, data, callback) {
 
 	var value = this.getValueFromData(data);
 
-	if (value !== undefined) {
-		if (Array.isArray(value)) {
-			// Only save valid dates
-			value = value.filter(function (date) {
-				return moment(date).isValid();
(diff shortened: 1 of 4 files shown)


mxstbr:
This looks good to me!

We rewrote those entire functions just a few weeks ago, they aren't released yet, so this isn't really a breaking change…

mxstbr:
I've marked it as ready, we'll see what @JedWatson says. :blush:

suryagh:
@mxstbr thank you for the feedback. Meanwhile, is there any in-progress pull request (or branch) up somewhere for those re-writes you are referring? Just wanted to see how it looks now :)

JedWatson:
@suryagh what @mxstbr means is that the code you've changed in this PR is only a few weeks old, and those code paths aren't really active yet unless you're using the new API (which, until the new Admin UI is merged, nobody is anyway!)

Thanks for your work :)

PR #2764, opened 2016-05-05
test(e2e): Add testing for password field

jstockwin:
cc @webteckie #2291 

Added password field checking. It's not possible to assertInput on the password fields, since they never display the password (obviously!). However, I've added a check in that enters different password into the `password` and `confirm password` fields, and check that it throws an error when this happens. 

Also fixed m coming before n in the alphabet >.>


mxstbr:
:fire: :fire: :fire: :fire: :clap: @jstockwin :fire: :fire: :fire: :fire:

PR #2806, opened 2016-05-10
Fix #2803 Keystone.lists.map is not a function

### LINKED ISSUE #2803: Keystone.lists.map is not a function
I've migrated to current master and i'm getting the error `Keystone.lists.map is not a function`

Anybody can help me on why e2e tests works, but my wesite is not?

I'll also try the `keystone-demo` github repo with latest keystone from upstream to check

jacargentina:
Fixes #2803

Keystone.lists now is an object, not an array as previously


mxstbr:
Good catch, awesome! Thanks so much!

Issue #2808, opened 2016-05-10
Validation errors don't show in the admin ui


jstockwin:
Here is a sample validation error:
```
########## Console.log the error:
{ [ValidationError: SelectFieldOnInitial validation failed]
  message: 'SelectFieldOnInitial validation failed',
  name: 'ValidationError',
  errors:
   { food:
      { [ValidatorError: Path `food` is required.]
        message: 'Path `food` is required.',
        name: 'ValidatorError',
        properties: [Object],
        kind: 'required',
        path: 'food',
        value: undefined } } }
POST /keystone/api/legacy/select-field-on-initials/create 400 19.532 ms
```

The code which throws the error in `admin/server/api/list/legacyCreate.js` is as follows:
```
if (err) {
	if (err.name === 'ValidationErrors') {
		return res.apiError(400, 'validation errors', err.errors);
	} else {
		return res.apiError(500, 'error', err);
	}
}
```

The error is displayed as just "Error" in the admin ui, a 500 error is logged to Chrome's console, because the if statement is not being triggered due to plurals. 

Will PR a fix that adds an `s` to the end, but this assumes that in all cases a `ValidationError` is thrown, and not a `ValidationError`. This is potentially part of an inconsistency problem, rather than just an innocent typo, but I'm not sure how to check this. @mxstbr

jstockwin:
We fixed this in #2809 :)

Issue #2586, opened 2016-04-07
[0.3] validator errors incorrectly parsed


creynders:
It isn't entirely clear whether something changed in KS or mongoose, but now if model fails validation the message shown in userland hides the details:

```
There was an error creating the new Post:
Post validation failed
```

Anybody any idea where the change was? Did mongoose start spewing out Errors with a different structure, or?

This is with mongoose v4.4.4.

webteckie:
These are the kinds of regressions we're trying to prevent with e2e...you just don't know something broke until some time later...very expensive! :-)

creynders:
Tell me about it! ;)

creynders:
I'm pretty sure this commit baff5604292ee77d1391d96d2e8474ed3e4a629d probably fixed some validation errors, but introduced a new one.
If you [look at the diff](https://github.com/keystonejs/keystone/commit/baff5604292ee77d1391d96d2e8474ed3e4a629d#diff-1216fa2cb9fb0cbc959b8bdba7272512R154) you can see there used to be two types of errors: "ValidatorError" and "ValidationError" (as it is the case in Mongoose, a ValidationError contains multiple ValidatorErrors in its `errors` prop), however this commit uses only ValidatorError.

creynders:
Fixed it in my own fork, be reverting the error names. I have no idea whether this breaks something else, i.e. the bug the original commit was trying to fix in the first place...

creynders:
Of course it wasn't as simple as that ... It didn't work by reverting the error names, but I thought it did. Turns out I tested the wrong situation. The buggy situation is the following: when field validation fails with a custom validator for an `initial` field, while all other fields validate... Phew...

mxstbr:
Since this is fixed (thanks @creynders), I'll close the issue!

PR #2809, opened 2016-05-10
Fix validation errors not showing in the admin UI

jstockwin:
@mxstbr - a potential fix for #2808

### DIFF
--- a/admin/server/api/list/legacyCreate.js
+++ b/admin/server/api/list/legacyCreate.js
@@ -7,7 +7,7 @@ module.exports = function (req, res) {
 	var item = new req.list.model();
 	item.getUpdateHandler(req).process(req.body, { flashErrors: false, logErrors: true }, function (err) {
 		if (err) {
-			if (err.name === 'ValidationErrors') {
+			if (err.name === 'ValidationError') {
 				return res.apiError(400, 'validation errors', err.errors);
 			} else {
 				return res.apiError(500, 'error', err);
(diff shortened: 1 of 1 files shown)


mxstbr:
That looks reasonable. When was that `s` added, who put it there?

jstockwin:
It hasn't been updated since it was [created](https://github.com/keystonejs/keystone/commit/4988eb7f20705e926b932936eced3975e5c5e850) 6 days ago (Sorry @JedWatson !)

creynders:
No, the file was moved. (Sidenote: a bit surprising GH can't track history over file renames/moves)
I thought it only happened in 0.3, see #2586 
The change was introduced in baff560
The solution isn't straight forward thought, It's a cess pool TBH

jstockwin:
Ah okay, thanks @creynders. Does that mean this PR won't actually fix the issue? 

Is it related to #2814, which is what caused me to notice the problem in the first place!

JedWatson:
@jstockwin does this resolve the issue for you in your testing? If so, I'm happy to merge it.

jstockwin:
@JedWatson It resolves that specific issue yes, but I was worried it might have the opposite affect on other errors, i.e. if the error name _is_ sometimes `ValidationErrors`, and that's why it was like this in the first place.

JedWatson:
Since this is a temporary thing anyway, let's accept both in that condition maybe?

jstockwin:
Fine with me. Will update the PR - will have to wait until a little later if that's okay though I'm afraid

jstockwin:
@JedWatson PR updated to accept both.

JedWatson:
:+1: Thanks @jstockwin!

PR #2810, opened 2016-05-10
test(e2e): Add testing for relationship field

jacargentina:
Initial work for testing relationship fields #2291


webteckie:
@jacargentina we are trying to get rid of the adminUI.js in favor of page objects.  Please see all existing page objects under `test/e2e/adminUI/pages` and see if you can follow the pattern and do something similar.  tnx

jstockwin:
@jacargentina Did you want to have a crack at adding the page objects and test files too? 

If not, I'm planning to get a few of the other fields done over the next few days and can do this one too.

jacargentina:
@jstockwin go with it

jstockwin:
In which case, I'll close this PR. Thanks for adding what you did, if any of it can be converted to page objects easily then I'll start with your code :)

PR #2822, opened 2016-05-11
Adds the utility keystone.listFor(item)

jacargentina:
For a given model instance get back the ketystone List instance

Works by comparing the collection name on the schema for the item, and the available lists on `keystone.lists`


jacargentina:
@JedWatson clarified on slack about having already the `list` available on each instance.

PR #2833, opened 2016-05-12
Remove date assert e2e tests, re-enable travis e2e

jstockwin:
Date and datetime input assertions have been temporarily removed, and a TODO has been added in the code. This is because timezones are causing problems and making travis fail. 

Travis e2e tests have been re-enabled, as it was only these tests which were failing. 

I had a quick look, and couldn't find an open issue on this. I'll open one unless anyone can point me at one that's already open.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ before_script:
 
 script:
   - npm run test-cov
+  - npm run test-e2e-saucelabs
 
 sudo: required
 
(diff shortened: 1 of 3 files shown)


jstockwin:
Found a related issue:
https://github.com/keystonejs/keystone/issues/1345

mxstbr:
:+1: :tada:

PR #2850, opened 2016-05-14
test(e2e): Add location and key field tests

jstockwin:
In line with #2291 

cc @webteckie bit of a monster this one - location adds 202 assertions, and key adds 41. 

I've added an "args" parameter to the config of `assertUI`, which gets passed into the individual `verifyUI` commands. This was needed as I need to tell verifyUI if it's checking before or after the show more button has been clicked. 

@mxstbr See the diff on `fields/types/location/LocationField.js`. Similar issue to #2715, and was causing testing to break until I did this.


webteckie:
@jstockwin this looks 👍

PR #2854, opened 2016-05-14
test(e2e): add ui tests for edit form

jstockwin:
cc @webteckie 

Password and Select were a little different, as they look different on the initial form compared with the edit form in this case. (Passwords display as a "Set Password" button in the edit form, but as input fields in the initial form, Select's placeholder is not there in the edit form, since we have already selected the value before save). I've therefore added an `args` parameter to their verifyUI commands, and pass in whether I'm calling it from the editForm or not. If you think there's a better way of doing this then let me know. 

Similarly, notice we don't pass the "name" field into the edit form's verifyUI. This is because in the edit form there is no label next to the name's test field, which causes the tests to fail. My fix for this was similar to the above, but that made the code messy in every test file, which I decided it was best not to do, so we just skip checking the name field. I figured we're running these tests for FieldA and fieldB really, the name field is just there because it has to be.

### DIFF
--- a/test/e2e/adminUI/pages/fieldTypes/password.js
+++ b/test/e2e/adminUI/pages/fieldTypes/password.js
@@ -16,13 +16,20 @@ module.exports = function PasswordType(config) {
 					.waitForElementVisible('@value');
 				return this;
 			},
-			verifyUI: function() {
+			verifyUI: function(args) {
 				this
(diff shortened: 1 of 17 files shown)


jstockwin:
This also takes us to 1122 e2e assertions, breaking the 1000 barrier by quite a long way :tada:

PR #2856, opened 2016-05-15
test(e2e): Add cloudinary field e2e tests

jstockwin:
cc @webteckie #2291 

Some slightly annoying names ending in `x` to deal with plurals. Also, we need `cloudinaryimageList` in some places (e.g. [here](https://github.com/keystonejs/keystone/blob/master/test/e2e/adminUI/pages/initialForm.js#L84)) due to calling `.toLowerCase` to get the list name. Can change this if you prefer.

@mxstbr: Again, I had to add a couple of `htmlFor` properties. Hope that's ok!


mxstbr:
LGTM, that's obviously very alright!

JedWatson:
Looks good to me! Thanks @jstockwin

PR #2878, opened 2016-05-17
Add Region to S3 Image URL

alechewitt:
The current hard coded format for S3 location:
`https://s3.amazonaws.com/[BUCKET]/[FILENAME]`
Doesn't work when the s3 bucket is located in a different AWS region.
Instead the subdomain s3 needs to be replaced by the region specific subdomain:
`https://[S3_REGION].amazonaws.com/[BUCKET]/[FILENAME]`
See here for list of s3 subdomains for different regions: http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region


mxstbr:
This LGTM, thanks!!

JedWatson:
Looks great @alechewitt, thanks!

PR #2880, opened 2016-05-17
React component unit tests

suryagh:
1. React component test setup
2. First test for `ItemsTableCell.js`

@JedWatson @mxstbr I wanted to throw this out there; let me know your views. 

Note: I have not hooked this to `CI` pipeline yet. But that should be easy if we plan to go forward with this.

Edit: `npm run test-react`


suryagh:
updated the `pr`

mxstbr:
Good idea! I'm a big fan of `enzyme`, what do you think about using that instead of `TestUtils.createRenderer()`?

suryagh:
@mxstbr I looked at the `enzyme`, it looks much better that `test-utils` though the latter if through facebook. Thanks for the pointer. I have updated the `pr`.

mxstbr:
This is great @suryagh, love it! LGTM!

JedWatson:
Looks great, I'm merging this & will update the CI pipeline!

(want to have a play with this now in master 😀)

PR #2884, opened 2016-05-18
e2e array field tests

jstockwin:
cc @webteckie #2291 

Thought I'd PR this one, as I've slightly changed the way I'm selecting the date inputs in datearrays. This was is much cleaner and messing around with the IDs that update, and further those IDs don't exist for numberarrays. 

Let me know what you think.


test/e2e/adminUI/pages/initialForm.js
```diff
@@ -16,10 +16,12 @@ var MarkdownList = require('./lists/markdown');
 var MoneyList = require('./lists/money');
 var NameList = require('./lists/name');
 var NumberList = require('./lists/number');
+var NumberArrayList = require('./lists/numberArray');
```
webteckie: goes before NumberList :-)

test/e2e/adminUI/pages/initialForm.js
```diff
@@ -49,10 +51,12 @@ module.exports = {
 				moneyList: new MoneyList(),
 				nameList: new NameList(),
 				numberList: new NumberList(),
+				numberarrayList: new NumberArrayList(),
```
webteckie: goes before NumberList :-)

jstockwin:
@webteckie: True. Why does eslint not pick that up? Anyway, should be fixed in latest commit.

Issue #2865, opened 2016-05-16
The ui with TextArray type doesn't work well.


w01fgang:
### Expected behavior
Open and edit the records in admin panel.
### Actual behavior
There is an error `TypeError: Cannot read property 'map' of undefined` in `fields/mixins/ArrayField.js ` in 
```
renderField: function () {
		return (
			<div>
				{this.state.values.map(this.renderItem)}
				<Button ref="button" onClick={this.addItem}>Add item</Button>
			</div>
		);
	},
```
`this.state.values` is undefined
### Steps to reproduce the behavior
Enter to admin panel -> first open record for model without `TextArray` type -> then open record of model with `TextArray` type.
If I enter in the record of model with `TextArray` type, `this.state.values` in an empty array, but after enter in record without `TextArray` type it's becoming `undefined`.

And another:
I create item in `TextArray` field. If I paste numbers all works well, but if I paste text, it clears after saving.

This is the latest version from `master`

mxstbr:
Awesome bug report, thanks @w01fgang!

jstockwin:
I just got the same thing with DateArray too. Same error, same place. 

Further, after this error has been thrown, no item is created and when I try to navigate between tabs I get 
`TypeError: Cannot read property 'componentWillReceiveProps' of null` at `packages.js:15`

w01fgang:
@jstockwin yes, I also have this error.
Accordingly the error in the `ArrayField` mixin.

jstockwin:
So, `this.state` is somehow stuck onto the last field you viewed. 
So if you create an email `test@example.com`, and then try and view the date array, `this.state` in the snippet of the original issue is `values: 'test@example'com'`, even though you're currently viewing a datearray.

PR #2886, opened 2016-05-18
Fix text array behaviour

jstockwin:
Potential fix for #2865 

It certainly resolves the issue, but I'm perhaps not addressing the underlying problem? 
Sometimes, `renderUI` is called when `this.state.values.map` is not a function, however `this.renderItem` is. 

Always using `this.renderItem` doesn't work all the time either, specifically it breaks the buttons to add new fields. 

Happy to be pointed in the right direction of a better fix if this is not it.

### DIFF
--- a/fields/mixins/ArrayField.js
+++ b/fields/mixins/ArrayField.js
@@ -72,12 +72,21 @@ module.exports = {
 	},
 
 	renderField: function () {
-		return (
-			<div>
-				{this.state.values.map(this.renderItem)}
-				<Button ref="button" onClick={this.addItem}>Add item</Button>
-			</div>
-		);
+		if (this.state.values !== undefined && this.state.values.map !== undefined) {
+			return (
+				<div>
+					{this.state.values.map(this.renderItem)}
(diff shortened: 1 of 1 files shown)


suryagh:
@jstockwin can you see if the issue source is at line [23](https://github.com/keystonejs/keystone/blob/394a60b7afd24beb4f8844207a76651694ae2f32/fields/mixins/ArrayField.js#L23):

try by changing it to: 
`values: Array.isArray(this.props.value) ? this.props.value.map(newItem) : [],`

jstockwin:
@suryagh Nice! Yeah that does fix it. Will just open a new PR since what I've done here is a bit pointless.

suryagh:
@jstockwin I think you can push the updated change to the same `jstockwin:array-fix` branch, and use this `pr`. With this approach you can have the history of the entire conversation in one place for anyone else interested.

jstockwin:
@suryagh Yes, but since my original commit doesn't really solve the problem properly it'd need to be reverted. Figured in this case it'd be better to open a new PR to keep the commit history clean if it gets merged. Can always reference this PR from the new one.

PR #2894, opened 2016-05-19
Fix timezone handling for Datetime field

### LINKED ISSUE #1345: DateTime field type timezone bug

geloescht:
## Description of changes

Changed behaviour of Datetime field type:
* UTC date is always stored in database and used on the server
* Client sends timezone offset of currently set date & time to the API when saving
* If field option `utc: true` is set, UTC is used on the client, local time otherwise

## Related issues (if any)

Fixes #1345 for v0.4 (not v0.3!)

## Testing

Passed `npm run test`, going check e2e tests later. Feel free to leave pull request open until then.

### DIFF
--- a/fields/types/datetime/DatetimeField.js
+++ b/fields/types/datetime/DatetimeField.js
@@ -22,7 +22,7 @@ module.exports = Field.create({
 		return {
 			dateValue: this.props.value && this.moment(this.props.value).format(this.dateInputFormat),
 			timeValue: this.props.value && this.moment(this.props.value).format(this.timeInputFormat),
-			tzOffsetValue: this.props.value ? this.moment(this.props.value).format(this.tzOffsetInputFormat) : moment().format(this.tzOffsetInputFormat),
+			tzOffsetValue: this.props.value ? this.moment(this.props.value).format(this.tzOffsetInputFormat) : this.moment().format(this.tzOffsetInputFormat),
 		};
 	},
 
@@ -32,15 +32,14 @@ module.exports = Field.create({
 		};
 	},
 
-	moment (value) {
-		var m = moment(value);
-		if (this.props.isUTC) m.utc();
-		return m;
+	moment () {
+		if (this.props.isUTC) return moment.utc.apply(moment, arguments);
+		else return moment.apply(undefined, arguments);
 	},
 
 	// TODO: Move isValid() so we can share with server-side code
 	isValid (value) {
-		return moment(value, this.parseFormats).isValid();
+		return this.moment(value, this.parseFormats).isValid();
 	},
 
 	// TODO: Move format() so we can share with server-side code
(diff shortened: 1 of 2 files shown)


jstockwin:
@geloescht Nice one!

If you would also like to do the honours of un-commenting all the TODO tests in
`test/e2e/adminUI/tests/group005Fields/testDateArrayField.js` 
`test/e2e/adminUI/tests/group005Fields/testDateFieldField.js` 
`test/e2e/adminUI/tests/group005Fields/testDatetimeField.js` 
that would be amazing. (I think there are three chunks of code commented out in each file).

geloescht:
@jstockwin Thanks for pointing me at the proper tests.
Nothing is changed about Date or DateArray field types. To me, those are separate issues, because conceptually a date is not a moment in time. I need to think that over and do some research on how this is handled in other projects.

jstockwin:
@geloescht ah fair enough. I'd assumed it was the same problem, but I guess that doesn't really make sense. (Unless Dates are actually just stored with a time of midnight?)

mxstbr:
This looks really good!

Travis is failing currently because of this:

```
react-addons-test-utils is an implicit dependency in order to support react@0.13-14. Please add the appropriate version to your devDependencies. See https://github.com/airbnb/enzyme#installation
```

Would you mind fixing that in this PR too so we can merge it? Thanks!

geloescht:
@mxstbr Jed did that just an hour ago. Rebased on current master.
Please don't merge yet, I found some issues in e2e testing.

mxstbr:
Ah yeah, just noticed, sorry!

JedWatson:
Looks great @geloescht, thanks!

geloescht:
Oi, this fails e2e tests if we enable them (they are disabled because of #1345). I wanted to fix at least those before merging. But no worries, I'm going to file another PR for that (if I manage to fix them...)
I just noticed that time handling in Datetime fields is still utterly broken, with or without this patch. The time value is not even validated. Enter an invalid value and the whole field will just be reset to 1970.

andrew-kennedy:
is this pull request feasibly cherry-pickable into my own fork of 0.3.x? I'm wondering if it's even worth looking into it

geloescht:
No, it mainly deals with frontend code that has been rewritten from scratch using React.

PR #2933, opened 2016-05-24
Fix array fields behaviour #2865

### LINKED ISSUE #2865: The ui with TextArray type doesn't work well.

jstockwin:
## Description of changes
cc @suryagh 
Updated the `getInitialState` function in `ArrayField.js`

## Related issues (if any)

Should fix #2865 

## Testing

### DIFF
--- a/fields/mixins/ArrayField.js
+++ b/fields/mixins/ArrayField.js
@@ -20,7 +20,7 @@ function reduceValues (values) {
 module.exports = {
 	getInitialState: function () {
 		return {
-			values: Array.isArray(this.props.value) ? this.props.value.map(newItem) : this.props.value,
+			values: Array.isArray(this.props.value) ? this.props.value.map(newItem) : [],
 		};
 	},
 


geloescht:
LGTM!

fields/mixins/ArrayField.js
```diff
@@ -20,7 +20,7 @@ function reduceValues (values) {
 module.exports = {
 	getInitialState: function () {
 		return {
-			values: Array.isArray(this.props.value) ? this.props.value.map(newItem) : this.props.value,
+			values: Array.isArray(this.props.value) ? this.props.value.map(newItem) : [],
```
jbpin: ```values: Array.isArray(this.props.value) ? this.props.value.map(newItem) : [].concat(this.props.value)``` ??

JedWatson:
Not sure why this got left for so long - thanks @jstockwin! Merging now.

PR #2934, opened 2016-05-24
Check blacklisted lists when deleting resources

fidelisrafael:
## Description of changes

I'm know that Keystone uses mongoDB as default(at least only for now) database driver, and for this reason primary key `_id` is a `ObjectId` value(which is generated using  timestamp, machine ID, process ID, and a process-local incremental counter), and this make almost impossible to two collection share the **same id**, butttttt MongoDB also supports [`Auto Increment Field`](https://docs.mongodb.com/manual/tutorial/create-an-auto-incrementing-field/) as primary key, so in this scenario is possible to share **multiples EQUALS ids** through the whole database. 
(eg: user.id = 10, post.id = 10, notification.id = 10)

So, this change **only** make sure to validate if user is really trying to delete **YOURSELF**, and not simply check if `userId`(logged user id) is the same that one of the resources which will be deleted.

In the current scenario, suppose your user is the `id = 10`, and you try to deleted a post with `id = 10`, this will not be possible...
## Testing

### DIFF
--- a/admin/server/api/list/delete.js
+++ b/admin/server/api/list/delete.js
@@ -1,6 +1,9 @@
 var async = require('async');
 var keystone = require('../../../../');
 
+// lists which user.id will be validated through collection ids to avoid self-delete
+var LISTS_TO_CHECK_USER_PERMISSIONS = ['user']
+
 module.exports = function (req, res) {
 	if (!keystone.security.csrf.validate(req)) {
 		console.log('Refusing to delete ' + req.list.key + ' items; CSRF failure');
@@ -17,13 +20,16 @@ module.exports = function (req, res) {
 	if (!Array.isArray(ids)) {
 		ids = [ids];
 	}
+
 	if (req.user) {
+		var checkResourceId = (LISTS_TO_CHECK_USER_PERMISSIONS.indexOf(req.list.key.toLowerCase()) >= 0)
 		var userId = String(req.user.id);
-		if (ids.some(function (id) {
+		// check if user can delete this resources based on resources ids and userId
+		if (checkResourceId && ids.some(function (id) {
 			return id === userId;
 		})) {
-			console.log('Refusing to delete ' + req.list.key + ' items; ids contains current User');
-			return res.apiError(403, 'not allowed', 'You can not delete yourself');
+			console.log('Refusing to delete ' + req.list.key + ' items; resources ids contains current User id');
+			return res.apiError(403, 'not allowed', "You can not delete this resource");
 		}
 	}
 	var deletedCount = 0;
--- a/admin/server/api/list/delete.js
+++ b/admin/server/api/list/delete.js
@@ -2,7 +2,7 @@ var async = require('async');
 var keystone = require('../../../../');
 
 // lists which user.id will be validated through collection ids to avoid self-delete
-var LISTS_TO_CHECK_USER_PERMISSIONS = ['user']
+var LISTS_TO_CHECK_USER_PERMISSIONS = ['user'];
 
 module.exports = function (req, res) {
 	if (!keystone.security.csrf.validate(req)) {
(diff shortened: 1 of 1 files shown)


geloescht:
So this is to make sure users don't delete themselves? You should probably put that into the title of this PR.
What about situations where users actually want to delete their account? Can we maybe extend the API with an option to force deletion of a user? Then we can have an unregister dialog or ask the user to confirm a potentially dangerous operation before resubmitting the request.

jstockwin:
@geloescht No, users cannot delete themselves.

The issue is that it is possible for a User and a Post to share the same ID, and if this is the case then the user will not be able to delete the post, as Keystone detects that the item the user is deleting has the same ID as the user, and so thinks the user is trying to delete themselves, when in fact they're trying to delete the post. 

That was one long sentence, so I hope it makes sense. I've not actually checked any of this, just interpreting what the PR says.

mxstbr:
This looks like a reasonable solution, I'm only not a fan of the name `LISTS_TO_CHECK_USER_PERMISSIONS`. Any other ideas? (Also /cc @keystonejs/core-devs any input?)

creynders:
@fidelisrafael @mxstbr the user list name could be something else than "user", which is why it's stored in the "user model" setting, i.e. I'd drop `LISTS_TO_CHECK_USER_PERMISSIONS` and replace L25 with:

```js
var checkResourceId = (keystone.get('user model') === req.list.key);
```

mxstbr:
Good catch @creynders, sounds good!

fidelisrafael:
@creynders @mxstbr  Guys, thank you very much for this corrections and suggestions! I pushed the commit ;)

mxstbr:
LGTM! Thanks @fidelisrafael!

admin/server/api/list/delete.js
```diff
@@ -17,13 +17,17 @@ module.exports = function (req, res) {
...
 		var userId = String(req.user.id);
-		if (ids.some(function (id) {
+		// check if user can delete this resources based on resources ids and userId
+		if (checkResourceId && ids.some(function (id) {
 			return id === userId;
 		})) {
-			console.log('Refusing to delete ' + req.list.key + ' items; ids contains current User');
-			return res.apiError(403, 'not allowed', 'You can not delete yourself');
+			console.log('Refusing to delete ' + req.list.key + ' items; resources ids contains current User id');
+			return res.apiError(403, 'not allowed', 'You can not delete this resource');
```
JedWatson: Curious about the language change here... isn't it more self-explanatory if the error message explicitly states why you can't delete the resource?

fidelisrafael: Hi! This is because the main implementation only allows user do not delete yourself, but I'm thinking maybe this can be expanded sometimes to allow other kind of ACL, so I changed to a more generic message. (In the client application this is not the error message that appear for user)

JedWatson: I think we're better off leaving it as it was until we expand the functionality - at which point we can add something more descriptive based on the condition that failed. While it's not a message that surfaces in the UI, generic error messages make debugging much more difficult.

JedWatson:
This is great @fidelisrafael, except I'd prefer the error messages stay as-is until there's a _more_ descriptive one we can send back. If you can change them back (or convince me otherwise ;) then this is good to merge :+1:

admin/server/api/list/delete.js
fidelisrafael: @JedWatson You're right, I don't want to convince you cause I agree with your point, I will change back to the original message and let the change happens when this update really ocurrs.

fidelisrafael: @JedWatson Changed it! I just keep the last part of first message(mentioning the `id`) to make it more specific, I mean, as you sad, this kind of thing really happens when debugging!

JedWatson:
Looks great @fidelisrafael, thanks!

PR #2950, opened 2016-05-26
Fix typo in RelationshipField (fixes #2855)

### LINKED ISSUE #2855: Option createInline: true breaks admin ui
When the option `createInline: true` is setted on the relationship field the page doesn't loads, I see only the loading progress.
The console writes error with `this._itemsCache`. I tried to find the error, but I found only one wierd thing:
*fields/types/relationship/RelationshipField.js*
```
cacheItem (item) {
		item.href = Keystone.adminPath + '/' + this.props.refList.path + '/' + item.id;
		this._itemsCache[item.id] = item;
	},
```
In one of iteration `this._itemsCache` is `undefined`, but I can't find why it disappears.

geloescht:
## Description of changes

This fixes a small typo that imported the wrong object from the list store in RelationshipField

## Related issues (if any)

#2855 

## Testing

@jstockwin I moved the test to misc and renamed it to 'Should be able to create an inline relationship'. Maybe you can extend it to test the + button and the create form. Either you merge it and fix it in master or I will try to give you commit rights to this branch of mine.

### DIFF
--- a/fields/types/relationship/RelationshipField.js
+++ b/fields/types/relationship/RelationshipField.js
@@ -1,6 +1,6 @@
 import async from 'async';
 import Field from '../Field';
-import listsByKey from '../../../admin/client/utils/lists';
+import { listsByKey } from '../../../admin/client/utils/lists';
 import React from 'react';
 import Select from 'react-select';
 import xhr from 'xhr';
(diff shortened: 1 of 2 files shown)


jstockwin:
@geloescht LGTM Re: test file movement. Just merge this with a TODO, and keep that as the standard practise when bugs are fixed. 

Then once we get quite a few in the misc folder they can all be re-ordered and added to, eventually making more complete tests.

JedWatson:
Great!

PR #2964, opened 2016-05-28
Configurable codemirror bundle path

webteckie:
## Description of changes

Codemirror is a complicated piece of software with many configuration options.  As such, users should have a saying as to how they want to bundle Codemirror.  This change allows users to generate their own Codemirror bundles and configure the bundle's url path.  If a custom path is not specified the default keystone bundle will be used.  

**Usage**
To define your own bundle path add a configuration like the following example to your keystone initialization:
```
'codemirror url path': '/lib/codemirror'
```
Make sure that the configured path is directly accessible from the site's root url.  Also make sure to place both the `codemirror.css` and the custom bundle `codemirror-compressed.js` at the specified custom path.   Using the example custom path above, these should be accessible as follows:
```
http://www.mysite.com/lib/codemirror/codemirror.css
http://www.mysite.com/lib/codemirror/codemirror-compressed.js
```

## Related issues (if any)
#1979 (although this PR takes a global approach to the bundle configuration, which keeps things relatively simple).

## Testing


mxstbr:
I like it, clean, simple solution. :+1: LGTM

webteckie:
@mxstbr then it should be called `reviewed` or `reviewed-ready-for-merge` :-)

mxstbr:
Hmm, I think it's fine as-is really, since normally submitters of PRs can't set labels, and calling it `reviewed ready` seems like redundant information.

webteckie:
Sound good...I learned my lesson :-)

JedWatson:
Looks great @webteckie!

Issue #2561, opened 2016-04-01
Inherited Users can't log in


autoboxer:
In keystone.js, there is a single model listed under 'user model' which can be used to sign in.  In order to get multiple user types on a site, creating a model that inherits from the one listed under 'user model' seems to be the only way, however, there's an issue when inherited users attempt to log in.

When any user attempts to log in the following happens:
1.  there's a call to doSignin(...).  A few checks pass and there's an attempt to get a user of the type 'user model' specified in keystone.js.
2.  There's a call to user._.password.compare(...) on the user object that returns from step 1, which is defined in PasswordType.js.  The compare function runs 'bcrypt.compare(candidate, item.get(this.path), callback);', where candidate is the unencrypted password, and item.get(this.path) is the encrypted password.

This is where we see the difference, bcrypt.compare(..) on the model listed under 'user model' in keystone.js is successful, but bcrypt.compare(...) on the model inheriting from the model listed under 'user model' is not.

It's the same behavior every time too.  When the 60 character password hash created by bcrypt and stored in the user object is compared with the unencrypted password based on bcrypt's logic, they don't match.  There's some complicated behavior at play regarding a salt that is created from the first 30 characters of the encrypted hash in order to process the unencrypted password and make the match.  If a non-inherited model is listed in keystone.js under 'user model', logging in works just fine.

Example model (this would be the one listed in keystone.js under 'user model':

```
var keystone = require('keystone'),
    Types = keystone.Field.Types;

// Create model
var User = new keystone.List('User', {
    track: true,
    map: { name: 'name.last' },
    defaultSort: 'name.last'
});

// Create fields
User.add('General Information', {

    name: {
        first: { type: Types.Text, label: 'first name', required: true, index: true, initial: true },
        last: { type: Types.Text, label: 'last name', required: true, index: true, initial: true }
    },

    password: { type: Types.Password, label: 'password', required: true, initial: true },
    avatar: { type: Types.CloudinaryImage, label: 'avatar', folder: 'users/admin', autoCleanup: true }

    email: { type: Types.Email, label: 'email address', unique: true, required: true, index: true, initial: true },

});

// Provide access to Keystone (simplified for the test)
User.schema.virtual('canAccessKeystone').get(function() {
    'use strict';
    return true;
});

// Define default columns in the admin interface and register the model
User.defaultColumns = 'name.first, name.last, email';
User.register();
```

Example model which inherits from the one listed in keystone.js under 'user model':

```
[...]

PR #2972, opened 2016-05-29
#2561 | Inherited Users can't log in

autoboxer:
## Description of changes

fix an issue where a model inheriting from the model allowed to sign in to the admin UI wasn't able to log in.  This was caused by inherited models being saved twice (Mongoose functionality I believe) which caused the password hash to be re-encrypted and made unusable.

## Related issues (if any)

#2561 - Inherited users can't log in

## Testing

No unit tests were written as the change is on the v0.3.x branch.

### DIFF
--- a/fields/types/password/PasswordType.js
+++ b/fields/types/password/PasswordType.js
@@ -59,25 +59,23 @@ password.prototype.addToSchema = function() {
 			return next();
 		}
 		var item = this;
-		if(!item.isEncrypted) {
-            bcrypt.genSalt(field.workFactor, function(err, salt) {
+        bcrypt.genSalt(field.workFactor, function(err, salt) {
+            if (err) {
+                return next(err);
+            }
+            bcrypt.hash(item.get(field.path), salt, function () {}, function(err, hash) {
                 if (err) {
                     return next(err);
                 }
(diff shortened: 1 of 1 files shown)


JedWatson:
@autoboxer there's a much simpler fix for this; we're already tracking whether or not the password needs to be hashed (see the setter), the flag just needs to be reset when the hashing process is run.

I've fixed it in 0.4 (see the commit I just tagged) - should be the same for 0.3, thought I'd leave it for you to update this PR with the new version.

I also haven't had a chance to test this locally so let me know if you find any issues with it!

autoboxer:
I tested it locally and a similar fix seems to work perfectly.  The only difference is 'this' loses context in the anonymouse function so I reference it with item instead.  Creating a PR now.

JedWatson:
Nice work @autoboxer, that's great. Mind cleaning up the whitespace though? should be tabs, and the spaces are misaligned on the new lines.

autoboxer:
I'm not sure why GitHub is showing it that way in the commit diff breakdown, but the formatting is already in tabs and aligned.  Here's a screenshot of the current state:

fix-screenshot

JedWatson:
That's really weird. The code review tab even has one of the comments misaligned... no worries, if it comes through in the merge I'll just clean it up.

Thanks!

PR #2981, opened 2016-06-01
support pre dynamic routing

gemscng:

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,7 +24,7 @@ var moduleRoot = (function (_rootPath) {
  * @api public
  */
 var Keystone = function () {
-	grappling.mixin(this).allowHooks('pre:static', 'pre:bodyparser', 'pre:session', 'pre:routes', 'pre:render', 'updates', 'signout', 'signin', 'pre:logger');
+	grappling.mixin(this).allowHooks('pre:static', 'pre:dynamic', 'pre:bodyparser', 'pre:session', 'pre:routes', 'pre:render', 'updates', 'signout', 'signin', 'pre:logger');
 	this.lists = {};
 	this.paths = {};
 	this._options = {
--- a/server/createApp.js
+++ b/server/createApp.js
@@ -72,6 +72,14 @@ module.exports = function createApp (keystone, express) {
 		keystone.callHook('pre:logger', req, res, next);
 	});
 
+	// Pre dynamic config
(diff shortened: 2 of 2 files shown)


gemscng:
By support pre dynamic routing, developer can route the page to their own reactjs component, which should let developer have the power to customize adminui

gemscng:
Any comment about this Pull Request? @JedWatson @mxstbr

mxstbr:
Sorry, I honestly have no idea. I think @JedWatson might have something planned for this?

webteckie:
@gemscng  can you please give a usage sample for this feature?  Think of what a user would want to read on a guide on how to use this feature.

gemscng:
I added this because we need to add admin page in Admin UI which is not a List. Like we add some report page which is not a list. Those report pages need a permission check.

gemscng:
@webteckie I add this to add our own page in admin ui which is not a list. For example, we add a report page to report usage of our api. 
When the report route go through createDynamicRoute in keystone core, it will return following error.
`
ReferenceError: Unknown keystone list "report"
` 

By having pre dynamic route middleware, we can then simply route to our own front end component without going through dynamic route of keystone.

JedWatson:
Sorry for the delay @gemscng - I keep thinking there is already a way to do this, just haven't had a chance to dig into it yet :/

JedWatson:
I've added a similar hook in, called `pre:admin` which I think is clearer.

It doesn't evaluate with the `app` object, for consistency - if we want to be able to insert custom middleware (not a hook) before the admin UI routes are handled, I think it's better to add an option for that (since this is how other custom middleware is implemented)

gemscng:
Thanks @JedWatson   This is a better implementation.

gemscng:
@JedWatson Although this is closed, i want to check if we should also add the configuration which will also call  

``
keystone.get('pre:admin')(app);
``

gemscng:
@JedWatson What do you think?

spesalvi:
Is there any documentation for 'pre:admin' hook.

When I hook following middleware, it is called for all routes. Not just for admin routes. Is it the expected behavior? 

`keystone.pre('admin', function(req, res, next) {
    let admin_path = keystone.get('admin path');
    logger.debug('admin middleware. path requested %s. admin path %s', req.path, admin_path);
    //res.redirect('/');
    next();
});`

Issue #2940, opened 2016-05-25
Relationship links are broken


webteckie:
### Expected behavior
Clicking a relationship link should navigate to the referenced relationship.

### Actual behavior
In the following image notice the link to the `Post Widget` but then notice the requests issued to the server.  It is doing a GET on the current list with the relationship definition but with the _id of the referenced item. 
![image](https://cloud.githubusercontent.com/assets/5872515/15524199/32b708ca-21ee-11e6-96e2-84f0d0e68204.png)

mxstbr:
That's definitely a bug I introduced with the admin ui overhaul, my bad! Good catch!

jstockwin:
@webteckie Just tried this on the relationship field on the e2e server. Added a new relationship and choose "e2e member". Went back to the relationship list view and clicked on "e2e member" in the "field A" column and it took me to e2e member's user item view, which is what is supposed to happen. 

What am I missing to recreate this?

pravk:
This is the same issue I faced and reported few days back.
https://github.com/keystonejs/keystone-demo/issues/51

webteckie:
This is reproducible by running the following e2e test from the keystone root directory:

`node test/e2e/server.js --env default --config ./test/e2e/adminUI/nightwatch.json --test test/e2e/adminUI/tests/group999FixMe/2940.js`

webteckie:
Closing ... PR #2986 fixed this.

export-mike:
Tests updated in the nightmare testing branch: https://github.com/keystonejs/keystone/commit/8bd457b25eae2e85503e793e3ae696e41b587318#diff-4dba9b0fc99b66b3ee6eb30ea2c37f60

PR #2986, opened 2016-06-02
Set next list upon transitioning to another item

webteckie:
## Description of changes

Sets the next list upon a list item transition happening within an item as when there are relationship links.  This ensures that the correct item will be loaded.

## Related issues (if any)

#2940

## Testing


JedWatson:
Looks good @webteckie!

PR #2989, opened 2016-06-02
Fix bug while checking for allowed S3 filetypes

### LINKED ISSUE #2966: _.contains is not a function on S3FileType
What the title says. I saw a merge request to fix this on Cloudinary File Types by replacing .contains with .includes. Can you guys do it for S3Files? S3FileType.js line 341 :D

I know you guys got a lot on your plate, but S3Files have been forsaken for a bit... share the love

TSFoster:
lodash doesn't offer a `contains` function (underscore.js does, maybe that's why it was mistakenly used). I've changed the code to make the check without the use of a library.

* Fixes #2966

### DIFF
--- a/fields/types/s3file/S3FileType.js
+++ b/fields/types/s3file/S3FileType.js
@@ -338,7 +338,7 @@ s3file.prototype.uploadFile = function (item, file, update, callback) {
 		update = false;
 	}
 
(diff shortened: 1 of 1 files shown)


JedWatson:
Good catch @TSFoster, thanks for the fix!

PR #3015, opened 2016-06-09
preview for the Attachments (small img + link + styling)

chiss2906:


mxstbr:
Awesome, thanks so much @chiss2906!

Would you mind rebasing onto the current `master` branch? Your version is a bit out of date…

fields/types/localfiles/LocalFilesField.js
```diff
@@ -37,9 +37,8 @@ var LocalFilesFieldItem = React.createClass({
 	},
 
 	render () {
-		let { filename } = this.props;
-		let ext = filename.split('.').pop();
-
+		const { filename, originalname, itemId, listKey } = this.props;
+		const isPicture = originalname && originalname.match(/\.(jpeg|jpg|gif|png)$/) != null;
```
geloescht: This should probably based on filetype instead of originalname. I'd test filetype first, and if it's `application/octet-stream`, test originalname.

geloescht:
One nitpick, other than that would love to see this merged.

fields/types/localfiles/LocalFilesField.js
wmertens: Yes, that sounds like a good improvement for after this is merged ;)

fields/types/localfiles/LocalFilesField.js
```diff
@@ -52,13 +51,22 @@ var LocalFilesFieldItem = React.createClass({
...
 		return (
 			<FormField>
-				<img key="file-type-icon" className="file-icon" src={Keystone.adminPath + '/images/icons/32/' + iconName + '.png'} />
-				<FormInput key="file-name" noedit className="field-type-localfiles__filename">
-					{filename}
-					{this.props.size ? ' (' + bytes(this.props.size) + ')' : null}
-				</FormInput>
-				{note}
-				{this.renderActionButton()}
+				<div style={{display: 'flex'}}>
```
wmertens: note that this will need prefixing so it should move to the css

fields/types/localfiles/LocalFilesField.js
```diff
@@ -52,13 +51,22 @@ var LocalFilesFieldItem = React.createClass({
...
+					{isPicture &&
+						<div style={{display: 'flex', width: 150, marginRight: 10}}>
+							<img style={{width: '100%', height: '100%'}} key="file-type-icon" className="file-icon" src={`/files/${listKey}/${itemId}/${filename}/${originalname}`}/>
+						</div>
+					}
+					<div style={{display: 'flex', flexDirection: 'column', justifyContent: 'space-between', alignItems: 'flex-start', minHeight: 100, width: '100%'}}>
+						<FormInput key="file-name" noedit className="field-type-localfiles__filename">
+							<a href={`/files/${listKey}/${itemId}/${filename}/${originalname}`}>{originalname}</a>
```
wmertens: and of course this either needs to be a generic file download endpoint in Keystone (with access control) or a configurable url (via a function with the above variables as attributes)

geloescht: On the server, there is the property .href for that purpose. Does it get passed to the admin UI as well?

wmertens:
Hmm deleted this by accident, but in fact we made some changes and we'll do this again on the 0.4 branch

PR #3021, opened 2016-06-10
Fixes problem with cloudinary.url formats

danielmahon:
## Description of changes

Moves Cloudinary url format to ONLY be set in the options object. This solves a problem where the return of `cloudinary.url()` will append the `options.format` to the default extension instead of replacing it. I think this is a problem with cloudinary itself, but this should fix for the time being and should be future proof even if they fix the problem. See https://github.com/cloudinary/cloudinary_npm/issues/103

## Related issues (if any)

https://github.com/cloudinary/cloudinary_npm/issues/103

## Testing

- [?] Please confirm `npm run test-all` ran successfully.

Tests fail: `Uncaught Error: connect ECONNREFUSED 127.0.0.1:27017`
But I don't think it's related to this simple change.

### DIFF
--- a/fields/types/cloudinaryimage/CloudinaryImageType.js
+++ b/fields/types/cloudinaryimage/CloudinaryImageType.js
@@ -159,7 +159,9 @@ cloudinaryimage.prototype.addToSchema = function () {
 
 		options.version = item.get(paths.version);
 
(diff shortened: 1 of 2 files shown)


JedWatson:
Looks good @danielmahon, thanks!

jstockwin:
@danielmahon FYI that error looks as though you don't have mongodb running, at least not on "127.0.0.1:27017".

PR #3027, opened 2016-06-11
Fix date and datearray e2e tests

jstockwin:
## Description of changes

cc: @webteckie. This is a solution to our problems about making the datearray popup go away so we can click save. It's not exactly a neat solution, but it works. Let me know what you think. 

## Related issues (if any)

Should fix travis.

## Testing


jstockwin:
These tests are going to fail running the `createInlineRelationshipTest` test. 
I've moved this test into `FixMe` in master, and created an issue about it. 

Given the above have been done, I'm fairly confident that once this has been merged, travis will pass on our master branch again.

webteckie:
LGTM

Issue #3041, opened 2016-06-15
Implement error states in the Admin UI


JedWatson:
We need to improve our handling of various different error states in the Admin UI.

The ones I can think of are:

## Common

- [ ] Server error - if we get an error (that's not in the list of expected cases below) from the API, it should display an error message with a button to resend the request (not poll every few seconds like we currently do)

## List View

- [ ] Invalid List - should display that the `listKey` (from the URL) doesn't match any registered lists, and have a big button to go back to the home screen.
- [ ] Invalid Columns - if you've specified columns that can't be expanded, it should display a warning and a button that resets the columns to default
- [ ] Invalid Filters - if you've specified filters that can't be expanded, it should display a warning and a button that resets the filters

## Items View

- [ ] Invalid List - should display that the `listKey` (from the URL) doesn't match any registered lists, and have a big button to go back to the home screen.
- [ ] Invalid Item ID (404) - should display the 404 and have a link to go back to the List view. We could improve on that by having a search field as well, that submits back to the List view
- [ ] Invalid Relationships Config - I'm not sure where this is at, we had a message that would be displayed below the Edit Form (back when we were rendering Jade templates) that displayed configuration issues with Relationships that prevented them from being displayed. We should look at re-implementing that if it's been lost, as relationships are way too easy to misconfigure.

mxstbr:
Agreed, currently all error states are the same which isn't perfect.

JedWatson:
Ref #2846 and #3013

mxstbr:
For invalid list, we throw a reference error in [lib/core/list.js:7](https://github.com/keystonejs/keystone/blob/c9a244dd2af4599186042a77fd26f621c9012b7a/lib/core/list.js#L7) when a list doesn't exist, which gets returned from the server as a 500.

Should we change that behavior to render the frontend and show a custom error?

PR #3059, opened 2016-06-20
Implement item not found error handling

mxstbr:
**Description of changes**

Shows an error when users go to the URL of an item that doesn't exist.

Probably needs more styling, @jossmac could you look into this? We should probably have a general error display that works for all sorts of messages.

**Related issues:** #3041

**Testing**

Getting `Error retrieving a new session from the selenium server` when running `npm run test-all`, that's probably environment related though. (Updated to the macOS Sierra beta, now my laptop barely works)

### DIFF
--- a/admin/client/App/screens/Item/actions.js
+++ b/admin/client/App/screens/Item/actions.js
@@ -5,7 +5,6 @@ import {
 	DATA_LOADING_ERROR,
 	DELETE_ITEM,
 } from './constants';
-import { NETWORK_ERROR_RETRY_DELAY } from '../../../constants';
 
 /**
  * Select an item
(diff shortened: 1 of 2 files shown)


wmertens:
So now it doesn't try to re-fetch the data if the error was something else than 404?

mxstbr:
Nope, that's not happening at the moment. I'm not 100% certain which error cases exist to be honest, would be great if @JedWatson has a list somewhere maybe?

mxstbr:
Accounting for network errors now too, prompting the user to refresh the page.

For some reason the error I get back from the request only says "Internal XMLHttpRequest error" and I can't get any more information out. (e.g. what kind of error was it) Any ideas what that could be about?

JedWatson:
Needs some UX work but I think this is good to merge & take forward in master...

I think actually having a "retry" button would be good, rather than just advising to refresh.

@mxstbr not sure about why you're not getting more info, let's figure out how to debug in slack?

mxstbr:
> I think actually having a "retry" button would be good, rather than just advising to refresh.

Agreed, but the thing is that we don't know what we want to retry since we don't get any information from the error.

Issue #2319, opened 2016-02-23
[0.4] multiple relationship display confusing


webteckie:
If I have a list that defines multiple relationships as follows:
```
MyList.relationship({ ref: 'RefList', path: 'listPath1', refPath: 'refPath1' });
MyList.relationship({ ref: 'RefList', path: 'listPath2', refPath: 'refPath2' });
MyList.relationship({ ref: 'RefList', path: 'listPath3', refPath: 'refPath3' });
```
 then the Admin UI shows it like this:
```
Relationships

Ref List

No related Ref List

Ref List

No related Ref List

Ref List

Name	Foo	      Bar
name1	foo1	  bar1
```

No where is the `path` defined in the relationship definition used, instead the referenced relationship name (`ref`) is used, which makes looking at the display confusing since it's hard to tell which is which.

creynders:
@webteckie doesn't the `label` field work in v0.4? E.g.

```js
MyList.relationship({ 
  ref: 'RefList', 
  path: 'listPath1', 
  refPath: 'refPath1',
  label: 'My super awesome ref list'
 });
```

webteckie:
@creynders I will check it out and report back on outcome.

webteckie:
@creynders I looked into this.  You cannot define a label in a relationship, or at least, the label doesn't show in the Admin UI that I can see.  I think the issue here is that the Admin UI is supposed to be using the `path` field of the relationship to clarify the display but it isn't:

If you have:
```
var TwoWayRelationshipA = new keystone.List('TwoWayRelationshipA', {});

TwoWayRelationshipA.add({
    name: { type: String, required: true },
    relWithCOfTypeA: {
        type: Types.Relationship,
        ref: 'TwoWayRelationshipC',
        many: true,
        label: "Rel C Relationships",
        filters: {type: 'A'}
    },
    relWithCOfTypeB: {
        type: Types.Relationship,
        ref: 'TwoWayRelationshipC',
        many: true,
        label: "Rel C Relationships",
        filters: {type: 'B'}
    }
});

TwoWayRelationshipA.defaultColumns = 'name';
TwoWayRelationshipA.register();
```
```
var TwoWayRelationshipC = new keystone.List('TwoWayRelationshipC', {});

TwoWayRelationshipC.add({
    name: { type: String, required: true },
    type: { type: Types.Select, options: 'A, B', default: 'A'}
});

[...]

PR #3063, opened 2016-06-21
Better relationship display.

webteckie:
## Description of changes

This change hopes to provide a better user experience when browsing the relationship display for items.  As currently implemented the updated display looks as follows:
image

## Related issues (if any)

#2319 

## Testing

### DIFF
--- a/admin/client/App/screens/Item/components/RelatedItemsList.js
+++ b/admin/client/App/screens/Item/components/RelatedItemsList.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Columns } from 'FieldTypes';
 import { Alert, Spinner } from 'elemental';
+import { titlecase } from '../../../../utils/string';
 
 const RelatedItemsList = React.createClass({
 	propTypes: {
@@ -22,7 +23,7 @@ const RelatedItemsList = React.createClass({
 	getColumns () {
 		const { relationship, refList } = this.props;
 		const columns = refList.expandColumns(refList.defaultColumns);
-		return columns.filter(i => i.path !== relationship.refPath);
+		return columns.filter(i => i.path === refList.nameField.path || i.path === relationship.refPath);
 	},
 	loadItems () {
 		const { refList, relatedItemId, relationship } = this.props;
@@ -62,13 +63,13 @@ const RelatedItemsList = React.createClass({
 	},
 	renderTableCols () {
 		const cols = this.state.columns.map((col) => <col width={col.width} key={col.path} />);
-		return <colgroup>{cols}</colgroup>;
+		return <colgroup>{[<col key="Relationship" />].concat([<col key="Parent" />]).concat(cols)}</colgroup>;
 	},
 	renderTableHeaders () {
 		const cells = this.state.columns.map((col) => {
 			return <th key={col.path}>{col.label}</th>;
 		});
-		return <thead><tr>{cells}</tr></thead>;
+		return <thead><tr>{[<th key="Relationship">Relationship</th>].concat([<th key="Parent">Parent</th>]).concat(cells)}</tr></thead>;
 	},
 	renderTableRow (item) {
 		const cells = this.state.columns.map((col, i) => {
@@ -76,16 +77,17 @@ const RelatedItemsList = React.createClass({
 			const linkTo = !i ? `${Keystone.adminPath}/${this.props.refList.path}/${item.id}` : undefined;
 			return <ColumnType key={col.path} list={this.props.refList} col={col} data={item} linkTo={linkTo} />;
 		});
-		return <tr key={'i' + item.id}>{cells}</tr>;
+		const listHref = `${Keystone.adminPath}/${this.props.refList.path}`;
+		const relationshipCol = [<td key={'Relationship' + item.id}>{this.props.relationship.label || titlecase(this.props.relationship.path)}</td>];
+		const parentCol = [<td key={'Parent' + item.id} className="Relationship__link"><a href={listHref}>{this.props.refList.label}</a></td>];
+		return <tr key={'i' + item.id}>{relationshipCol.concat(parentCol).concat(cells)}</tr>;
 	},
 	render () {
 		if (this.state.err) {
 			return <div className="Relationship">{this.state.err}</div>;
 		}
-		const listHref = `${Keystone.adminPath}/${this.props.refList.path}`;
 		return (
 			<div className="Relationship">
-				<h3 className="Relationship__link"><a href={listHref}>{this.props.refList.label}</a></h3>
 				{this.state.items ? this.renderItems() : <Spinner size="sm" />}
 			</div>
 		);
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -47,7 +47,7 @@ update the test suite so that any broken tests pass again.  You can run any of t
 from keystone's root directory:
 
     Pre-requisites:
-        - Make sure that you have Firefox(or Chrome) installed.  Firefox is the default browser used.
+        - Make sure that you have Firefox(or Chrome) installed and in the system path.  Firefox is the default browser used.
           Using Chrome requires specifying a different --env parameter (see below).  For any tests below
           you may replace the "--env default" parameter with one of the following:
           
--- a/test/e2e/models/misc/SourceRelationship.js
+++ b/test/e2e/models/misc/SourceRelationship.js
@@ -12,9 +12,14 @@ SourceRelationship.add({
 		type: Types.Relationship, 
 		ref: 'TargetRelationship'
 	},
+	fieldB: { 
+		type: Types.Relationship, 
+		ref: 'TargetRelationship',
+		many: true
+	},
 });
 
 SourceRelationship.register();
-SourceRelationship.defaultColumns = 'name, fieldA';
+SourceRelationship.defaultColumns = 'name, fieldA, fieldB';
 
 module.exports = SourceRelationship;
--- a/test/e2e/models/misc/TargetRelationship.js
+++ b/test/e2e/models/misc/TargetRelationship.js
@@ -13,7 +13,13 @@ TargetRelationship.add({
 TargetRelationship.relationship({
 	ref: 'SourceRelationship',
 	refPath: 'fieldA',
-	path: 'sourceFieldA'
+	path: 'sourceFieldA',
+	label: 'Source Relationship A References'
+});
+TargetRelationship.relationship({
+	ref: 'SourceRelationship',
+	refPath: 'fieldB',
+	path: 'sourceFieldB'
 });
 
 TargetRelationship.register();
--- a/test/e2e/server.js
+++ b/test/e2e/server.js
@@ -11,7 +11,7 @@ var child_process = require('child_process');
 var path = require('path');
 
 var dbName = '/e2e' + (process.env.KEYSTONEJS_PORT || 3000);
-var mongoUri = 'mongodb://' + (process.env.KEYSTONEJS_HOST || 'localhost') + dbName;
+var mongoUri = 'mongodb://' + (process.env.KEYSTONEJS_HOST || '127.0.0.1') + dbName;
 
 var selenium = null;
 
@@ -19,7 +19,7 @@ keystone.init({
 	'name': 'e2e',
 	'brand': 'e2e',
 
-	'host': process.env.KEYSTONEJS_HOST || 'localhost',
+	'host': process.env.KEYSTONEJS_HOST || '127.0.0.1',
 	'port': process.env.KEYSTONEJS_PORT || 3000,
 
 	'mongo': mongoUri,
@@ -150,6 +150,7 @@ function runSeleniumInBackground (done) {
 	{
 	  var line = buffer.toString();
 	  if(line.search(/Selenium Server is up and running/g) != -1) {
+			console.log([moment().format('HH:mm:ss:SSS')] + ' e2e: selenium server running in background...');
 			running = true;
 			done();
 	  }
@@ -157,6 +158,7 @@ function runSeleniumInBackground (done) {
 
 	selenium.on('close', function (code) {
 		if(!running) {
+			console.log([moment().format('HH:mm:ss:SSS')] + ' e2e: selenium server run error...');
 			done(new Error('Selenium exited with error code ' + code));
 		}
 	});
@@ -196,7 +198,7 @@ function runKeystone(cb) {
 function start() {
 	var runTests = process.argv.indexOf('--notest') === -1;
 	var dropDB = process.argv.indexOf('--nodrop') === -1;
-	var runSelenium = !(process.argv.indexOf('--selenium-in-background') === -1);
+	var runSelenium = process.argv.indexOf('--selenium-in-background') !== -1;
 
 	async.series([
 


mxstbr:
LGTM, nice work!

webteckie:
@mxstbr although I think it looks better I also think

webteckie:
Bad key press in cell phone closed this unexpectedly...reopened :=)  @mxstbr^

webteckie:
@JedWatson @mxstbr this is now looking like this:
image

As you can see all relationships are now under a single table and the column names are more explicit to match the relationship configuration in a list.

admin/client/App/screens/Item/components/RelatedItemsList.js
```diff
@@ -22,7 +23,9 @@ const RelatedItemsList = React.createClass({
 	getColumns () {
 		const { relationship, refList } = this.props;
 		const columns = refList.expandColumns(refList.defaultColumns);
-		return columns.filter(i => i.path !== relationship.refPath);
+		const refListNameColumn = columns.filter(i => i.path === refList.nameField.path);
+		const refListRelationshipFieldColumn = columns.filter(i => i.path === relationship.refPath);
```
mxstbr: For clarity, I'd prefer those two to be:

```JS
columns.filter((column) => column.something === somethingElse);
```

(am not too fussed about the parentheses, but why call the variable inside the filter `i` when it's a `column`?)

admin/client/App/screens/Item/components/RelatedItemsList.js
```diff
@@ -47,48 +50,69 @@ const RelatedItemsList = React.createClass({
...
-			return <th key={col.path}>{col.label}</th>;
-		});
-		return <thead><tr>{cells}</tr></thead>;
+	renderParentColumn (item) {
+		const listHref = `${Keystone.adminPath}/${this.props.refList.path}`;
+		return <td key={'Parent' + item.id} className="Relationship__link"><a href={listHref}>{this.props.refList.label}</a></td>;
+	},
+	renderNameColumn (item) {
+		const column = this.state.columns[0];
+		const ColumnType = Columns[column.type] || Columns.__unrecognised__;
```
mxstbr: What is `Columns.__unrecognised__`, where does it come from? Keystone-internal thing?

admin/client/App/screens/Item/components/RelatedItemsList.js
```diff
@@ -47,48 +50,69 @@ const RelatedItemsList = React.createClass({
...
-		});
-		return <tr key={'i' + item.id}>{cells}</tr>;
+		return (
+			<tr key={'i' + item.id}>{[
+				this.renderRelationshipColumn(item),
+				this.renderParentColumn(item),
+				this.renderNameColumn(item),
+				this.renderFieldColumn(item),
+				this.renderValueColumn(item),
+			]}</tr>
```
mxstbr: A key of `i` could be non-unique, thus lead to confusion and unnecessary rerenders from React. Maybe use `'table-row-item-' + item.id` or something like that?

admin/client/App/screens/Item/index.js
```diff
@@ -70,6 +70,24 @@ var ItemView = React.createClass({
 			createIsOpen: visible,
 		});
 	},
+	renderTableCols () {
+		return (<col key="Relationship" />,
```
mxstbr: nit: Enter after `return (`

admin/client/App/screens/Item/index.js
```diff
@@ -70,6 +70,24 @@ var ItemView = React.createClass({
...
+			<col key="Parent" />,
+			<col key="Item" />,
+			<col key="Field" />,
+			<col key="Values" />
+		);
+	},
+	renderRelationshipTableHeaders () {
+		return (<thead><tr>{[
```
mxstbr: nit: Enter after `return (`, proper indentation

mxstbr:
Did a proper code review (should've done previously, my bad!), left a few comments and nitpicks but the new table looks great! :+1: Good work!

admin/client/App/screens/Item/components/RelatedItemsList.js
webteckie: was just following the consistency of whomever did it first :-)  will update

admin/client/App/screens/Item/components/RelatedItemsList.js
webteckie: don't know...it was already there :-)

admin/client/App/screens/Item/index.js
webteckie: Yeah, had it like that but lint just wouldn't take it...will try again

admin/client/App/screens/Item/index.js
webteckie: ditto

webteckie:
@mxstbr thanks for the code review...this is still in-progress as need to still handle nameless relationships.  btw, not sure why my other github ID (epihacker) is being used for the commits...very confusing :-)

webteckie:
@JedWatson @mxstbr I now deem this change set complete.  I tested as many use cases of relationships as I could think of to make sure all displays fine.  The display is now flexible in that it handles named/unnamed relationships, labeled/unlabeled relationships, and different relationship cardinalities.  Again, the display also tries to reflect the relationship configuration to make it easier to reason about.

mxstbr:
LGTM from a code perspective, ping @JedWatson who has a bigger overview of the feature set here.

JedWatson:
@webteckie tbh I'm a bit confused about the problem this is solving, and what your solution actually does (vs. how I expected relationships to behave; agreed there are bugs and/or missing features in the current implementation that we need to address!)

In ca use-case with lots of related items coming in from a single relationship, I would think this update (as I understand it) would make it _harder_ to use, esp. for example in a scenario where pagination / etc is involved...

But I don't know that I understand what you've done properly. Can you describe it for me, maybe with some example List definitions to go along with the screenshots of how it now works?

webteckie:
@JedWatson to me the way things are now is a bit difficult to make reason about when relationship values exist as well as the display doesn't make sense when you have multiple relationships defined and no, yet, filled.  Obviously, this seems to be a subjective thing I guess since you seem to think the current display is adequate :-)  As far as showing examples, I think it's best that you pull it down and compare the existing and the proposed with different relationship configurations.

webteckie:
Latest display looks like this...the parent relationship is now always displayed even if no relationship items configured:

image

webteckie:
@JedWatson not sure pagination is a relevant issue here since the rows in the table always relate to the number of relationship fields configured in the active list.  You would never expect that to be a huge number of fields.  To see the actual full set of relationship values in the field you still need to go to the parent list (same as the previous display).  I tried to go back to take some snapshots of the previous display to compare and can't even do it.  For the exact same item above this is what I see:

image

Doesn't compare! :-)

admin/client/App/screens/Item/components/RelatedItemsList.js
wmertens: Actually, keys need only be unique in the current parent, so just using `item.id` is fine (as long as that is defined).

admin/client/App/screens/Item/components/RelatedItemsList.js
```diff
@@ -47,48 +53,68 @@ const RelatedItemsList = React.createClass({
...
+		);
 	},
 	render () {
+		let tbody = <tr><td><Spinner size="sm" /></td></tr>;
 		if (this.state.err) {
-			return <div className="Relationship">{this.state.err}</div>;
+			tbody = <tr><td>{this.state.err}</td></tr>;
+		} else if (this.state.items && this.state.items.results && this.state.items.results.length) {
```
wmertens: This would read a lot easier with
```
} else {
   const results = this.state.items && this.state.items.results
   if (results) {
      if (results.length) {...} else {...}
  }
```
also, making the spinner tbody only in a final else clause prevents unnecessary React.createElement calls.

admin/client/App/screens/Item/index.js
```diff
@@ -79,19 +99,25 @@ var ItemView = React.createClass({
...
-								list={this.props.currentList}
-								refList={refList}
-								relatedItemId={this.props.params.itemId}
-								relationship={relationship}
-							/>
-						);
-					})}
+					<div className="ItemList-wrapper">
+						<table cellPadding="0" cellSpacing="0" className="Table ItemList">
+							<colgroup>{[this.renderTableCols()]}</colgroup>
```
wmertens: Why not make colgroup part of renderTableCols, like thead in the headers call?

admin/client/App/screens/Item/index.js
```diff
@@ -79,19 +99,25 @@ var ItemView = React.createClass({
...
+							{this.renderRelationshipTableHeaders()}
+							{keys.map(key => {
+								const relationship = relationships[key];
+								const refList = listsByKey[relationship.ref];
+								return (
+									<RelatedItemsList
+										key={relationship.path}
+										list={this.props.currentList}
+										refList={refList}
+										relatedItemId={this.props.params.itemId}
```
wmertens: you could put this and this.props.currentList in a const at the top of the function to increase clarity and reduce repeated deferencing in the loop.

wmertens:
👍

admin/client/App/screens/Item/components/RelatedItemsList.js
```diff
@@ -96,25 +86,43 @@ const RelatedItemsList = React.createClass({
...
+				<td></td>
+			</tr>
+		);
+	},
+	renderError () {
+		return <tr><td>{this.state.err}</td></tr>;
+	},
+	renderRelationshipTableBody () {
+		const loadedItems = this.state.items;
+		const results = this.state.items && this.state.items.results;
```
wmertens: const results = loadedItems && loadedItems.results;

admin/client/App/screens/Item/components/RelatedItemsList.js
```diff
@@ -96,25 +86,43 @@ const RelatedItemsList = React.createClass({
...
-				<tr>
-					{this.renderRelationshipColumn({})}
-					{this.renderParentColumn({})}
-					<td>None</td>
-					<td></td>
-					<td></td>
-				</tr>
-			);
+			tbody = this.renderError();
+		} else if (loadedItems && results.length) {
```
wmertens: results && results.length is the same and more indicative of function. You could also nest this and the next if to show that this section is concerned with results

admin/client/App/screens/List/components/ItemsTable/ItemsTableRow.js
```diff
@@ -38,7 +38,7 @@ const ItemsRow = React.createClass({
 		});
 		// item fields
 		var cells = this.props.columns.map((col, i) => {
-			var ColumnType = Columns[col.type] || Columns.__unrecognised__;
+			var ColumnType = Columns[col.type] || Columns.text;
```
wmertens: @JedWatson I had a KS frontend crash due to `Columns.__unrecognised__`, so I'm happy with this line, but I presume that the idea was to actually define a column type `__unrecognised__` so that it would show in a specific way in the UI?

fields/types/text/TextColumn.js
```diff
@@ -17,7 +17,7 @@ var TextColumn = React.createClass({
 	render () {
 		const value = this.getValue();
 		const empty = !value && this.props.linkTo ? true : false;
-		const className = this.props.col.field.monospace ? 'ItemList__value--monospace' : undefined;
+		const className = this.props.col.field && this.props.col.field.monospace ? 'ItemList__value--monospace' : undefined;
```
wmertens: Had a crash here, too :) good to see this 👍

admin/client/App/screens/Item/components/RelatedItemsList.js
```diff
@@ -22,7 +23,12 @@ const RelatedItemsList = React.createClass({
 	getColumns () {
 		const { relationship, refList } = this.props;
 		const columns = refList.expandColumns(refList.defaultColumns);
-		return columns.filter(i => i.path !== relationship.refPath);
+		const refListNameColumn = columns.filter(column => column.path === refList.namePath || (column.path === 'id' && refList.namePath === '_id'));
+		const refListRelationshipFieldColumn = columns.filter(column => column.path === relationship.refPath);
+		return [refListNameColumn[0], refListRelationshipFieldColumn[0]];
```
wmertens: @webteckie, I'm seeing this second column as undefined, causing problems during render. Thoughts?

wmertens: So unless the referring field is in the default columns of the referring list, this crashes.

webteckie: @wmertens probably an issue in the original code as well?  I can look into this, but I'm hesitant to spend any more time on this if @JedWatson is not going to merge it :=)

wmertens:
@webteckie I fixed the column issue, split out the crash fixes, and rebased.

wmertens:
@webteckie now that I've had some time to play with it, I do think the information could be presented better. I would prefer a sub-heading per relationship, and then a list with the named links to the referrers, with a simple `- referringField: value` after it. That way, long names and long many-relationships can use the whole line.

Also, I would prefer if the value were only shown for many-relationships, otherwise you get a lot of repetition.

wmertens:
Pie in the sky: how about, when you click on a relationship, it folds open a quick view of the item, without editing fields? Kind of like the movie detail view in NetFlix

admin/client/App/screens/List/components/ItemsTable/ItemsTableRow.js
JedWatson: Yeah, there used to be an `__unrecognised__` type, which would warn that an implementation was missing. Looks like it was lost - I think we should add it back again, rather than quietly defaulting to `text`. Having a type without an implementation is an error case, and means something's gone wrong at a really low level.

fields/types/text/TextColumn.js
JedWatson: If we're having crashes here, it means something lower level has gone horribly wrong - this check shouldn't be necessary, we should be validating it higher up

wmertens: @JedWatson this is the crash you get when it applies the `text` column type instead of the `__unrecognized__` one, so it's indeed secondary and probably not be needed once `__unrecognized__` is implemented.

webteckie:
@wmertens 

> I would prefer a sub-heading per relationship, and then a list with the named links to the referrers, with a simple - referringField: value after it.

it seems like what you describe above is the way things are :-)  The reason for this PR is because that doesn't seem to work that well for me :-)

> how about, when you click on a relationship, it folds open a quick view of the item, without editing fields?

That could work, I guess.  Feel free :-)

admin/client/App/screens/List/components/ItemsTable/ItemsTableRow.js
webteckie: @JedWatson how would you handle cases where a list does not have a name mapping?  In that case,  the ID, which doesn't not have a representative field shows.  That's what the `text` default is doing here ;=)

JedWatson: just leaving a note here to clarify that this has been fixed in master

PR #3078, opened 2016-06-24
Repeated subexpressions in DomParser.js

kilida:
## Description of changes

`sibling.name === 'ul'` is repeated. Based on the comment on line 132, looks like this should be `ol`?

## Related issues (if any)

-

## Testing

Tested against travis


JedWatson:
This looks accurate, and I'm happy to merge, but has this been raised in TinyMCE? This code is a straightforward copy of their production build so we should make sure it's fixed upstream or we'll lose it next time TinyMCE is updated.

(alternatively if this has been fixed and our bundled version is lagging behind, we should fix that too!)

JedWatson:
p.s. Thanks @kilida :)

PR #3087, opened 2016-06-28
(e2e) Add geckodriver for various platforms and corresponding configuration

geloescht:
## Description of changes

This adds binaries from https://github.com/mozilla/geckodriver/releases like the ones we use for Chrome. They are needed for Firefox and Selenium to play nice beginning with FF47.
Some remaining questions, and the reason I am doing this in a PR: (@keystonejs/testing)
* What about licensing, do we need to add a note somewhere? Is there a note about the license of chromedriver already?
* Should we try and auto-detect OS and Browser versions in the future, now that it's all not as straightforward as `npm run test-e2e` anymore?


geloescht:
Running single tests works with FF47 now, but running multiple tests triggers this bug: https://github.com/mozilla/geckodriver/issues/58
It is hopefully going to be fixed soon in GeckoDriver 0.9.

geloescht:
We also need to fix e2e with selenium in background, because we are not passing cli_opts from the config to selenium. Unless, hopefully people from nightwatch are going to fix https://github.com/nightwatchjs/nightwatch/issues/470 soon.

webteckie:
Don't you need to also patch nightwatch-no-process.json with the same changes?

geloescht:
I updated geckodriver to v0.9 but I still can't run our testsuite with it. So keeping this on hold for now.

JedWatson:
On a related node, I'd really like to get these binaries out of Keystone's git repo. They're going to blow out the size of the history, forever. Maybe we can encapsulate them in a separate npm package and depend on that? (if they're not already available like that, surely we're not the only ones who have this issue?)

JedWatson:
https://www.npmjs.com/package/soprano-saxophone

morenoh149:
https://www.npmjs.com/package/chromedriver

geloescht:
Still failing due to Selenium bug, see here: https://github.com/mozilla/geckodriver/issues/58#issuecomment-230257288

JedWatson:
@jstockwin and @webteckie is this still relevant or should we close it?

jstockwin:
@JedWatson and @geloescht - As of #3351, I'd suggest this is no longer relevant. As Jed's above comment says, we don't actually want these drivers in the repo. 

@geloescht If you can find a package to do this, that'd be great - otherwise just update the readme with instructions on where to download these drivers from. I'd also like to get a package working for the chromedrivers, but it's nowhere near the top of my list now we're doing a full refactor (#3359).

PR #3100, opened 2016-07-01
More testing

mxstbr:
Ref: #3108

Two problems:

1. ~~In `Signin.js` we save the CSFR key and value _before the component is rendered_. Mocha parses the files before the test files, so that throws because `Keystone.csrf` is undefined since I can only set it in the test files. Any ideas how to circumvent that? Can we move that call to `componentWillMount` possibly @JedWatson? Why is that even there?~~ Fixed by @JedWatson 

2. In `CreateForm.js` we do `import { Fields } from 'FieldTypes'`, but the `FieldTypes` package doesn't exist in the testing environment. Any ideas how to get that?


JedWatson:
Let's figure out the answer to 2) first. I am looking into the stuff in CreatForm and it needs to be updated, I'm on that.

mxstbr:
@JedWatson 646516d73296ad7d0788f10b3d6d4827fa4381b4 wasn't actually the problem, the CSRF in the component itself is perfectly fine. These are the problematic lines: https://github.com/keystonejs/keystone/blob/a522f5711891ce0d4e183272f5a3bc968be5751d/admin/client/Signin/Signin.js#L15-L18

JedWatson:
@mxstbr perfectly fine but shouldn't be there :wink:

I've fixed the headers in the SignIn component as well.

mxstbr:
Awesome, thanks! Do you know how to fix `import { Fields } from 'FieldTypes'` in the test env?

mxstbr:
I've added (currently not working) tests for the `PrimaryNavigation` component. It turns out that it relies on `window.innerWidth` in `componentDidMount`, which isn't called when shallow rendering. Any ideas?

mxstbr:
The `App` component used to `import` `listsByPath`, which always failed because you cannot mock before any test runs with Mocha, which means `Keystone` was undefined.

By moving that to a dynamic require into the component (cdd9975ecc7a9ce926c95ac65f22979a931f93fd) we avoid this issue, but that doesn't seem like the cleanest solution…

admin/client/App/shared/FlashMessages.js
```diff
@@ -15,6 +15,8 @@
  */
 
 import React from 'react';
+import { isPlainObject } from 'lodash';
```
JedWatson: looks good @mxstbr but remember to `import _ from 'lodash'` then use `_.isPlainObject` so we don't confuse browserify and start including two copies of lodash in the client again...

mxstbr: Ah, thanks, done!

mxstbr:
I've renamed and changed the `test-react` command. Having a specific command for React components doesn't make any sense (what about actions?), it makes more sense to test area-specific, so we now have `test-admin` instead.

I've also gotten rid of `.test.jsx` and made it `.test.js`, found a bunch of linting errors because eslint only lints `.js` files.

mxstbr:
Let's merge this for now when CI passes, this is getting too big!

JedWatson:
@mxstbr

> Awesome, thanks! Do you know how to fix import { Fields } from 'FieldTypes' in the test env?

No, unfortunately. Have you had any ideas yet?

JedWatson:
@mxstbr I'm down with merging this before it becomes unwieldy as a PR :+1:

Nice work!

wmertens:
For the FieldTypes, the webpack branch I started actually builds that and it will be available for everyone that way.

For loading the field code on the fly, I just realized we can use a wrapper component that asynchronously requires the field code and renders it once loaded. That way, the entire keystone UI can be a single modular build which only requires a schema configuration.

PR #3129, opened 2016-07-07
Testing

mxstbr:
- Added more tests
- Refactored the List active reducer to be a lot better! (lodash is sooo good…)
- Refactored HomeView into reusable components, removing lots of duplicate logic

- [x] Fix tests for HomeView and new components
- [x] Move `getHeadingIconClass` into own, unit-tested util

**READY TO BE MERGED ASSUMING CI PASSES**

### DIFF
--- a/admin/client/App/screens/List/reducers/active.js
+++ b/admin/client/App/screens/List/reducers/active.js
@@ -49,30 +49,12 @@ function active (state = initialState, action) {
 				columns: action.columns,
 			});
 		case ADD_FILTER:
-			let existsIndex;
-			let filters;
-			let activeFilters = state.filters;
-			// Check if a filter for the passed path exists already, and save the
-			// index if it does
-			const filterExists = Object.keys(activeFilters).some((filter, index) => {
-				existsIndex = index;
-				return activeFilters[filter].field.path === action.filter.field.path;
-			});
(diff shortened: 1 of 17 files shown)


admin/client/App/screens/List/reducers/active.js
```diff
@@ -49,33 +49,15 @@ function active (state = initialState, action) {
...
-				filters,
+				// Override existing filter with field path,
+				// otherwise add to filters array
+				filters: _.unionWith([action.filter], state.filters, (stateFilter, actionFilter) => {
+					return stateFilter.field.path === actionFilter.field.path;
+				}),
 			});
 		case CLEAR_FILTER:
-			let newFilters = remove(state.filters, (filter) => {
+			let newFilters = _.filter(state.filters, (filter) => {
```
mxstbr: This was a bug, as lodash `remove()` mutates the array passed in. (in our case `state.filters`) `filter()` instead does the same thing but doesn't mutate, which is preferential!

(this was verified using the unit tests)

mxstbr:
I don't understand why the Push CI passes but the PR CI fails. All the tests work locally, so this is good to merge?

JedWatson:
Good to merge. Nice work!

PR #3131, opened 2016-07-07
Max characters validation added

dariasamo:
## Description of changes

We added max character validation to validateInput method and 'should invalidate string over max characters' test case

## Related issues (if any)
#126 

## Testing

### DIFF
--- a/fields/types/text/TextType.js
+++ b/fields/types/text/TextType.js
@@ -8,6 +8,7 @@ var utils = require('keystone-utils');
  * @api public
  */
 function text (list, path, options) {
+	this.options = options;
 	this._nativeType = String;
 	this._properties = ['monospace'];
 	this._underscoreMethods = ['crop'];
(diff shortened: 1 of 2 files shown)


mxstbr:
Your first Pull Request! :tada: :tada: :tada:

This looks brilliant, let's merge this!! *(our workflow: the reviewer, in this case me, marks it as `ready` so @JedWatson can take a last look before merging this into master)*

----

*What about adding a `min` option too? Maybe I want to require the string to be over a certain amount of characters. I'll add it to the potential todos!*

r3wt:
Awesome, this is a really great feature! Thank you @dariasamo

JedWatson:
:+1: Nice @dariasamo! Thanks!

devisscher:
This is great, thanks for the addition. 

I was wondering, is there an option to display a validation message which explains why the field is invalid?  

I used this: 

```
field: { type: String, max: x } 
```
And in the admin ui, it says 

``` 
Field is invalid 
```
Is it possible to add a custom validation message?

Also does this option work in the admin ui?

```
field: { type: String, displayChars: true }
```

Taken from: Issue #126

areknow:
it would be great to use 'displayChars: true'

VinayaSathyanarayana:
is this not in latest beta?

PR #3133, opened 2016-07-08
🔒 Add automated Let's Encrypt support 🎉

wmertens:
## Description of changes

Adds support for https://letsencrypt.org. All you need to do is enable it, set your domain name, and you will have https support!

Example configuration lines:
```
                'letsencrypt': (process.env.NODE_ENV === 'production'),
                'letsencrypt email': 'yours@foo.com',
                'letsencrypt domains': ['www.bar.com', 'bar.com'],
                'letsencrypt register': true,
                'letsencrypt agree TOS': true,
                'letsencrypt production': true,
```

## Testing

I didn't implement any tests for this. I suppose there could be a unit test that verifies it fails to accept registration without email and domain.

### DIFF
--- a/lib/core/start.js
+++ b/lib/core/start.js
@@ -56,11 +56,21 @@ function start (events) {
 		var ssl = keystone.get('ssl');
 		var unixSocket = keystone.get('unix socket');
 		var startupMessages = ['KeystoneJS Started:'];
+		var letsencrypt = keystone.get('letsencrypt');
 
 		async.parallel([
+			// Let's Encrypt
+			function (done) {
+				if (!letsencrypt || unixSocket) return done();
+				require('../../server/startLetsEncrypt')(keystone, app, function (err, msg) {
+					fireEvent('onHttpServerCreated');
+					startupMessages.push(msg);
+					done(err);
+				});
+			},
 			// HTTP Server
 			function (done) {
-				if (ssl === 'only' || unixSocket) return done();
+				if (ssl === 'only' || letsencrypt || unixSocket) return done();
 				require('../../server/startHTTPServer')(keystone, app, function (err, msg) {
 					fireEvent('onHttpServerCreated');
 					startupMessages.push(msg);
@@ -69,7 +79,7 @@ function start (events) {
 			},
 			// HTTPS Server
 			function (done) {
-				if (!ssl || unixSocket) return done();
+				if (!ssl || letsencrypt || unixSocket) return done();
 				require('../../server/startSecureServer')(keystone, app, function () {
 					fireEvent('onHttpsServerCreated');
 				}, function (err, msg) {
--- a/server/startLetsEncrypt.js
+++ b/server/startLetsEncrypt.js
@@ -1,0 +1,100 @@
+/**
+ * Configures and starts express server via Let's Encrypt.
+ *
+ * Events are fired during initialisation to allow customisation, including:
+ *   - onHttpServerCreated
+ *
+ * consumed by lib/core/start.js
+ *
+ * @api private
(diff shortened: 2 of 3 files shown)


lib/core/start.js
```diff
@@ -56,11 +56,21 @@ function start (events) {
 		var ssl = keystone.get('ssl');
 		var unixSocket = keystone.get('unix socket');
 		var startupMessages = ['KeystoneJS Started:'];
+		var lex = keystone.get('letsencrypt');
```
mxstbr: Could we rename this to something that signifies what it is? Maybe `letsencrypt`? :grin:

mxstbr:
This is awesome! Let's fix CI, I'm fine with the code but not very aware of that part of the code base, @JedWatson will have to take a deeper look.

server/startLetsEncrypt.js
```diff
@@ -0,0 +1,103 @@
...
+var http = require('http');
+var https;
+try {
+	// Use spdy if available
+	https = require('spdy');
+} catch (e) {
+	https = require('https');
+}
+var path = require('path');
+var LEX = require('letsencrypt-express');
```
JedWatson: ALL_CAPS implies constants, let's call this `letsencrypt`

server/startLetsEncrypt.js
```diff
@@ -0,0 +1,103 @@
...
+var lexDir = path.join(require('os').homedir(), 'letsencrypt', 'etc');
```
JedWatson: Path is more consistent with our naming conventions, let's go with `letsencryptPath`

server/startLetsEncrypt.js
```diff
@@ -0,0 +1,103 @@
...
+
+	if (!opt('production')) { LEX.testing(); }
+
+	var lex = LEX.create(options);
+
+	var message = keystone.get('name') + ' is ready on ';
+	var toStart = 1;
+	function ready (err) {
+		if (err) { callback(err); }
+		if (--toStart !== 0) { return; }
```
JedWatson: This seems kind of awkward, do we really need this count? Could this code be written more simply?

JedWatson:
Firstly, :+1: awesome!

Specific implementation feedback inline.

Regarding the options, I know you're following other patterns here but I think we're better off going with a configuration options object rather than lots of strings. I think I'd like to see

```
letsencrypt: (process.env.NODE_ENV === 'production') ? {
    email: 'yours@foo.com',
    domains: ['www.bar.com', 'bar.com'],
    register: true,
    tos: true,
    production: true,
} : false
```

wmertens:
That works for me, I hate spaces in object keys :)

server/startLetsEncrypt.js
wmertens: Actually now that I think about it, both servers are always required, at least that's my understanding of the protocol. So I could set this to 2? I need a barrier until both were started before calling the callback.

Alternatively, use `async`?

wmertens:
@JedWatson Another thing, perhaps `production` is not needed? I believe the testing setup is more for doing e2e testing of the letsencrypt libraries.

So I will drop that?

server/startLetsEncrypt.js
JedWatson: use `async`

wmertens:
@JedWatson @mxstbr how does this look?

mxstbr:
LGTM (assuming CI passes)

JedWatson:
Looks good to me too @wmertens! Thanks!

wmertens:
Ok, I'll add that erroring semi and merge :)

Issue #3075, opened 2016-06-24
Admin interface analytics


mxstbr:
For us maintainers it's really hard to judge which features of the admin interface people are using/not using.

It'd be nice to add opt-out analytics where we can collect (anonymized, obviously) information about what people are doing with the Keystone admin interface so we can adapt to what users are actually doing!

_Yes I realize this isn't a bug, but this wouldn't make sense to be on Product Pains since it's mostly maintainer-facing_

rturk:
Can we also use this thread for missing features?

mxstbr:
No, this issue is specifically about introducing something like Google Analytics to the admin interface. Please submit individual issues to Github or feature requests to Product Pains!

vkammerer:
Could this issue relate to the overall architecture of the project, and more specifically to the potential split between a "core" codebase and "plugins"?
What I mean by that is that if we had a "core" package, and then "plugin" packages, you would know which additional features are being used just by looking at the number of downloads / issues created / PR submitted for each individual plugin.
I feel that adding another feature to the project (this time, analytics) is not a good solution to help shipping v4... Instead, it should be the opposite, the codebase should be getting lighter.
It doesn't really make sense that there are so many issues related to Cloudinary and so on, when a lot of users will never use that service and are just waiting for a stable release of basic v4 (like me :) )
The example of babel and their strategy for going from v5 to v6 seems a good example, don't you think? They isolated a core and externalized a lot of plugins, and in the process had to design a clear API for each plugin to integrate with core. I would love to see that with Keystone as well.

mxstbr:
> They isolated a core and externalized a lot of plugins, and in the process had to design a clear API for each plugin to integrate with core.

That's what we're working towards and is one of the big reasons v4 is such a long update! (sorry about that…)
We are now using react, which will very easily allow us to split out and integrate new fields (like Cloudinary), which in combination with #3052 will make our life maintaining the core a lot easier.

We will definitely not make that happen for v4, since we do want to get that released asap, but it's very very high up on our roadmap!

> I feel that adding another feature to the project (this time, analytics) is not a good solution to help shipping v4...

This is not a feature per-se, as no user will ever see this. This is solely for us maintainers, and it's very important and I do want to get this in before the v4 release.

[...]

PR #3150, opened 2016-07-12
Implement admin interface analytics

mxstbr:
This is a first pass at implementing admin interface analytics. By abstracting Google Analytics behind a common `Analytics` class we only need a single gate to disable it with an option. (not yet implemented)

This is a very rough early prototype to test the waters if anybody can think of a better approach. @keystonejs/core-devs

Ref: #3075 

**DO NOT MERGE, PROTOTYPE**

### DIFF
--- a/admin/client/Signin/Signin.js
+++ b/admin/client/Signin/Signin.js
@@ -13,6 +13,8 @@ import Brand from './components/Brand';
 import UserInfo from './components/UserInfo';
 import LoginForm from './components/LoginForm';
 
+import Analytics from '../utils/Analytics';
+
 var SigninView = React.createClass({
 	getInitialState () {
 		return {
@@ -25,6 +27,8 @@ var SigninView = React.createClass({
 		};
 	},
 	componentDidMount () {
+		Analytics.init();
+		Analytics.sendEvent('Page', 'render', 'Signin');
 		// Focus the email field when we're mounted
 		if (this.refs.email) {
 			ReactDOM.findDOMNode(this.refs.email).select();
--- a/admin/client/utils/Analytics.js
+++ b/admin/client/utils/Analytics.js
@@ -1,0 +1,40 @@
+/**
+ * Analytics
+ *
+ * Utility class to handle admin interface analytics. Exports Analytics.init, Analytics.sendEvent,
+ * Analytics.sendPageview and Analytics.send
+ */
+
+// Constants
+const TRACKER_NAME = 'adminui';
+const UA_STRING = 'UA-49258834-9';
+
+class Analytics {
+	// Initialize admin interface Google Analytics tracker
+	static init () {
+		this.ga('create', UA_STRING, 'auto', TRACKER_NAME);
+		this.sendPageview();
+	}
+
+	// Send an event
+	static sendEvent () {
+		this.send('event', ...arguments);
+	}
+
+	// Send a pageview
+	static sendPageview () {
+		this.send('pageview');
+	}
+
+	// Send something with our tracker name
+	static send () {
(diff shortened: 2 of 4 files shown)


admin/client/utils/Analytics.js
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	// Send something with our tracker name
+	static send () {
+		this.ga(`${TRACKER_NAME}.send`, ...arguments);
+	}
+
+	static ga () {
+		// TODO Gate here against option to disable
+		window.ga && window.ga(...arguments);
```
mxstbr: This is the only place we'd ever need a gate to disable analytics with an option, which makes that very nice and easy to handle.

admin/server/templates/index.jade
```diff
@@ -65,18 +65,20 @@ html
 		if wysiwygOptions
 			script.
 				Keystone.wysiwyg = { options: !{JSON.stringify(wysiwygOptions)} };
-
+				
+		// Initialize Google Analytics
+		script
+			| window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
```
mxstbr: Need to get the `ga` variable in before any code in our bundle can call it, which is why this script tag is above the bundles

admin/client/Signin/Signin.js
```diff
@@ -25,6 +27,8 @@ var SigninView = React.createClass({
 		};
 	},
 	componentDidMount () {
+		Analytics.init();
+		Analytics.sendEvent('Page', 'render', 'Signin');
```
mxstbr: This is just to show what the usage of this module would look like.

JedWatson:
Heads up, the ga_property and ga_domain you're using are user-settable. The Admin UI can currently send (very simple) statistics to the user's account, that's what you're piggybacking here. Is it possible to configure two accounts to collect stats?

mxstbr:
That's what I'm doing. :grin:

This definitely works as you said, if the user adds an GA account it'll track the pageview for them too, if not only our analytics are used.

admin/client/utils/Analytics.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+// Constants
+const TRACKER_NAME = 'adminui';
+const UA_STRING = 'UA-49258834-9';
+
+class Analytics {
+	// Initialize admin interface Google Analytics tracker
+	static init () {
+		this.ga('create', UA_STRING, 'auto', TRACKER_NAME);
```
mxstbr: By specifying the `TRACKER_NAME` here…

admin/client/utils/Analytics.js
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	// Send a pageview
+	static sendPageview () {
+		this.send('pageview');
+	}
+
+	// Send something with our tracker name
+	static send () {
+		this.ga(`${TRACKER_NAME}.send`, ...arguments);
```
mxstbr: …and prefixing the `TRACKER_NAME` when sending anything this is a completely separate instance of Google Analytics that's entirely separate from the user configured one. See here for more information: https://developers.google.com/analytics/devguides/collection/analyticsjs/creating-trackers#working_with_multiple_trackers

LorbusChris:
What about supporting alternate analytics tools (thinking of Piwik)? Also much like keystone-email, I think this should be optional. Maybe create a separate repo?

bassjacob:
Hey @mxstbr, we're going to close this PR for now and circle back around to this. We'll use this as the basis for the ongoing work.

PR #3180, opened 2016-07-20
Auto-build mongo text indexes for searchable fields

molomby:
Introduces a `searchUsesTextIndex` flag for lists that causes Keystone to create text indexes for the specified `searchFields` while being registered (defaults to false). Listeners have been added so that, if mongo returns an error during index creation, it will bubble out to stderr. If a text index has already been specified *in the Keystone list definition* no new text index will be added (only one text index can exist on a collection).

The `searchUsesTextIndex` flag is also used to enable the filtering logic that utilises these indexes for searches initiated from the admin UI. 

I've also added some debug calls (using the npm `debug` module) to some of the functions that deal with query/filter generation.

The main effect of the above changes is to dramatically reduce response time when searching large lists using the Keystone admin UI. However search behaviour differs slightly compared to the default regex based search. In both cases all `searchFields` are searched for all terms and the search remains case insensitive but, when using text indexes the search behaviour, *only whole words are matched*.

Eg. given a value like `address: {street: '191 Clarence St', suburb: 'Sydney', state: 'NSW'}` in which all fields are indexed, a search for 'clarence Sydney' will return the document with either the text index or regex based approaches but, when searching for just 'clarence syd', the text index search will not return the document since 'syd' doesn't match any complete words.

### DIFF
--- a/lib/list.js
+++ b/lib/list.js
@@ -26,8 +26,9 @@ function List(key, options) {
 		track: false,
 		inherits: false,
 		searchFields: '__name__',
+		searchUsesTextIndex: false,
 		defaultSort: '__default__',
-		defaultColumns: '__name__'
+		defaultColumns: '__name__',
 	};
 
 	// initialFields values are initialised on demand by the getter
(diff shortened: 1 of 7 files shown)


molomby:
Closing this for now.. it doesn't behave the way I expected when there's an existing text index.

Issue #3157, opened 2016-07-14
Columns.__unrecognised__ is not defined.


Noviny:
This is in `ItemsTableRow.js`, line 

### Steps to reproduce the behavior
Have a column that is not in the list of columns (we have encountered a type of ID). It is expected to fall back to 

load fails with error: 
```
type should not be null, undefined, boolean, or number. It should be a string (for DOM elements) or a ReactClass (for composite components). Check the render method of `ItemsRow`.
```

`CreateStaticRouter.js`, has a function `buildFieldTypesStream` that we are now using to build the columns. This looks like it should  include `Columns.__unrecognised__` as a fallback option.

mxstbr:
Yes, this has also come up in #3063

promethe42:
#3063 (the `relationship-display-enhancements` branch) works.

wmertens:
@promethe42 yes, there all the problems were stubbed out (in commit https://github.com/keystonejs/keystone/pull/3063/commits/5ab098a7f2c23980a17c4eff7c5d3727c68614be), but those should not happen in the first place…

JedWatson:
Fixed by #3188

Issue #3130, opened 2016-07-07
ArrayField


jbpin:
Here is a fix for ArrayField mixin
```
      componentWillReceiveProps: function (nextProps) {
		if (!nextProps.value || Array.isArray(nextProps.value)) {
			return;
		}
		if (nextProps.value.join('|') !== reduceValues(this.state.values).join('|')) {
			this.setState({
				values: nextProps.value.map(newItem),
			});
		}
	},
```

mxstbr:
Would you mind submitting a PR with this change so we can take a look? What does this fix?

jbpin:
Sorry.. I'm bit busy but I'll try. I 4.0.0-alpha. I have issue with Array type like NumberArray for exemple. When I create a new "thing" containing this type of property the admin crash because `this.props.value` is undefined. In a same way validator on NumberArray receive a string or a unique number from the field when there is only one element, so inputIsValid return false.

promethe42:
I guess it fixes the issue related to using fields of types Types.TextArray or Types.NumberArray.
Without this, the Admin UI crashes as soon as you create an iteam with a field of one of those types.

I've tried:
* latest master
* 0.3.21
* 0.3.20
* 0.3.19
* 0.3.17
* 0.3.9

On master, the error is:
```
Error: Minified React error #130; visit http://facebook.github.io/react/docs/error-decoder.html?invariant=130&args[]=undefined&args[]=%20Check%20the%20render%20method%20of%20%60ItemsRow%60. for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
TypeError: e is null
require<[449]</i<.getHostNode()
 packages.js:3005
require<[398]</b.getHostNode()
 packages.js:2492
require<[449]</i<.getHostNode()
 packages.js:3005
require<[391]</u<.updateChildren()
 packages.js:2397
require<[440]</v<.Mixin._reconcilerUpdateChildren()
 packages.js:2929
require<[440]</v<.Mixin._updateChildren()
 packages.js:2941
require<[440]</v<.Mixin.updateChildren()
 packages.js:2941
require<[402]</h.Mixin._updateDOMChildren()
 packages.js:2633
require<[402]</h.Mixin.updateComponent()
 packages.js:2622
require<[402]</h.Mixin.receiveComponent()
 packages.js:2613
require<[449]</i<.receiveComponent()
 packages.js:3007
require<[398]</b._updateRenderedComponent()
 packages.js:2515
[...]

PR #3188, opened 2016-07-21
ItemsTableRow to handles 'id' columns

Noviny:
With `Columns.__unrecognised__` currently not existing, and 'id' columns have a set of values not stored in keystone.

This solution allows ID columns to render but leaves two underlying problems:

There should be handling for 'id' fields, or id columns should have the type text intrinsically.

When passing props, the id column object in `itemsTable` has an empty field object, but this does not seem to be passed down.

This addresses issues #3130 and the most common case of #3157.


mxstbr:
Hmm, the code looks fine, is this the solution we want @JedWatson?

JedWatson:
I reviewed this with @Noviny, original fix was good but the new version addresses the underlying issue! Good to merge.

PR #3191, opened 2016-07-22
Add close database connection function

Noviny:


JedWatson:
@Noviny something to consider - how is this invoked?

Look at how `openDatabaseConnection` is added to the Keystone prototype.

Also, in my implementation I didn't need a `keystone` local variable :)

lib/core/closeDatabaseConnection.js
```diff
@@ -0,0 +1,11 @@
+var debug = require('debug')('keystone:core:closeDatabaseConnection');
+
+module.exports = function closeDatabaseConnection (callback) {
+	this.closeDatabaseConnection = function () {
```
JedWatson: You don't need this line at all. `closeDatabaseConnection` is already on the `keystone` prototype, you added it in `index.js`. It's just lines 5:8 that you need in here.

lib/core/closeDatabaseConnection.js
```diff
@@ -0,0 +1,11 @@
...
+		this.mongoose.disconnect(() => {
```
JedWatson: We're still targeting Node 0.12 which means vanilla ES5 on the server-side stuff. We'll switch to ES6 for node 4 (LTS) shortly after that's out.

JedWatson:
:+1:

PR #3212, opened 2016-07-25
Use autoFocus where possible

wmertens:
React does the right thing with it and Elemental-UI already supports it: https://github.com/elementalui/elemental/issues/165

This removes stateful code. It also works in the Modal because the Modal will not render its contents until it opens: https://github.com/elementalui/elemental/blob/620efc246d7f5fd123a2af3fa7f07bf4e745a0fb/src/components/Modal.js#L127-L128


JedWatson:
:+1: looks good @wmertens.

Issue #136, opened 2014-01-22
Add min. password complexity options


JedWatson:
At the moment, password fields validate with any complexity / length.

We should create options on the password field to specify complexity, min. length, valid characters, etc. and enforce a reasonable default.

foobar8675:
I'd like this feature as well.

UrvashIBanthia:
I would like to contribute in this project. Can you please tell us where can I find its source code?

mxstbr:
This would have to be added in `fields/types/password/PasswordType.js`, specifically the `validateInput` function in there!

mxstbr:
We're closing all feature requests to keep the issue tracker unpolluted. From now on submit feature requests on [productpains.com](https://productpains.com/post/keystonejs/add-min-password-complexity-options)!

PR #3218, opened 2016-07-26
Feature/password complexity

dariasamo:
## Description of changes

Password complexity options were added and validateInput function changed to validate password against complexity options. We can set some options to true by default, for example, asciiChar or digitChar to enforce some reasonable default, but user should be able to switch it off as well by setting option to false. If password doesn't meet some of the requirements, a message will be displayed with validation error detail.

## Related issues (if any)

https://github.com/keystonejs/keystone/issues/136

## Testing

- [x ] Please confirm `npm run test-all` ran successfully.


fields/types/password/PasswordType.js
```diff
@@ -133,6 +150,17 @@ password.prototype.validateInput = function (data, callback) {
 		detail = 'passwords must match';
 	}
 	// TODO: we could support a password complexity option (or regexp) here
```
mxstbr: This TODO is no longer needed I'm sure 😉 feel free to remove it!

fields/types/password/PasswordType.js
```diff
@@ -133,6 +150,17 @@ password.prototype.validateInput = function (data, callback) {
...
+	if (min && typeof passwordValue === 'string') {
+		result = passwordValue.length > min;
+	}
+	for (var prop in complexity) {
+		if (complexity[prop] && typeof passwordValue === 'string') {
+			result = (regexChunk[prop]).test(passwordValue);
```
mxstbr: There needs to be a check before here that makes sure that `regexChunk[prop]` exists, and if not console.logging the fact that the `complexity` option was configured wrongly on field with name `"whatever"`.

mxstbr:
This is awesome, amazing job! :+1: 

I left two comments inline with small fixes, also there should be tests for each complexity option, validating both a positive and a negative result + the resulting error messages! Other than that this looks brilliant.

fields/types/password/PasswordType.js
JedWatson: We should be passing errors to the `callback`, not throwing

mxstbr: This is a developer error, not a user error. The developer has a wrongly configured field…

JedWatson: Oh right. If we're validating options, that should happen in the constructor and throw there (so it shows the error when keystone starts, not when validation is called while the app is running)

mxstbr: Does that make sense @dariasamo?

dariasamo: We think so ;) Let us amend PR with changes and we can always fix it if it's not correct.

fields/types/password/PasswordType.js
```diff
@@ -123,16 +138,30 @@ password.prototype.compare = function (item, candidate, callback) {
  * Asynchronously confirms that the provided password is valid
  */
 password.prototype.validateInput = function (data, callback) {
-	var detail;
+	var detail = '';
```
mxstbr: Can detail not be `undefined` here?

mxstbr:
This looks awesome! One small question, what happens when you define a field with an invalid complexity option?

```JS
someField: {
  type: PasswordType,
  complexity: {
    doesntExist: true
  }
}
```

I think that _might_ throw an error?

ostruzina:
@mxstbr: Yeah, good catch. This is actually still work in progress - we still need to work on the wrong config check. I didn't realize that my commit would be added into the PR automatically :)

mxstbr:
No worries, it's good to keep the updates rolling 🚢

mxstbr:
This is awesome, looks perfect! One tiny thing then this is ready to be merged, you'll need to guard the `for-in` statement where the validation of the complexity option happens. (see http://eslint.org/docs/rules/guard-for-in)

mxstbr:
Awesome, amazing work @dariasamo @ostruzina!! 🎉

webteckie:
This is great!   Can this PR be updated with Usage instructions so that when it is time to document this it will be trivial :-)  Or open a related issue in the doc repo and specify the usage there.

dariasamo:
We will update PR with usage instructions then

mxstbr:
> We will update PR with usage instructions then

No worries, submitting a PR to `keystonejs-site` with the documentation would be the best!

ostruzina:
Ok! We plan to do that for all the new features we contributed with.

webteckie:
Actually, isn't the Readme for the password field suppose to document the options?   They show pretty nice in fields explorer:-)

PR #3043, opened 2016-06-15
Api search

WelltokCaleb:
## Description of changes

Adding two API search improvements: 
1. Previously if a search field was a relationship, it failed to parse the object id
2. Can now pass a valid JSON object to search a specific field
## Testing
- [X] Please confirm `npm run test-all` ran successfully.

PR #3226, opened 2016-07-27
implement fix to determine whether input param is string or object an…

pr1ntr:
## Description of changes
modified the `getOptions` function in `RelationshipField` to discern between `input` type. if the input type is a string the ternary breaks and returns '' as `q=`.  Added logic to check `typeof` an act accordingly.

## Related issues (if any)
https://github.com/keystonejs/keystone/pull/3043
There could be others. I am not sure there are a bunch of issues regarding autocomplete/relationships

## Testing

…d act accordingly to set q  value

### DIFF
--- a/fields/types/relationship/RelationshipField.js
+++ b/fields/types/relationship/RelationshipField.js
@@ -117,7 +117,12 @@ module.exports = Field.create({
 	},
 
 	buildOptionQuery: function (input) {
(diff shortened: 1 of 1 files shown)


JedWatson:
Great, thanks @pr1ntr!

pr1ntr:
I think it broke the linter.

PR #3255, opened 2016-08-02
add window.open for click handler of relationship field

pr1ntr:
## Description of changes
WARNING: OPINIONATED
We were having a lot of issues going into our related fields from page to page. I modified the relationship field class to add the onOptionLabelClick handler to the Select (react-select) field. It works. 

IMO, its better than nothing.

## Testing

### DIFF
--- a/fields/types/relationship/RelationshipField.js
+++ b/fields/types/relationship/RelationshipField.js
@@ -184,6 +184,13 @@ module.exports = Field.create({
 		}
 	},
 
+	handleOptionLabelClick: function(value, event) {
+		var refPath = this.props.refList.path;
+		if(value && value.value && refPath) {
+			window.open('/keystone/'+refPath+'/'+value.value);
+		}
+	},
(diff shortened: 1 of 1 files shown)


JedWatson:
@pr1ntr would you mind expanding on this a bit?

> We were having a lot of issues going into our related fields from page to page

Other than that this looks good, we have implemented similar functionality for 0.4. I'd be more happy to merge this if it navigated to the related item in the same window by default, and supported ctrl/cmd+click to open in a new window though!

pr1ntr:
I admin this is a bit hacky, I couldn't figure out how to render an <a> in react-select's  Value -> optionLabel. I couldn't update the react-select to latest either since it requires react 15. Was having a lot of trouble. I realized 0.4.x works well enough and just upgraded anyway.   You can close this if you want.

JedWatson:
OK, thanks @pr1ntr

PR #3258, opened 2016-08-02
Relationship Field: Fix inline create popup close

wmertens:
Without this change, the create popup stays put and the console gets an error about toggleCreate missing.


mxstbr:
LGTM thanks!

PR #3281, opened 2016-08-05
[WIP] Travis restructure and fix e2e

### LINKED ISSUE #3275: Move selenium out of repo
I just had to wait a couple minutes to `git pull` keystone because there was 20 megs of changes in a point release of selenium. Because of how git works the repo contains binary copies of every version of selenium which has ever been checked in. Thats totally unnecessary.

This is bad, and it will only get worse (20 megs at a time with each selenium point release)

```
keystone$ du -hs .git
156M	.git
```

Can we please move selenium out of the repo? It should be in its own module in npm and pulled in as a `devDependancy`. (And the FF & Chrome drivers should be pulled out too.)

And it turns out some other people have already done this: https://www.npmjs.com/search?q=selenium . If none of them are suitable we should make our own `keystone-selenium` package on npm.

Ping @webteckie

jstockwin:
## Description of changes

This is currently a WIP, do not merge. 

I've updated `travis.yml`, which is now a little complicated, but we have a nice build matrix. There are now 10 jobs per build:
- 4 for unit tests, one for each node version (0.12, 4, 5, 6).
- 1 for test coverage reporting and linting
- 5 for e2e, one group of tests at a time. 

Since builds run in parallel, 5 at a time, this should help to speed up builds, especially when there are multiple queued builds. 

I've also put in a fix for some of the sauce_connect errors we sometimes see, so I'm hoping they'll stop. 
There are currently some unused environment variables being set. This is so when you're looking at a travis build (e.g. [here](https://travis-ci.org/keystonejs/keystone/builds/150008919)) you can see which jobs are doing which tests, and so you can quickly see what has failed. 

Meanwhile @webteckie has been working on fixing the e2e tests. This is why this is still a WIP, as I think there is still work to do. (@webteckie can you please check I've managed to merge all your work correctly? I had some problems). 

We have also removed the selenium drivers from the repo, and instead added a dev npm package. This fixes #3275. I'll also be removing the Chrome drivers shortly, as discussed in that issue. 

We hope that this should improve the testing experience, and that there will now be more of a focus on keeping e2e running green. 

Keen for comments/suggestions. @keystonejs/core-devs 

## Related issues (if any)

- Travis occasionally randomly fails. 
- E2E is currently broken. 
- If a queue builds up, it can take a long time to get pass/fail. 

## Testing


mxstbr:
Ohhh this is so awesome, great job @jstockwin 👏

webteckie:
@jstockwin this should have been merged way back then :=)  There's no way to keep up with the updates being done.  So the best thing, IMO, was to merge this in and then keep updating e2e as core changed.  There may be a ton of changes by now TODO.  I'll pull it sometime and see but may need your help catching up e2e with all the breaking changes :=)

jstockwin:
@webteckie no worries. Will happily help update, but am away for a week so will have to wait until after then

mxstbr:
Is there any reason this _shouldn't_ be merged? I mean, this sounds like such a useful version compared to our current Travis setup!

mxstbr:
Can we merge this? What is happening here?

jstockwin:
This has been replaced by #3321

Issue #3269, opened 2016-08-04
Cloudinaryimage/s fields need to be updated to new spec


JedWatson:
The `File` field type has been updated to match the new API spec, now we need to update the `CloudinaryImage` and `CloudinaryImages` components too.

To work with this, you'll need to switch over to the new API by removing both instances of `/legacy` from this file: https://github.com/keystonejs/keystone/blob/master/admin/client/utils/List.js

First up is the `Type` changes: See b16df8facbcabf3a14c74f0b3b6f8b46719815f9 for the equivalent work that was done to `File`.

Once that's in place, update the React `Field` to submit data in the new format, as per these changes: 8ccc4c3df0e723e9ea6a4134a3f5263bfa69bcee

JedWatson:
#3283 has been merged... we're done!

PR #3283, opened 2016-08-05
Update Cloudinary Fields

mxstbr:
Basically I have no clue what I'm doing. :sweat_smile: I tried to take the changes from `816df8f` and apply them here, but nothing seems to be working - what am I missing?

Adresses #3269


JedWatson:
This is now working, except that it breaks compatibility with the legacy API. To fix that, I'm going to switch over to the new endpoints and drop the old file-type fields so we can merge this.

PR #3293, opened 2016-08-10
Adding support for binding express.Router with keystone.set('routes')

JedWatson:
This adds support for using express router instances instead of binding routes on the root express app. It is backwards compatible with the old method of binding routes.

Previously application routes needed to be set on the express app that Keystone initialises, like this:

```js
var initRoutes = function (app) {
  app.get('/', (req, res) => res.send('hello world'));
}
keystone.set('routes', initRoutes);
```

Now, you can provide an `express.Router` instance, which will be used by keystone’s app instance:

```js
var myRouter = new express.Router();
myRouter.get('/', (req, res) => res.send('hello world'));
keystone.set('routes', myRouter);
```

### DIFF
--- a/server/createApp.js
+++ b/server/createApp.js
@@ -118,8 +118,22 @@ module.exports = function createApp (keystone, express) {
 	});
 
 	// Configure application routes
-	if (typeof keystone.get('routes') === 'function') {
-		keystone.get('routes')(app);
+	var appRouter = keystone.get('routes');
+	if (typeof appRouter === 'function') {
+		if (appRouter.length === 3) {
+			// new:
+			//    var myRouter = new express.Router();
+			//    myRouter.get('/', (req, res) => res.send('hello world'));
+			//    keystone.set('routes', myRouter);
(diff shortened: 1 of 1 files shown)


mxstbr:
What's the use case here, why do we need this? (is there a discussion about this somewhere?)

JedWatson:
No background, it came up during an internal discussion at Thinkmill, and I realised it would be a really easy backwards-compatible thing I could implement. @josephg might be able to expand on my thinking here.

Basically, the current method (pass app to a method to initialise routes) is a hangover from when the system was designed for Express 3 (before `express.Router` was a thing)

A lot of the internals have been refactored (e.g. the Admin UI Server) to use routers, because they provide a more encapsulated / scalable way to structure express applications. Initialising an app by passing it around is a really leaky abstraction, and "feels wrong" in a modern express context.

So basically I think allowing a router to be provided, rather than modifying the app directly, is cleaner and makes more sense in an express 4/5 world.

PR #3313, opened 2016-08-14
feat (admin ui custom field): extend keystone field with application custom field

phucpnt:
## Description of changes

Extend keystone admin UI field with custom field defined by application.

## Testing
- [x] Add Unit test for `createStaticRouter`

### DIFF
--- a/admin/server/app/createStaticRouter.js
+++ b/admin/server/app/createStaticRouter.js
@@ -11,12 +11,12 @@ var express = require('express');
 var less = require('less-middleware');
 var path = require('path');
 var str = require('string-to-stream');
-var compose = require('../../../fields/utils/compose').default;
+var compose = require('../../../fields/utils/compose');
 
 function buildFieldTypesStream (next) {
 	return (fieldTypes) => {
 		var types = Object.keys(fieldTypes);
-		let sections = { Columns: {}, Fields: {}, Filters: {} };
+		var sections = { Columns: {}, Fields: {}, Filters: {} };
 		['Column', 'Field', 'Filter'].forEach(i => {
 			types.forEach(type => {
 				sections[i + 's'][type] = '../../fields/types/' + type + '/' + fieldTypes[type] + i;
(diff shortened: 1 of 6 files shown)


mxstbr:
Woah this is an awesome PR @phucpnt!

We're definitely going to work on admin interface extensibility, but I think that these specific changes are too small in scope. We don't just want you to be able to add custom fields, we want you to be able to completely customize every part of it! This means I can't guarantee we'll land this as-is, but this is an amazing PR with tests and everything! 👏

Going to defer to @JedWatson for the final call on if we want to merge this?

phucpnt:
thanks @mxstbr !

I enjoy using keystone, the admin field UI mechanism is well established. It would be awesome for me if keystone have an extendable field ui. I hope the small updates will give provide some level flexibility for keystone.

Let me know @JedWatson

admin/server/app/createStaticRouter.js
```diff
@@ -11,33 +11,50 @@ var express = require('express');
...
+		var types = Object.keys(fieldTypes);
+		var sections = { Columns: {}, Fields: {}, Filters: {} };
+		['Column', 'Field', 'Filter'].forEach(i => {
+			types.forEach(type => {
+				sections[i + 's'][type] = '../../fields/types/' + type + '/' + fieldTypes[type] + i;
+			});
+			if (i === 'Column') {
+				sections[i + 's'].id = '../../fields/components/columns/IdColumn';
+				sections[i + 's'].__unrecognised__ = '../../fields/components/columns/InvalidColumn';
+			}
```
adnasa: is there a good reason why you chose e.g `Column` instead of `Columns` in your iteration?
You could avoid concatenate the key-value with an `s` if you named your keys accordingly

```js
var sections = ['Columns', 'Fields', 'Filters'].reduce((acc, key) => {
  acc[key] = {}
  // ... your logic ...
  return acc
}, {})

return next(sections)
```

PR #3322, opened 2016-08-15
Updatehandler rewrite

### LINKED ISSUE #3319: UpdateHandler Update for 0.4
This is one of the last remaining tasks for 0.4

Now that the APIs have been rewritten and all the field types have been updated to the new spec, we need to rewrite the UpdateHandler to finally address it's inversion-of-control issues (see #1150) and use the new methods.

It should be backwards-compatible, we can look at improving the API another time.

JedWatson:
Mostly a drop-in replacement for the old UpdateHandler functionality, see the notes in the Upgrade Guide for specifics.

Fixes #3319


wmertens:
Looks like it breaks a bunch of tests atm but I like where this is going! 👍 

I assume this needs doing because there is legacy code out there relying on UpdateHandler?

JedWatson:
This is ready!

PR #3343, opened 2016-08-18
Fix pre:validate errors not displaying

mxstbr:
@josephg and @jossmac alerted (hah get it, alerted) me of this problem.
When passing on an error from the `pre:validate` hook it would display in the admin interface as `"Database error"` instead of showing the error messages because Error objects cannot be stringified.

Something like this would show up as "Database error":

```
Model.schema.pre('validate', function (next) {
  if (this.something === false) {
    next(new Error('My error message'));
  }
});
```

This is now fixed, but I'm not sure if there isn't a better way of going about passing that error down to the frontend. @JedWatson @creynders

### DIFF
--- a/admin/server/api/item/update.js
+++ b/admin/server/api/item/update.js
@@ -9,7 +9,7 @@ module.exports = function (req, res) {
 		req.list.validateInput(item, req.body, function (err) {
 			if (err) return res.status(400).json(err);
 			req.list.updateItem(item, req.body, { files: req.files }, function (err) {
-				if (err) return res.status(500).json(err);
+				if (err) return res.apiError(500, err.detail);
 				res.json(req.list.getData(item));
 			});
 		});


creynders:
@mxstbr Do you have any idea why it gets translated to a database error to begin with? Is it a catch-all for any error originating from mongoose?

mxstbr:
> Is it a catch-all for any error originating from mongoose?

Yeah

creynders:
That doesn't sound right IMO. We won't be able to differentiate between "real" database errors and developer defined ones...

mxstbr:
We kinda need to fix this, so I'm inclined to merge as-is and tackle that afterwards?

mxstbr:
Found much better solution now, this works perfectly without needing to change anything in the frontend! :tada:

mxstbr:
Travis failure unrelated to PR, peerDep errors on Node 0.12. Merge!

PR #3344, opened 2016-08-18
Redux fixes Round 2

### LINKED ISSUE #3310: [0.4] admin double tapping loadItem calls, triple tapping on a direct load to  keystone/list?search=xyz URI
Looks like the state machine is borked up in the latest builds in master.  After trying to trace this down i also noticed that a direct load to /keystone/list/search?xyz also does a naked loadItem call twice as well as a search call to the API so you often get incorrect results back in the List view. 

My noobie guess as to whats wrong is that it looks like there is something with the cache invalidator in the store.js file and its relationship to updateQueryParams. Im also a bit suspicious of the loadItem call in Admin/client/App/screens/List/index.js function as it seems extraneous on a lot of the UX.

### LINKED ISSUE #3338: Search fires request with every keypress
We really need to debounce the search field, waiting say 250ms after the last keypress before actually requesting results. Otherwise, a request is fired for every single keypress and it causes unnecessary load on the database.

Open the Keystone Admin UI and navigate to a List. Type `abc` in the search field, quickly.

### Expected behaviour

The action should set a timeout (which is reset with each keypress) and wait for you to finish typing (or at least pause). A single request for the results `?q=abc` should be made.

### Actual behavior

Three requests are made immediately, one each for `?q=a`, `?q=ab` and `?q=abc`

cc @mxstbr @gwyneplaine

gwyneplaine:
Fixing the redux setup with redux-saga.

Fixes #3338, fixes #3310 and a bunch of other bugs

Good to merge.


JedWatson:
@mxstbr @gwyneplaine what's the status of this?

mxstbr:
These aren't all the fixes that need to land, but I think what we ripped out and replaced is good to go?

PR #3360, opened 2016-08-20
Webpackify

wmertens:
Continuing the initial setup by @cameronroe in #2608 

To give this a try, run `git clone keystonejs/keystone#webpackify` and then `npm i && npm run fields-explorer`. Then try changing some field code.

You can also try it in your project with `npm i -S keystonejs/keystone#webpackify-build`. That branch is auto-created by `npm run git-build` and includes the admin UI build, but is not meant for development as it gets force-pushed on build. Npm is fine with that.
## Goals
- replace browserify with webpack
- static build in npm
- no bundled code in git repo
- no bundler deps for keystone
- npm install from git still works
- optimize code size for admin ui, don't download what is not used
## ToDo
- [x] fix array fields
- [x] use webpack for field explorer
- [x] remove browserify-related code and deps
  - [x] npm
  - [x] staticrouter
- ~~[ ] fix silent hang when navigating between lists~~ See #3369, not related to webpack
- [x] clean up webpack config
- [x] remove all built js from git
- [x] npm prepublish setup
- [x] git postinstall setup
- [x] clean up the commits
- [ ] resolve the problem with `.getDefaultValues()`
## Followup tasks

These can be done at any time, and are not strictly part of this PR.
- [x] hot reloading admin UI / fields explorer
- [x] make minimum chunk size work
  - Not a big deal, but would be nice if we didn't have a bunch of 400 byte chunks for small fields. Net effect is probably not noticeable
- [ ] remove 3rd party js from git (see codemirror commits for how)
  - [ ] jquery
  - [x] codemirror
  - [x] tinymce
  - [ ] field bookkeeping code (for tinymce and jquery)
- [ ] build css from webpack too
- [ ] Webpack v2
- [ ] make npm install not bundle fields source for minimal install size
  - Might be as simple as excluding the dirs but maybe something reads those


admin/client/App/shared/CreateForm.js
```diff
@@ -152,8 +133,8 @@ const CreateForm = React.createClass({
 			// If there was no focusRef set previously, set the current field to
 			// be the one to be focussed. Generally the first input field, if
 			// there's an initial name field that takes precedence.
-			if (!focusRef) {
-				fieldProps.ref = focusRef = 'focusTarget';
+			if (!focusWasSet) {
+				fieldProps.autoFocus = focusWasSet = true;
```
mxstbr: Can we move this focus fix out to its own PR? Would love to merge that in...

fields/types/Field.js
```diff
@@ -64,14 +64,18 @@ var Base = module.exports.Base = {
...
+			<FormInput {...{
+				...inputProps,
+				autoFocus,
+				autoComplete: 'off',
+				name: this.getInputName(this.props.path),
+				onChange: this.valueChanged,
+				ref: 'focusTarget',
+				value,
+			}} />
+		);
```
mxstbr: Again, can we move this out to its own PR?

fields/types/html/HtmlField.js
```diff
@@ -40,6 +41,8 @@ module.exports = Field.create({
 
 	initWysiwyg () {
 		if (!this.props.wysiwyg) return;
+		// We only have it in the browser if we have wysiwyg
+		tinymce = require('tinymce');
```
mxstbr: Can we do `require.ensure` maybe?

webpack.config.babel.js
```diff
@@ -0,0 +1,100 @@
...
+		path: path.resolve(__dirname, './admin/public/js'),
+		// NOTE: hardcoding keystone
+		publicPath: '/keystone/js/',
+		pathinfo: true,
+	},
+	module: {
+		loaders: [
+			{
+				test: /fields\/.*(Field|Column|Filter)\.jsx?$/i,
+				loader: 'react-proxy',
```
mxstbr: What does `react-proxy` do?

fields/types/html/HtmlField.js
wmertens: later, see todo list :) that will require some state management

mxstbr:
Love it, this is looking very promising! 

While you're at it would you mind adding a Readme somewhere walking devs through the build process? (part of the better docs efforts)

webpack.config.babel.js
wmertens: It wraps a react component with a `require.ensure` wrapper. Since React doesn't require synchronous loading, that works great. Net result is that each field loads its own dependencies on the fly, like CodeMirror.

wmertens:
You mean, how to run the build or how webpack builds it?

admin/client/App/shared/CreateForm.js
wmertens: Done in #3363

mxstbr:
Everything! Which parts get bundled how, why the configuration is the way it is, how to build for publishing/dev,...

webpack.config.babel.js
mxstbr: That sounds epically good!

fields/types/html/HtmlField.js
wmertens: actually, not sure if this needs it. It's extra work, that field is the only field using tinymce, and loading that field for the first time is async anyway.

JedWatson:
So heads up, there's a specific requirement that we need to make sure has a solution:

We currently have the ability to browserify dynamic code. This is so that we can support field types that come from npm (or anywhere, really) without keystone code knowing about them.

I'm expecting some sort of plugin config for field types, like:

```js
{
  name: 'CloudinaryImage',
  field: require.resolve('./ui/Field.js'),
  filter: require.resolve('./ui/Filter.js'),
  column: require.resolve('./ui/Column.js'),
}
```

Keystone's build process would then consume that information and add it to the `FieldTypes` package. Having that package available (rather than importing from `../../../../fields/...`) is important.

There's probably a way to do this in Webpack, we either need to find it _or_ come up with a different solution for field types.

wmertens:
On Sun, Aug 21, 2016 at 10:09 AM Jed Watson <notifications@github.com>
wrote:

> I'm expecting some sort of plugin config for field types, like:
>
> {
>   name: 'CloudinaryImage',
>   field: require.resolve('./ui/Field.js'),
>   filter: require.resolve('./ui/Filter.js'),
>   column: require.resolve('./ui/Column.js'),
> }
>
>
The way I think this should be done is by exposing a global like
`_keystoneRegister` which takes a structure like that and adds it to the
Field registry.

Then the additional Field UI code needs to be bundled in a js that keystone
injects in the admin ui html.

So there would not be a need to rebuild any js at runtime.

BTW keystone starts up way fast now :)

wmertens:
It works in travis and presumably in npm publish but I just noticed that instalding from git makes it bail on the es6 webpack config, as if babelrc is not read ☹ will fix soon.

JedWatson:
@wmertens we're still going to need the ability to generate the fields bundle at run-time, using information that the node process has for this to work cleanly. There are other reasons and similar patterns we need to follow (e.g. including stylesheets in the admin HTML) so I'm not keen to change direction on this without having finalised the API for field type plugins, which is a post-0.4 thing.

How about, we just generate a `FieldTypes.js` file and map to `require('FieldTypes')` in Webpack? I've been avoiding generating code into keystone's filesystem (I must prefer to do it dynamically, which our browserify config allowed for) but if the only way we can use web pack and retain that capability is to generate files, we can look into that.

My concern about generating code then bundling it is that it means our app has state in its local filesystem, which breaks the "12 factor app" rules and I haven't done any research to make sure it won't cause problems in some hosting environments (like heroku). We'll have to make absolutely sure it works properly there before we commit to that strategy.

JedWatson:
Also worth mentioning I have other plans for generating dynamic bundles that haven't actually been implemented yet, we can talk about those separately, but just for now please take it as a given that I really do not want to have the capability taken away!

wmertens:
The hot reload functionality will be using webpack-dev-middleware in
createStaticRouter, which is basically the same as the browserify mechanism
that was there, and the hotness is configurable. So if that is implemented
as part of this PR it's ok?

We could make a keystone-customize repo that has a createStaticRouter
function with webpack, so that the build dependencies are included there
instead of in the main repo.

Come to think of it, we can probably have npm get a non-build-dep version
by tweaking the package.json at publish time, and then have git include the
build deps as regular deps?

wmertens:
So now hot reloading works too. To try, set `KEYSTONE_BUILD=hot`, run your app and start changing the fields and admin code. Sweet 👯

wmertens:
Just for clarity, if you set `KEYSTONE_BUILD` to `prod` or `dev`, you get the previous behavior.

JedWatson:
Re small chunks:

> Not a big deal, but would be nice if we didn't have a bunch of 400 byte chunks for small fields. 

In a high latency environment that will definitely be noticeable (e.g on mobile)

wmertens:
It is mitigated by the fact that all those requests happen in parallel.
Also nice is that you don't have to download an editor if the form you are
using does not need it.

wmertens:
And if that is a showstopper we can hand-pick fields to load dynamically…

wmertens:
@JedWatson the chunk sizes are also solved.

Any thoughts on how to handle the `.getDefaultValue()` call on the Field components? I feel this does not belong in there, and it doesn't work with dynamically loaded fields.

One approach would be to make the Fields be a registry on the `Keystone` global, and getting a field component is an asynchronous operation. Then we can start with the user-provided field types.

wmertens:
Now TinyMce is out of the repo too! 🎉 

It gets loaded only the first time you encounter a wysiwyg field. The code is hosted from keystone still, but via the npm package, for secure hosting, a clean repo and easy updating.

wmertens:
@JedWatson this is what the build result looks like: https://github.com/keystonejs/keystone/commit/650210f18ec522108c03bd8aa91c521b7867b390

This is on a separate branch that gets created and force-pushed by `npm run git-build`. The branch is a single build commit off the current branch, whatever the branch name or origin. That way, no postinstall build is needed, but it can still do it if necessary.

JedWatson:
@wmertens what if we turned off chunking builds?

I think that pre-generating the file to consume for FieldTypes (which we currently do, and provide to browserify as a string, but could be written to a file on start) is a reasonable workaround to ensure we are including the correct field types.

And while we lose the optimisation of not, for example, having the CodeMirror package loaded until you view a screen that requires it, it would mean that no developers need to be aware of environment weirdness that changes how you write client-side code to allow for asynchronous module loading.

IMO for the predictability that offers, and given what I think the usage patterns of the Admin UI are (you're not likely to load a significant amount of code that won't be used in a session), it's a good trade-off.

What do you think?

wmertens:
>
> @jedwatson the problem is that we then cannot have field plugins, which
means eg. cloudinary would have to be in the keystone repo forever.

We can make the proxying smarter, instead of using react-proxy we can
create our own async wrapper that provides the defaults synchronously but
the actualy rendering asynchronously.

Still, I think the component should not have the default value function at
all, it should inject that in its inputs itself.

>

wmertens:
@JedWatson if you want to merge this before 0.4, I recommend doing it earlier than later, it actually touches quite a bit of code and keeping it up to date is not trivial

wmertens:
Also, @jossmac not sure if you realize, but hot reload makes developing elemental a lot easier :)

jossmac:
@wmertens
> hot-reload makes developing ~~elemental~~ a lot easier

😄

ehd:
@wmertens This is awesome, can I help somehow? On a heroku 1x dyno the first requests to the JS bundles actually always time out (>30s), so this would be a super welcome change for production 😄

wmertens:
@ehd well I'm super busy so help would definitely be appreciated: This is the list of things to do, from conversations with Jed:

* [ ] rebase on current master. Note that some of the commits delete a bunch of vendor code and that code was updated so the commits need to be redone
* [ ] remove the react-proxy loader, and instead perform the `require.ensure` of the heavy libraries manually, sort of like I did in the codemirror field. This makes the `.getDefaults()` work again.
* [ ] maybe upgrade to webpack 2.1, however I had trouble doing that in another project. Better to do that after this lands, it's just an optimization

@JedWatson anything I missed?

jeremyzahner:
@wmertens @JedWatson We're thinking about making keystonejs our defacto agency-standard. Thou, bundling via webpack would be crucial to us. Is there anything we can do to make some progress here?

>  rebase on current master. Note that some of the commits delete a bunch of vendor code and that code was updated so the commits need to be redone
 remove the react-proxy loader, and instead perform the require.ensure of the heavy libraries manually, sort of like I did in the codemirror field. This makes the .getDefaults() work again.
 maybe upgrade to webpack 2.1, however I had trouble doing that in another project. Better to do that after this lands, it's just an optimization

Is this still the current state of things?

wmertens:
@jeremyzahner yes, correct. Help very welcome :)

jeremyzahner:
@wmertens So, in my opinion, we should delay the Webpack 2.1 upgrade and go for an "MVP" merge here. So I'm making the following assumptions:

- the whole bunch of vendor files that are conflicting should, after all, be deleted for the new feature.
- we are not bringing Webpack 2.1 in in this stage.

Please correct me if I'm wrong.
Otherwise ill go for that.

jeremyzahner:
@wmertens Could you reset onto `jshmrtn/keystone#webpackify` (#3758) here?

jeremyzahner:
@wmertens I guess we should make the reset so the rebase is in. After that, we can find a solution for the react-proxy-loader problem.

jeremyzahner:
@wmertens We did some additional improvements in #3765.

wmertens:
@jeremyzahner can you also try to make https://travis-ci.org/keystonejs/keystone/jobs/181950498#L1048 pass? I'm not sure if I fixed it before or not, I thought I did.

Basically, that test is failing because codemirror was updated…

jeremyzahner:
@wmertens I can try. I need to get a selenium env ready first. My firefox is like 16 versions up from what is needed for the selenium webdriver. We should really switch to Marionette thou.

https://developer.mozilla.org/en-US/docs/Mozilla/QA/Marionette

jstockwin:
@wmertens / @jeremyzahner Do you still need help with the e2e tests? I've got some time and can probably get them going for you if you like. Might be worth resolving conflicts with master before I do that?

wmertens:
@jstockwin resolved :) I think there's only that codemirror test that needs tweaking, thanks!

package.json
```diff
@@ -69,68 +55,79 @@
...
+    "redux-thunk": "2.1.0",
     "rimraf": "2.5.4",
+    "scriptjs": "^2.5.8",
     "sinon": "1.17.6",
+    "style-loader": "^0.13.1",
     "superagent": "3.2.1",
     "supertest": "2.0.1",
     "uglify-js": "2.7.5",
     "updtr": "0.2.3",
     "watch": "1.0.1"
```
LorbusChris: Tests fail here due to the missing comma behind watch

LorbusChris:
@wmertens Tests are (also) failing due to a syntax error in package.json, there's a comma missing before webpack.

jstockwin:
I'll fix that trailing comma with the rest of the stuff I'm doing now.

wmertens:
maybe the new web-based conflicd resolver is not that useful :) Thanks Jake!

>

jstockwin:
Something strange is happening with the build. The codemirror test _should_ be fixed, but this is happening:
https://travis-ci.org/keystonejs/keystone/jobs/184481175#L475

Notice that it only happens on 1 of the 11 jobs in [that build](https://travis-ci.org/keystonejs/keystone/builds/184481173), even though `npm install` runs on all of them.

I've re-ran that job 3 times now and it happens every time. ¯\\_(ツ)_/¯

jstockwin:
So, I've fixed the weird problem - we lost a dev-dependency in the merge from master. 

The code mirror tests now pass, but we now fail on markdown. 
In current master, the markdown field has some nice UI around it - you can click "bold" etc, and then preview your markdown. This is no longer present here. 

This is a real regression, so it's not a case of fixing the tests - we need to work out why this functionality has gone, and add it back in.

LorbusChris:
I found an issue with this branch:

If there is a custum admin URL path specified, it is not respected in the injected html script tags containing the chunks on the Admin UI:
```
<script type="text/javascript" charset="utf-8" async="" src="/keystone/js/0.js"></script>
<script type="text/javascript" charset="utf-8" async="" src="/keystone/js/2.js"></script>
```

Sources here should be "/customAdminPath/js/0.js" and "/customAdminPath/js/2.js".

Also, to me the postinstall ssetup seems rather fragile in general. 

Wouldn't this be the right opportunity to split the Admin UI out into its own package?

wmertens:
I don't use a custom URL, so any assistance with that issue would be
appreciated.

The postinstall is indeed fragile, but that is only meant for direct git
installs from the source branch; the `git-build` command makes a one-commit
branch with the proper build, and npm will also have the build.

>

LorbusChris:
The issue with this is that right now, webpack does its bundling right after install, without having any of the config options.

I think first we need to find out what AdminUI specific options, if any others than `admin path`, need to be known before bundling the Admin UI ('csv field delimiter'?).

My idea would then be to export a webpack config factory that can be invoked by the user to do custom builds on demand.

wmertens:
Hmm, the bundling does not generate html files, so they should be generated
with the given configuration (with the keystone init), no?

wmertens:
closing this unmerged to clean up my backlog but feel free to comment

Issue #1593, opened 2015-08-20
Cannot create new record


ChristianRich:
On a specific model I get a client-side JavaScript error:
Uncaught TypeError: this.props.value.map is not a function in fields.js 109

This is where it goes wrong:

getInitialState: function getInitialState() {
        return {
            values: this.props.value.map(newItem)
        };
    }

ChristianDen:
After deleting the List and removing the database record it worked again.

olivierlizotte:
I have the same javascript error happening on a production system. A single type of model is affected. The error gets thrown from the page listing the objects and from the editing page of an object. Its now impossible to create a new object of that type from the keystone admin.
Is that some sort of corruption of my data on the DB? Any known workaround?

ChristianRich:
@olivierlizotte I used MongoChef to backup and the drop the affected collection. Afterwards I restarted Keystone, made a few records in the admin UI and then re-imported the collection using MongoChef. This resolved the issue.

olivierlizotte:
Thanks for the input @ChristianRich. 
I installed the application on a different server and mongodump+mongorestored the database locally. The problem followed to the new system. So now I have a staging system I can test solutions on. I will try and have a look at the affected collection, see if I can spot anything suspicious.

olivierlizotte:
The collection seems fine. I noticed that my last change was to add a String.TextArray field type, and if I change it back to String, the error disappears for the collection. Changing it back to TextArray brings back the error. Anyone ever tested TextArray in the admin UI?

creynders:
You have string data for that field in one of your documents, which is why it chokes when you try to reconfigure it as a TextArray

olivierlizotte:
I checked and all documents have arrays. Even the empty documents have empty arrays.
The error is on the client side, while trying to open the dialog to create a new document. The list of documents is properly listed, and all of them are editable. 
I am running keystone 0.3.15 with mongoose 4.2.3 and mongodb 3.0.7.

vostreltsov:
I've run into exactly the same problem. Dropping database didn't help.
The error appears when i set "initial: true" on a TextArray field.
Though if i set "initial: false" and try to edit a newly-created record, it works.

NebirosFeR:
Same problem.

olivierlizotte:
I confirm that setting "initial: false" for the TextArray field makes the admin fully usable again in my case (although two saves are now necessary to create the object).

loffek:
[...]

PR #3384, opened 2016-08-23
Fix array field bug#1593

loffek:
fixes issue https://github.com/keystonejs/keystone/issues/1593

Problem:
- `initial: true` on an TextArrayType breaks the React front end with "Uncaught TypeError: this.props.value.map is not a function in fields.js 109", unless `default: ['']` is set explicitly
- If the array of a TextArrayType has length zero or one, and some validation rule fails so that the "Create new"-modal is shown again with an error-message, the React Frontend breaks with "Uncaught TypeError: this.props.value.map is not a function in fields.js 109".
- Old values in the db that are not arrays probably also breaks the front end.
## Description of changes

The problem is that in the above scenarios, an ArrayField is created with `props.value` not being an array (eg. undefined or just a string). By ensuring that the value is an array before calling `.map()`, the bug is avoided.
## Related issues (if any)

https://github.com/keystonejs/keystone/issues/1593
## Testing

I couldn't find any established tests for the React frontend, so I haven't written any tests for this bug. Please let me know how this could be done if it's required.


mxstbr:
This looks like a solid fix, will check it out tomorrow! Thanks so much!

AurelioDeRosa:
Is this fix good to merge?

hulvei3:
How about that merge? :) Dunno if how it's behaving with the upcoming 0.4+ though..

Noviny:
We're not accepting PRs into 0.3 anymore, sorry. Thankyou for your contribution.

PR #3393, opened 2016-08-24
Fix field defaults

JedWatson:
cc @mxstbr


wmertens:
Yes awesome! :)

wmertens:
I cheered too soon.

On closer inspection, this is problematic in the webpack proxied fields case. The reason is that the wrapped field component does not have the `.getDefaultValue` method.

So in a future with dynamically loadable fields, this method needs to be called only when the actual field component is loaded.

However, taking a step back, it is also a bit of a leaky abstraction. Why is the display component providing the form manager with the default data value?

So I wonder, is this method needed at all? The Field should have an internal default if its provided `value` is `null` (React 15 assumes that `undefined` means unbound), and whenever it changes it should call `onChange`. When you post the form, the default value is extracted from the DOM.

PR #3397, opened 2016-08-24
FileField: show image thumbnails [WiP]

wmertens:
Make filefields look like this if they are an image:

image

Still to do:
- [ ] proper styling, instead of the inline `flex`. How should I do this?
- [ ] allow disabling thumbnails, how about with `plain: true`?

### DIFF
--- a/fields/types/file/FileField.js
+++ b/fields/types/file/FileField.js
@@ -10,6 +10,48 @@ import { Button, FormField, FormInput, FormNote } from 'elemental';
 import FileChangeMessage from '../../components/FileChangeMessage';
 import HiddenFileInput from '../../components/HiddenFileInput';
 
+const FileThumb = ({ url }) => {
+	const isPicture = url && url.match(/\.(jpeg|jpg|gif|png)$/i) != null;
+	if (!isPicture) {
+		// TODO generic icons
+		return false;
+	}
+	return (
+		<div style={{ width: 150, marginRight: 10, flexShrink: 0 }}>
+			<img style={{ width: '100%', height: '100%' }} src={url}/>
+		</div>
+	);
+};
+
+const FileDom = ({ url, filename }) => {
+	return (
+		<div style={{ display: 'flex' }}>
+			<FileThumb {...{ url }}/>
+			<div style={{
+				display: 'flex',
+				flexDirection: 'column',
+				justifyContent: 'flex-end',
+				alignItems: 'flex-start',
+				minHeight: 100,
+				width: '100%',
+			}}>
+				<FileChangeMessage>
+					{url ? (
+						<a href={url}>{filename}</a>
+					) : (
+						filename
+					)}
+				</FileChangeMessage>
+				{url && (
+					<span style={{ fontSize: 10 }}>
+						url: {url}
+					</span>
+				)}
+			</div>
+		</div>
+	);
+};
+
 let uploadInc = 1000;
 
 const buildInitialState = (props) => ({
@@ -43,8 +85,10 @@ module.exports = Field.create({
 		return this.props.collapse && !this.hasExisting();
 	},
 	componentWillUpdate (nextProps) {
+		const value = this.props.value || {};
+		const nextVal = nextProps.value || {};
 		// Show the new filename when it's finished uploading
-		if (this.props.value.filename !== nextProps.value.filename) {
+		if (value.filename !== nextVal.filename) {
 			this.setState(buildInitialState(nextProps));
 		}
 	},
@@ -60,9 +104,11 @@ module.exports = Field.create({
 		return this.props.value && !!this.props.value.filename;
 	},
 	getFilename () {
-		return this.state.userSelectedFile
-			? this.state.userSelectedFile.name
-			: this.props.value.filename;
+		const { value } = this.props;
+		const { userSelectedFile } = this.state;
+		return userSelectedFile
+			? userSelectedFile.name
+			: value && (value.originalname || value.filename);
 	},
 
 	// ==============================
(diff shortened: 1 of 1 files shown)


mxstbr:
How about using the same styling/component the CloudinaryImage field uses?

wmertens:
Sounds good, but that code seems a bit involved. Could you outline the changes needed?

mxstbr:
No idea, I didn't write it 😀 you'll have to dig in!

JedWatson:
:+1: good feature, @wmertens 

IMO this should be opt-in, because unlike images from cloudinary which we can dynamically resize, general images may be *really large* (i.e multiple megabytes) which I don't think we should be blindly foisting on users by default.

How about `thumbnail: true` as the field option to turn this on?

Also, it might be worth looking at something like [sharp](https://github.com/lovell/sharp) for thumbnail processing in Keystone. There's also a neat [image-resizer](https://github.com/jimmynicol/image-resizer) project that's sort of like a DIY cloudinary, but it's nontrivial in size and to implement so I don't want to bundle it with Keystone by default.

Then again, we're getting into `Image` field territory here.

fields/types/file/FileField.js
```diff
@@ -10,6 +10,48 @@ import { Button, FormField, FormInput, FormNote } from 'elemental';
...
+		<div style={{ width: 150, marginRight: 10, flexShrink: 0 }}>
+			<img style={{ width: '100%', height: '100%' }} src={url}/>
+		</div>
+	);
+};
+
+const FileDom = ({ url, filename }) => {
+	return (
+		<div style={{ display: 'flex' }}>
+			<FileThumb {...{ url }}/>
```
adnasa: Wait. why are you sending an object containing `url` when only `url` is interesting to the `FileThumb` component?

SharadKumar:
This middleware might help replace most of Cloudinary stuff and work well with S3!
 https://www.npmjs.com/package/image-steam

--
Sharad

On 31 Aug 2016, at 5:27 PM, Jed Watson <notifications@github.com<mailto:notifications@github.com>> wrote:


? good feature, @wmertens<https://github.com/wmertens>

IMO this should be opt-in, because unlike images from cloudinary which we can dynamically resize, general images may be really large (i.e multiple megabytes) which I don't think we should be blindly foisting on users by default.

How about thumbnail: true as the field option to turn this on?

Also, it might be worth looking at something like sharp<https://github.com/lovell/sharp> for thumbnail processing in Keystone. There's also a neat image-resizer<https://github.com/jimmynicol/image-resizer> project that's sort of like a DIY cloudinary, but it's nontrivial in size and to implement so I don't want to bundle it with Keystone by default.

Then again, we're getting into Image field territory here.

-

fields/types/file/FileField.js
wmertens: @adnasa This is just React, it's exactly the same as saying `<FileThumb url={url}/>` and gets transpiled to the exact same `React.createElement` call.

wmertens:
@SharadKumar I think that in order to implement this properly we will need https://github.com/asilvas/node-image-steam/issues/39 but I haven't thought it all through yet.

fields/types/file/FileField.js
mxstbr: Might be worth changing that to `<FileThumb url={url} />` – I wasn't 100% sure what that did either.

wmertens: I'm hoping that JSX will allow writing `<FileThumb {url}/>`, see https://github.com/facebook/jsx/issues/23

Would that have been less confusing? If so, some upvotes would be nice :)

bassjacob:
Hi @wmertens, thanks for the work on this PR so far. The scope of this seems to have blown out a little bit in the comments, but I think it does need to be styled to get merged. Are you still interested in getting this in?

wmertens:
Yes, interested, but 0 time right now…

htor:
@wmertens If this feature gets merged then there will be lots of happy people out there - me included. Just the last bit that Keystone needs IMO. Hope you get some time soon!

fields/types/file/FileField.js
adnasa: since it is a single prop

`<Component singleProp={value} />`

is much more readable than

`<Component {...{ singleProp } />`

htor:
This should really be merged...

JedWatson:
@htor we'd love to have this feature but it needs proper styling before it can be merged, and a way for developers to opt in to it (or at least more thought around the logic that enables it). There are a number of scenarios where blindly rendering a preview of a file based on the file type could be bad UX; key being files can be quite large images, so on mobile or a slow connection simply navigating to the item details screen could kick off a huge download (e.g. original photos)

outluch:
Come on, it is so essential feature to show local images in admin UI...

wmertens:
@outlunch feel free to implement this then :-D

outluch:
Already, but I'm not familiar with github pullrequests, etc. Maybe someone could make it correct way? 
Here is my bad pull-request:
https://github.com/keystonejs/keystone/pull/4198

htor:
@wmertens  
this can now be closed as image thumbnails are now implemented in https://github.com/keystonejs/keystone/pull/4509

Pyav123:
@htor , I am using keystone@4.0.0-beta.8 but it is not showing image thumbnail using the local storage, I am using 
image:{
        type: Types.File,
        storage: newStorage,
        thumb: true
    }
Can someone please tell what else I have to do? I need this feature.

htor:
@Pyav123 
at the moment you need to depend on the git version of keystone (don't think npm is updated yet), so put this in your package.json and run npm install:

    "keystone": "git@github.com:keystonejs/keystone.git"
then see what happens.

Pyav123:
Hello @htor  I saw your suggestion on https://stackoverflow.com/questions/40879375/keystonejs-4-0-file-system-storage-adapter-image-preview but this is also not working for me, I am getting the code implemented for this inside keystone but unable to see the preview. Please suggest on this.

Issue #3404, opened 2016-08-24
Filters can't be filled in, modal disappears on keypress


wmertens:
### Steps to reproduce the behavior

* Run latest master
* go to a list view
* select a filter with a text field
* type a letter in the filter

### Expected behavior

* filter has letter

### Actual behavior

* filter modal gone

mxstbr:
Can replicate, on it!

PR #3405, opened 2016-08-24
Fix filter popout closing when any button is pressed

mxstbr:
I have no idea why this code was there. The filter popout was closing
because `this.props.isOpen` was _always_ undefined – we simply did't
pass anything in from `ListHeaderToolbar`! (and never have)

I also have no idea what this code is doing there and why this only
showed up now (probably due to some bug fix in the past few days), but
as far as a I can tell from digging through the code this is good to
remove.

Closes #404


JedWatson:
@mxstbr I don't actually think you meant to close #404 (although that can be closed anyway)

mxstbr:
Oops, I meant to close #3404 of course!

Issue #3440, opened 2016-08-30
Restore support for CloudinaryImage publicId option


JedWatson:
The `publicId` option for the CloudinaryImage field type was dropped in the upgrade, we should re-implement it.

ref http://keystonejs.com/docs/database/#fieldtypes-cloudinaryimage

andrewlinfoot:
I think this issue should've been closed after https://github.com/keystonejs/keystone/pull/3441 was merged

PR #3441, opened 2016-08-30
Add generateFilename api for CloudinaryImage

andrewlinfoot:
## Description of changes
Implemented the generateFilename api for CloudinaryImage

The _getFilename and _retryFilename functions are directly copied from the FS adapter: https://github.com/keystonejs/keystone/blob/master/lib/storage/adapters/fs/index.js
## Testing


mxstbr:
Are there docs?

andrewlinfoot:
Not yet. Good call. Will update in the morn (2:30am) in Seoul

fields/types/cloudinaryimage/CloudinaryImageType.js
```diff
@@ -31,10 +40,17 @@ function getEmptyValue () {
 function cloudinaryimage (list, path, options) {
 	this._underscoreMethods = ['format'];
 	this._fixedSize = 'full';
-	this._properties = ['select', 'selectPrefix', 'autoCleanup', 'publicID', 'folder', 'filenameAsPublicID'];
+	this._properties = ['select', 'selectPrefix', 'autoCleanup', 'publicID', 'folder', 'filenameAsPublicID', 'generateFilename', 'whenExists', 'retryAttempts'];
```
JedWatson: It could be clearer, but these are the properties that are serialised for the client. So we don't need to add the filename related ones here. I realised we can also take out the `'publicID', 'folder', 'filenameAsPublicID'` properties as well, the client doesn't use them.

fields/types/cloudinaryimage/CloudinaryImageType.js
```diff
@@ -31,10 +40,17 @@ function getEmptyValue () {
...
 
 	// TODO: implement filtering, usage disabled for now
 	options.nofilter = true;
+	if (options.filenameAsPublicID) {
+		// Produces the same result as the legacy filenameAsPublicID option
+		options.generateFilename = nameFunctions.originalFilename;
+		options.whenExists = 'overwrite';
+	}
+	options = assign({}, DEFAULT_OPTIONS, options);
+	options.generateFilename = ensureCallback(options.generateFilename);
```
JedWatson: `options.filenameAsPublicId` support is nicely done :)

Should we default to an undefined filename rather than random though? (leaving cloudinary to generate the public ID as per the old implementation)

JedWatson:
This looks great @andrewlinfoot, thanks! Some comments inline, and two general thoughts:

[1] If you can update the `README.md` in the `cloudinaryImage` folder, and the Upgrade Guide in `docs/guides` that would be awesome.

[2] What do you think of us adding the `prototype._retryFilename` and `cloudinaryimage.prototype._getFilename` to the `keystone-storage-namefunctions` package so we're not copying them into different field / storage adapter implementations?

e.g.

```
var storagePrototypeMethods = require('keystone-storage-namefunctions/prototype');
cloudinaryimage.prototype._getFilename = storagePrototypeMethods.getFilename;
cloudinaryimage.prototype._retryFilename = storagePrototypeMethods.retryFilename;
```

andrewlinfoot:
Thanks @jed. Will implement those changes and add the docs tonight.

Re moving `prototype._retryFilename` and `prototype._getFilename`: I think that's definitely a solid solution. Copy pasting didn't seem like the right answer but I wasn't sure where would be best to put them. Can you update the `keystone-storage-namefunctions` package and I'll make the change here when I do the other changes?

jed:
No worries, @andrewlinfoot. Keep up the good work!

JedWatson:
lol this happened again 😀

@jed if I ever make it to NYC or you ever come to Sydney, we should hang out.

jed:
I agree! meeting you in person will allow me to hone my GitHub
impersonation routine. good weekend to ya!

JedWatson:
note: linking this to #3440, it should close that issue

JedWatson:
Thanks @andrewlinfoot! looks great.

PR #3442, opened 2016-08-30
Custom authentication v3.22

danielbchapman:
## Description of changes

**NOTE: this is a merge on top of tag 3.22, I'm not sure what github is saying about wanting to merge 75 commits to master. I'm also not sure if you patch older versions, I assumed the npm "latest" was the stable build to patch against.**
Keystone.js (v3.22 which is what I believe is published on npm) crashes when you use a custom user object  and keystone is set to ignore authentication `keystone.init { ... , auth: false, ... }`. This patch just extends the 'auth: false' value to the view so that the base template doesn't attempt to render a 'signed in as' link (which crashes the administration console if it isn't a Keystone collection). 

I've made a stackoverflow post about my issue here and it just seemed like a better idea to fix it myself as it looks like v4.0+ is an entirely new template engine (React).

http://stackoverflow.com/questions/39215521/is-it-possible-to-user-custom-authentication-in-keystone-js-without-a-keystone-u/39231610#39231610

This post includes a patch-file if that is preferable. This simply makes keystone more flexible if you are not depending on it for authentication and authorization.
## Related issues (if any)

Unknown, but I think any application with a local `user` object that didn't share a User model with Keystone would probably crash.
## Testing

`npm run test` ran successfully, there was no 'test-all' configured on this branch.


bassjacob:
Hi @danielbchapman, thanks for the work on this PR.

Unfortunately I don't think we can merge this in as is, so I'm going to close it. Please let us know if this is still a feature you would like and we can discuss strategies for developing it.

PR #3443, opened 2016-08-31
removed call to populate createdBy and updatedBy fields in api/item/get

mikehazell:
## Description of changes
Remove the 2 calls which populate createdBy and updatedBy 

## Testing

No, this was too hard.


JedWatson:
Thanks @mikehazell!

PR #3457, opened 2016-09-05
Feature/twitter field

dariasamo:
## Description of changes

New twitter field type based on text field added. It handles input such as '@username' or 'username' as well as 'https://twitter.com/username', 'https://twitter.com/@username' and versions without protocol. 'Format' method defaults to 'stripUsername' which removes domain and @ from twitter username and returns a naked twitter username.
## Related issues (if any)

https://productpains.com/post/keystonejs/twitter-field
## Testing
- [x ] Please confirm `npm run test-all` ran successfully.


JedWatson:
@dariasamo this looks great!

Will do a more thorough review shortly, but some quick feedback:

- I recommend pulling out the `stripUsername` method into a separate module which you can require in both the client-side components and TwitterType file, to reduce duplication. This could be placed in `/fields/types/twitter/utils/stripUsername.js`
- Turning `indexOf` into a truthy value by adding `1` is clever but unusual, took me a second to work out what you're doing. I'd prefer to use the more verbose but immediately clear expression `if (indexOf(str) !== -1)` rather than `if (indexOf(str) + 1)`

dariasamo:
@JedWatson 
Thank you for feedback! Recommended changes pushed.

mxstbr:
Wow this looks amazing from a first glance! 👏 Will review more thoroughly next week if @JedWatson isn't faster!

bassjacob:
Hi @dariasamo, thanks for the work in this PR. We're going to implement twitter as one of the early field plugins, and as a result we're going to close this PR since we aren't going to merge it into core. We are going to use it as the basis for the twitter field in the plugin, and we'll credit your work once it's published.

PR #3458, opened 2016-09-06
Handle mongo error in ensureTextIndex

mikehazell:
## Description of changes

Checks for error code 26  in the collection.getIndexes callback. Defaults indexes object so the new text index gets created when the database is connected.

## Related issues (if any)
#3334 
#3180

Setting `searchUsesTextIndex: true` on a Model would cause keystone to
crash on first launch while initialising the database.


JedWatson:
Thanks @mikehazell!

PR #3459, opened 2016-09-07
Confirmation ux

twalve:
Added back ticks around the HTML content that appears in the confirmation dialogs, such as when members are removed from a list in the admin

Also, in the confirmation method, rather than expecting the reader to parse what the `e` param is, I went
ahead and was more explicit. It’s not in the conventions, but I warrant
it could be.

### DIFF
--- a/admin/client/App/screens/List/index.js
+++ b/admin/client/App/screens/List/index.js
@@ -333,13 +333,13 @@ const ListView = React.createClass({
 			checkedItems: {},
 		});
 	},
-	deleteTableItem (item, e) {
-		if (e.altKey) {
+	deleteTableItem (item, event) {
+		if (event.altKey) {
 			this.props.dispatch(deleteItem(item.id));
 			return;
 		}
 
-		e.preventDefault();
+		event.preventDefault();
 
 		this.setState({
 			confirmationDialog: {
--- a/admin/client/App/screens/List/index.js
+++ b/admin/client/App/screens/List/index.js
@@ -175,12 +175,12 @@ const ListView = React.createClass({
 				isOpen: true,
 				label: 'Delete',
 				body: (
-					`<p>
-						Are you sure you want to delete ${itemCount}?
+					<p>
+						Are you sure you want to delete {itemCount}?
 						<br />
 						<br />
 						This cannot be undone.
-					</p>`
+					</p>
 				),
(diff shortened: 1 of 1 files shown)


wmertens:
Are you saying that the confirmation dialog wasn't working without the backticks? I'm having trouble believing that.

You converted them from a react element to a string, which is also a valid react element, but now the html tags should be visible. Did you test this?

For the `e`, indeed, event handlers are so common that the event eventually gets abbreviated to `e` or `ev`.

mxstbr:
I'm pretty sure this was working as expected before? I fixed this for the beta release…

admin/client/App/screens/List/index.js
```diff
@@ -333,25 +333,25 @@ const ListView = React.createClass({
 			checkedItems: {},
 		});
 	},
-	deleteTableItem (item, e) {
-		if (e.altKey) {
+	deleteTableItem (item, event) {
```
JedWatson: Taking `e` as the event argument is pretty consistent and standard, I'm OK with leaving this as is

admin/client/App/screens/List/index.js
```diff
@@ -333,25 +333,25 @@ const ListView = React.createClass({
...
+		event.preventDefault();
 
 		this.setState({
 			confirmationDialog: {
 				isOpen: true,
 				label: 'Delete',
 				body: (
-					<p>
+					`<p>
 						Are you sure you want to delete <strong>${item.name}</strong>?
```
mxstbr: The real solution to #3463 would be to remove this `$`, instead of all of that backtick business. Would you mind reverting those changes and removing the `$` here? Thanks!

twalve: The back ticks were added in the hopes that Babel would then leap into action and complete the previous partial solution. Reversion is in place and I've retested simply removing the errant legacy `$`, and that change is ready to be committed.

wmertens:
@twalve looks like you forgot to commit the removal of the $ 🙂

jossmac:
I've fixed this in the `jossmac/0.4/elemental-refactor` branch. Safe to close

wmertens:
ok so looks like this will be fixed soon when merging the elemental branch. I'll close this, but thanks for your efforts @twalve!

PR #3472, opened 2016-09-10
Elemental Refactor

jossmac:
Replaced all components from the elemental UI library with internal alternatives running Aphrodite -- Beginning the transition to CSS in JS; meaning client-side validation and a theme-able Elemental in the near future.

This branch has had ~40h put into it and is much more of an undertaking than I’d first thought. Basically I don’t want the branch to go stale so I’m asking for more eyes on the code.

Thanks!


wmertens:
Just curious, why not branch elementalui to have aphrodite? It seems weird to have a component library be part of Keystone, especially for when field plugins become possible and they would also need ui components…

JedWatson:
@wmertens so we can focus exclusively on *just* the features Keystone is using first. Reworking Elemental is an even bigger task that we will get to once the components are stable here.

To be clear, this is a temporary stage during the transition, the components will eventually be moved out again when we're ready. In the meantime this "monorepo-inspired" approach streamlines the process a huge amount.

JedWatson:
Oh also

> Just curious, why not branch elementalui to have aphrodite

This isn't just a simple port from LESS to aphrodite (there is no such thing), most of the components (particularly the styles but also some of the APIs) have been comprehensively re-thought by @jossmac

jossmac:
@wmertens when we're happy with the functionality and appearance, Elemental itself will be replaced with the new components and the `import` refs replaced by the `npm` package again.

It was easier to build them in situ to make sure the API was correct for Keystone's requirements - which is why we created Elemental to begin with :)

JedWatson:
Awesome! So glad to get this landed. Thanks for all the work @jossmac and the help getting the tests over the line @jstockwin!

PR #3513, opened 2016-09-21
resolve issue #3512

### LINKED ISSUE #3512: Fields with a filter of a relationship breaks filtering in columns
`field: { type: Types.Relationship, ref: 'list2', filters: { list3: ':list3' },`

In the above case where a relationship field is being filtered by a second relationship, the column filter breaks, and is unable to retrieve the relationship's items to filter by. It tries to filter by `list3`, however the database errors as `:list3` is not a valid objectId and cannot be parsed.

The `:notation` is intended for limiting options when viewing a specific object, and should not be applied in the general column case.

PR to follow with one possible solution implementation.

Noviny:
Resolves #3512

### DIFF
--- a/fields/types/relationship/RelationshipFilter.js
+++ b/fields/types/relationship/RelationshipFilter.js
@@ -88,6 +88,7 @@ var RelationshipFilter = React.createClass({
 	buildFilters () {
 		var filters = {};
 		_.forEach(this.props.field.filters, function (value, key) {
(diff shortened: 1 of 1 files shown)


JedWatson:
I need to review #3512 more closely before merging this, will let you know how that goes

JedWatson:
Makes sense @Noviny, thanks!

PR #3540, opened 2016-09-29
Fix relationship links

### LINKED ISSUE #3528: Relationship link to linked model page is broken
In the Demo website, go to posts
http://demo.keystonejs.com/keystone/posts

Click author in anyone of those post has an author
### Expected behavior
Should link to Author/User page

### Actual behavior
Shows an `Error!`

I can't see the log for the demo, but is my own app, the API call is incorrect.

It should be 
`GET /keystone/api/users/{userId}?drilldown=true`

it is actually
`GET /keystone/api/posts/{userId}?drilldown=true 404 3.423 ms`

jstockwin:
## Description of changes

Fixes #3528. The ItemView's `componentDidMount` function didn't take into account the fact that the `currentList` will have changed when navigating from a relationship link, since the list changes from e.g. 'post' to 'user'. I've updated the `componentDidMount` function to fix this. 

I've also added additional assertions to the relationship field test, which should ensure this doesn't re-occur. (cc @webteckie if you want to have a quick look, but test passes so pretty confident this is fine). 

## Related issues (if any)
#3528 

## Testing

### DIFF
--- a/admin/client/App/screens/Item/index.js
+++ b/admin/client/App/screens/Item/index.js
@@ -39,7 +39,8 @@ var ItemView = React.createClass({
 	componentDidMount () {
 		// When we directly navigate to an item without coming from another client
 		// side routed page before, we need to select the list before initializing the item
-		if (!this.props.currentList) {
+		// We also need to update when the list id has changed
+		if (!this.props.currentList || this.props.currentList.id !== this.props.params.listId) {
 			this.props.dispatch(selectList(this.props.params.listId));
 		}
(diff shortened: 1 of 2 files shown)


jstockwin:
p.s. @webteckie Our new `modelTestConfig` stuff is working nicely here. The defaults are set at the top of the test file, but then when we navigate to the user item page, I override it in the api call. 💯

JedWatson:
Nice work @jstockwin

JedWatson:
Build errors are just saucelabs failing, so I'm going to merge it

PR #3542, opened 2016-09-29
Fix deleting items over multiple pages. Fixes #3541

### LINKED ISSUE #3541: Pagination doesn't work with deleting

Create a list with, for example, 10 items per page. Add 20 items. Go to the first page and delete an item, or even 10. 

### Expected behavior

Once an item is deleted, you should still see 10 items on the screen, since there was a whole other page of items. Once 10 items have been deleted, there should no longer be a second page. 

### Actual behavior

The item is removed from the list, but nothing from page 2 replaces it. What's worse is when you remove all 10 items you're actually told that there are no items at all, even though there's a whole other page.

jstockwin:
## Description of changes

Fix issue #3541. Makes pagination work properly when items are deleted. 

Unfortunately, the local state/props don't know anything about the items from the other pages, so we have to do a full `loadItems()` from the server to fix this. 

## Related issues (if any)

#3541 

## Testing

### DIFF
--- a/admin/client/App/screens/Item/actions.js
+++ b/admin/client/App/screens/Item/actions.js
@@ -3,8 +3,11 @@ import {
 	LOAD_DATA,
 	DATA_LOADING_SUCCESS,
 	DATA_LOADING_ERROR,
-	DELETE_ITEM,
 } from './constants';
+
+import {
+	loadItems,
+} from '../List/actions';
 
 /**
  * Select an item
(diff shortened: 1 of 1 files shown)


mxstbr:
LGTM

jstockwin:
Cool, thanks. @JedWatson?

JedWatson:
:+1: me too @jstockwin, nice work!

PR #3543, opened 2016-09-29
Fix re-ordering over multiple pages. Fixes #3494

### LINKED ISSUE #3494: Admin Pagination on a sortable list
Create a List with the 'sortable' option and 'perPage':2
`let Ticket = new Keystone.List('Ticket', {
    sortable: true,
    perPage: 5
});`

Create 20 tickets with the AdminUI

*Get 4 pages. Nice.*

Go to page number 2. 
Drag and drop to reorder an item.

Return to page 1.

*TIPS : How can i disable pagination on the admin UI ? For the moment, i just set 10000 to perPage option as workaround *
### Actual behavior
The item reordered in page 2 is now on the page 1.
It seems that the calcul, of the new sortOrder value, is based on the position in the list. 

### Idea to resolve the problem
Add a an infinite scroll (with the option perPage) in place of the pagination system.
When the user scroll down, fetch next page and display it.

Sorry for my bad english and thanks you.

jstockwin:
## Description of changes

Kudos to @mxstbr for helping out with this one. 

Reordering of pages just didn't work. There were multiple issues: 

 - Often an index was used as a sort order. This was wrong because index starts from 0 whereas sort order should start from 1. 

 - The logic when changing pages just did not work at all. It would cause lots of console errors. I think this is because of the new SPA stuff, and I don't think it's possible to get it working quite as it used to. 
Previously, you were able to drag and hover an item over a different page link, and it would drop that item into the new page for you. You now have to release the mouse (and actually drop the item onto the link) for anything to happen. 
In current master, if you try to do either, everything will completely break. 

- The index where the item dropped onto was also broken when changing pages. Again because of the +1 thing, I think. The way it works now (which I think is how was intended before) is that if you're changing from e.g. page 1 to page 2, it'll drop the item at the top of page 2. If you're going from page 2 to page 1, it'll drop it at the bottom of page 1.

- Trying to do more than one move per page refresh was broken. The local sortOrders were not updating at all, so say you move item 1 to position 2, and then change your mind and move it to position 3. The API requests should look like `change position 1 to 2` and then `change position 2 to 3`, whereas they actually looked like `change position 1 to 2` and then `change position 1 to 3`, because the sort orders hadn't updated locally. However, they had updated at the endpoint of the api call, so strange and unexpected things would happen. This was mainly due to the weird `if` clause in `dragdrop.js`. 

- Previously, as described in #3494, if you change the order on page 2, that item will actually change to a new position on page 1. This was because it was using indexes and pages and silly things. I now calculate the sortOrder properly, using the index, the current page, and the number of items on the page. This now works as it should. 

## Related issues (if any)
#3494

## Testing

### DIFF
--- a/admin/client/App/screens/List/components/ItemsTable/ItemsTableDragDropZoneTarget.js
+++ b/admin/client/App/screens/List/components/ItemsTable/ItemsTableDragDropZoneTarget.js
@@ -57,46 +57,39 @@ const dropTarget = {
 		const item = monitor.getItem();
 		item.goToPage = props.page;
 		item.prevSortOrder = item.sortOrder;
-		// if the new page is greater, we will place the item at the beginning of the page
-		// if the new page is less, we will place the item at the end of the page
+		// Work out the new sort order. If the new page is greater, we'll put it at the start of the page, and
+		// if it's smaller we'll put it at the end of the page.
 		item.newSortOrder = (targetPage < page) ? (targetPage * pageSize) : (targetPage * pageSize - (pageSize - 1));
-
 		return item;
 	},
+	/*
+	* TODO Work out if it's possible to implement this in a way that works.
 	hover (props, monitor, component) {
 		if (timeoutID) {
 			return;
 		}
-		const { page, currentPage, drag } = props;
-		const original = drag;
+		const { page, currentPage } = props;
 
 		// self
 		if (page === currentPage) {
 			return;
 		}
+
 		if (monitor.isOver()) {
 			timeoutID = setTimeout(() => {
-				const newIndex = (original.page === page) ? original.index : (currentPage < page) ? 0 : props.pageSize;
-				props.dispatch(setCurrentPage(page));
-				monitor.getItem().index = newIndex;
+				// If user hovers over the target for a while change the page.
+				// TODO Get this working. Currently, it looks like it's going to work, but when you
+				// drop onto a new page, no drop events are fired, and react-dnd doesn't have a way to
+				// manually force them to happen. Not sure what to do here.
+				props.dispatch(setCurrentPage(props.page));
+
 
 				clearTimeout(timeoutID);
 				timeoutID = false;
 			}, 750);
 		}
 	},
-	canDrop (props, monitor) {
-		// if we drop on a page target, move the item to the correct first or last position in the new page
-		// if we want to stop this behaviour set return false
-		if (!Keystone.devMode) return;
-
-		const original = props.drag;
-		// self
-		if (original.page === props.page) {
-			return false;
-		}
-		return true;
-	},
+	*/
 };
 /**
  * Specifies the props to inject into your component.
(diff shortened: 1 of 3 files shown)


admin/client/App/screens/List/components/ItemsTable/ItemsTableDragDropZoneTarget.js
```diff
@@ -57,46 +57,39 @@ const dropTarget = {
...
+
 		if (monitor.isOver()) {
 			timeoutID = setTimeout(() => {
-				const newIndex = (original.page === page) ? original.index : (currentPage < page) ? 0 : props.pageSize;
-				props.dispatch(setCurrentPage(page));
-				monitor.getItem().index = newIndex;
+				// If user hovers over the target for a while change the page.
+				// TODO Get this working. Currently, it looks like it's going to work, but when you
+				// drop onto a new page, no drop events are fired, and react-dnd doesn't have a way to
+				// manually force them to happen. Not sure what to do here.
```
mxstbr: Those this mean this part doesn't currently work?

jstockwin: So, firstly it doesn't work in current master anyway, but no - it still doesn't work. 

You used to be able to hover over the link to page two, whilst dragging an item, and the page used to change to page two, and it would have dropped the item at the start of page two. Now you will actually have to release the mouse for this to happen.

I don't actually think it's going to be possible to get this working. When the page changes the event seems to get lost, so when you do release the mouse, the drop event is never fired. What's worse is when you then go to try and drag another item, you get many many errors about the fact that you can't start dragging a new item while you're dragging an old item. 

I'll have another go at some point, but not in this PR. The current behaviour just doesn't work at all, so keen to get this merged pretty quickly, and making this little bit work is going to take quite a big effort, and might not be possible at all. 

TL;DR: No it doesn't work, but it doesn't work in master either, and may not be fixable at all. I will have another go at some point, but that'll come later.

jstockwin:
@mxstbr Are you happy with this one, or do you still think the tiny bit that's still not working is really important to have? I think given the amount this _does_ fix it's worth getting it merged?

jstockwin:
bump @mxstbr and @JedWatson Sorting not working is really messing up one of my client's app. Can we get this merged and published, or do I need to do something else?

mxstbr:
Let's shipit. I'm very ill and bed-bound, so can't review the code. @JedWatson please review and ship.

jstockwin:
Sorry to hear that. Get well soon.

I should also point out that sorting is also currently broken even when
pagination isn't involved due to the index thing, which I've fixed, so this
probably affects a lot of people.

JedWatson:
:+1:

mxstbr:
Nice new avatar @jstockwin :muscle:

jstockwin:
Thanks @mxstbr, got bored of the old picture

PR #3546, opened 2016-09-29
Re-implement management mode "Select All", fixes #3374

### LINKED ISSUE #3374: Management mode "Select all" only selects current page
I'm not 100% sure our backend would even support this at the moment, but when in management mode on a list with more than one page and clicking "Select all" it still only selects one page. (which there's a separate button for)

screen shot 2016-08-23 at 16 16 17

jstockwin:
## Description of changes

cc @mxstbr and @JedWatson, this fixes #3374. 

There's an inevitable API call in the code to list all the items. I toyed with making this call anyway and fully populating say `props.allItems`, but I figured in most cases that's a huge waste of resources (there are potentially lots of items). Instead, I make this call when the "Select All" button is clicked, and while the call is in progress I replace the text of the button with an ElementalUI Spinner. I think this is the best way to do this.

It all seems to work nicely IMO. 

## Related issues (if any)

#3374 

## Testing

### DIFF
--- a/admin/client/App/screens/List/components/ListManagement.js
+++ b/admin/client/App/screens/List/components/ListManagement.js
@@ -1,5 +1,5 @@
 import React, { PropTypes } from 'react';
-import { Button, GlyphButton, InlineGroup as Group, InlineGroupSection as Section } from '../../../elemental';
+import { Button, GlyphButton, InlineGroup as Group, InlineGroupSection as Section, Spinner } from '../../../elemental';
 
 function ListManagement ({
 	checkedItemCount,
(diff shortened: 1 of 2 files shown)


JedWatson:
This is really nice! I was going to handle it differently by adding `selectedItems: '*'` support, and then making that work with the API endpoints as well, but in the meantime I'm happy to go with this.

It'll run into trouble with lists that have, say, hundreds of thousands of items, but let's deal with that separately.

jstockwin:
@JedWatson, thanks. Your way makes sense too, although you've got to deal
with the case where I press select all, but then untick the first few
items. I think you could do it though. Currently you add a load of ids to
`checkedItems` and all of the items in that array look like `{id, true}`,
you never have a false in there. You could start adding things as false and
explicit exclude them in your `*` API bit.

Agree this works for now though.

JedWatson:
My plan for when a user starts unticking items was to flip back to "just on this page" mode, so it would go from "all" to "99 selected".

That means you can't delete everything but exclude just a few... but I also think it makes for a more robust UI because it's explicit which subset you'll be deleting.

What do you think of that?

jstockwin:
That makes sense, but also what if I want to delete all but the first few?
I think I'd personally throw a big warning about deleting items which are
not visible on the current page. But I can see pros and cons of both, so
whichever you think is best.

IVANLEE99:
it doesn't work

Issue #3551, opened 2016-10-01
WYSIWYG Editor doesn't refresh for dependsOn. (4.0.0-beta.3)


stuartforster:
If I use dependsOn with an $and or $or (or anything other than key>value) on a wysiwyg field tiny will not be refreshed on hide/show of the field.

Imagine my dependsOn looks like the below, which it needs to in order for $and to actually work:
```
dependsOn: {
  $and: { lang: 'en', type: 'layout1' }
}

// I use $and as per the docs for expression-match as the below does not seem to work properly when hiding/showing
dependsOn: {
  lang: lang, type: 'layout1'
}
```

This seems to be caused by the updated to expression-match and the inconsistency in it's use [here](https://github.com/keystonejs/keystone/blob/113703b1eb5cef6d2bcdaf868e516121402d9a49/admin/client/App/screens/Item/components/EditForm.js#L264).

### Steps to reproduce the behavior
Add two selects and two wysiwyg editors (wysiwyg set to true). Make one dependant on one combination of select values and the other on the another using a similar expression to mine above.

The wysiwyg editor is visible for the first state however if you switch the value of the select the wysiwyg editor will disappear and not come back. This seems to be because the callbacks are not getting called as it picks the keys from the wrong place. Not sure of a _quick_ way to fix myself looking at the formats allowed for expression-match.

### Expected behavior
Wysiwyg editor should stay.

### Actual behavior
It doesnt ;)

AurelioDeRosa:
I'm still experiencing this issue in beta 4

AurelioDeRosa:
I'm still experiencing this issue in beta-4.

PR #3554, opened 2016-10-02
Fix TinyMCE not showing bug in Html field when using new dependsOn format #3551

stuartforster:
Hey all,

This fixes the issue with TinyMCE not initing for fields using dependsOn with a format other than key>value. #3551 

Looks like the fix was fairly simple after all. 
Have switched the Html field to use evalDependsOn.
I've also removed the currentDependencies bit from EditForm as it doesn't appear to be used anywhere else.

Let me know what you think.

## Related issues (if any)

#3551


JedWatson:
Looks good @stuartforster, thanks!

PR #3561, opened 2016-10-04
Leverage the latest PO api refactoring.

webteckie:
## Description of changes

This change set leverages the latest e2e PO api refactoring.

## Related issues (if any)

N/A

## Testing

cc:  @jstockwin (please review, if all ok, please bump e2e dependency and ask for merge)

### DIFF
--- a/test/e2e/adminUI/tests/group002Home/uxTestHomeView.js
+++ b/test/e2e/adminUI/tests/group002Home/uxTestHomeView.js
@@ -102,11 +102,11 @@ module.exports = {
 			.click('@nameListSubmenu')
 			.waitForListScreen();
 
-		browser.adminUIListScreen
-			.click('@singleItemDeleteIcon');
+		browser.adminUIListScreen.clickDeleteItemIcon([
+			{ row: 1, column: 1 }
+		])
 
-		browser.adminUIDeleteConfirmation
-			.waitForElementVisible('@deleteButton');
+		browser.adminUIApp.waitForDeleteConfirmationScreen();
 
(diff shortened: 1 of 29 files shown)


test/e2e/adminUI/tests/group002Home/uxTestHomeView.js
```diff
@@ -102,11 +102,11 @@ module.exports = {
 			.click('@nameListSubmenu')
 			.waitForListScreen();
 
-		browser.adminUIListScreen
-			.click('@singleItemDeleteIcon');
+		browser.adminUIListScreen.clickDeleteItemIcon([
+			{ row: 1, column: 1 }
```
jstockwin: I'm still a little unsure about the `clickDeleteItemIcon` function. The delete icon is **always** in column 1 is it not?

jstockwin:
@webteckie One comment, which is to do with the other repo really, so happy to merge this anyway. 

Could you maybe also update the keystone-nightwatch-e2e version number in `package.json`, so that when I publish your PR from the other repo we can re-run the above tests and check travis passes?

test/e2e/adminUI/tests/group002Home/uxTestHomeView.js
webteckie: @jstockwin not sure and why make that assumption?  The list can be configured with options that cause other icons to show to the left of the items.  Not sure why you think this should be a special case.

jstockwin:
Alright, then just leave it. I think I've just never seen it not in the
first row.

PR #3568, opened 2016-10-05
update handler handles having no req

Noviny:
Currently if `getUpdateHandler` is passed no arguments, it will cause
the process to fail in an error, as we later check req.user and
req.files.

This change means we will continue to return undefined unless there is
a req.user ora req.files, which was the previous behaviour.

Additionally, req is set as a variable earlier, so am checking the
function-local variable.

### DIFF
--- a/lib/updateHandler.js
+++ b/lib/updateHandler.js
@@ -52,12 +52,12 @@ UpdateHandler.prototype.process = function (data, options, callback) {
 
 	// pass the user from the request by default
 	if (!options.user) {
-		options.user = this.req.user;
+		options.user = req && req.user;
 	}
 
 	// pass the uploaded files object from the request by default
 	if (!options.files) {
-		options.files = this.req.files;
+		options.files = req && req.files;
 	}
 
 	function flashErrors (err) {


bassjacob:
Does there need to be a check in the flashErrors function as well?

Noviny:
Probably, yes, although with that much bypassing, it might be worth turning req into an empty object where we are doing the same to options.

JedWatson:
There's honestly not a lot of point calling `getUpdateHandler` if you're not going to pass the request - better to use `List.updateItem` instead. @Noviny unless you disagree, I think we're better off throwing an error if the argument isn't provided.

Noviny:
I like that it is a schema method, and automatically fills in some of the details for you, as opposed to a list method where you pass the item in.

That said, since getUpdateHandler does not handle the res object, I'm not entirely certain of its purpose, so I was sort of extrapolating it as the schema method way to update items.

The usage pattern would be 
```
kitten2.getUpdateHandler().process({ name: 'latte' }, err => {
  if (err) console.log(err);
  else console.log('item updated')
```

It may be easier to do that either using `updateItem` or good old `.save()`

Noviny:
*goes back to write fifth version of the getUpdateHandler documentation as understanding shifts again*

JedWatson:
:+1:

PR #3573, opened 2016-10-06
V0.3.x

molomby:
Creates, maintains and utilises text indexes for the searchable fields on a list when 'searchUsesTextIndex' is true.


JedWatson:
We're not going to merge this into a new 0.3.x release, let's focus on making sure it's wrapped up nicely for 4.0

PR #3594, opened 2016-10-12
Support letsencrypt-express v2

wmertens:
* Now more integrated with Express and Keystone
* Make host and port defaults part of the keystone init
* Introduce ssl:force setting, with ssl public port for redirection

v0.3 is still broken, probably best to change the dependency of letsencrypt-express back to 1.x on that branch, because this PR won't easily backport.


wmertens:
@webteckie @jstockwin Any ideas for how to test this? Basically it would mean running with the letsencrypt settings set to staging and then verifying that it starts without errors and listens on the https port with the `localhost-daplie` certificate…

webteckie:
@wmertens you can update tests/e2e/server.js to add the https leg with the proper letsencrypt config.  Then add a test to the group001Login e2e group to redirect to the https leg and do a login test.

JedWatson:
:+1: Thanks @wmertens!

There's a linter error, but since this isn't in a keystone branch I'll merge and fix in master to get CI green again 😄

PR #3599, opened 2016-10-13
Don't cache relationship options, fixes #3598

### LINKED ISSUE #3598: 0.3.21 Bad Select Options for relationships depending on other relationships

Have a Model with the following fields, the idea being that 'two''s choices depend on what you selected in 'one'

` one:   { type: Types.Relationship, ref:"SomeModel", required:true, initial:true } `
` two:   { type: Types.Relationship, ref:"SomeOtherModel", required:true, initial:true, filters: { from: ':one' }}, `

Now in the admin ui if you choose a value for 'one', then go to the dropdown for 'two', the options in the dropdown lag behind what they should be, and to get the correct options you have to select a value, then unselect it, then try the dropdown again. 

### Expected behavior

you get values in the dropdown for 'two' of SomeOtherModel rows that have the 'from' field matching 'one'

### Actual behavior

you get values in the dropdown for 'two' of SomeOtherModel rows that have the 'from' field matching what 'one' was previously

LiquidHelium:
## Description of changes

disables caching on the select dropdown of relationship fields, fixing #3598
## Related issues (if any)

fixes #3598
## Testing
- [x] npm run test
- [x] npm run lint

### DIFF
--- a/fields/types/relationship/RelationshipField.js
+++ b/fields/types/relationship/RelationshipField.js
@@ -191,7 +191,7 @@ module.exports = Field.create({
 			);
 		}
 
-		body.push(<Select multi={this.props.many} onChange={this.updateValue} name={this.props.path} asyncOptions={this.getOptions} value={this.state.expandedValues} />);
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @LiquidHelium! We'll get a new 0.3 release out shortly with this fix.

PR #3639, opened 2016-10-25
Added documentation on uploading files.

blargity:
## Description of changes

Updates the README.md file for the file field type to include more detail about uploading files.

## Related issues (if any)

N/A

## Testing

- [√] Please confirm `npm run test-all` ran successfully.

(Changes are unrelated to code)

## Question since I'm new to Keystone

Is the information I put in about calling this needed in this document or is it enough to just mention that the method exists? As a new user of Keystone I wouldn't know how to actually call this method without that info, but it could make documenting everything very verbose. Maybe we should just have a sentence at the top of the method section that explains how to access field methods?

### DIFF
--- a/fields/types/file/Readme.md
+++ b/fields/types/file/Readme.md
@@ -69,6 +69,8 @@ file.updateItem(item, {
 
 ### Upload a new file (result will be stored in the field)
 
+To upload a new file, you can use either updateItem, or you can use the upload method below.
+
 ```js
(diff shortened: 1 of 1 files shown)


JedWatson:
Nice, cheers @blargity 

I think for now the extra context is worth having here; ultimately it should be covered in a higher-level "guides" section of the docs for 4.0 but we can review when more of that structure is in place.

PR #3650, opened 2016-10-29
Fixes issue where host is undefined

LorbusChris:
## Description of changes

Fixes an issue where host is localhost is expected but host is undefined

## Testing

- [x] `npm run test-all` ran successfully.


JedWatson:
Good catch, thanks @LorbusChris!

PR #3720, opened 2016-11-20
[0.4]align handling in create api with update api

gemscng:
## Description of changes
align handling in create api with update api, so that error message will be show in CreateForm if an Error object has been return from pre.save of model.


JedWatson:
Thanks @gemscng!

PR #3756, opened 2016-12-04
Update csrf.js, fixes #3420

### LINKED ISSUE #3420: Admin API: Update List Item Broken Due to CSRF Issue

I first reported this bug on [this issue thread](https://github.com/keystonejs/keystone/issues/3209). I thought it might help if I filed it as its own separate issue. Here is the summary from that thread:

I ran the test script in the [keystone-api-tests](https://github.com/christroutner/keystone-api-tests) repository while running KeystoneJS under node inspector. Keep in mind that I'm running the latest master commit (cbf9ff4).

When I ran the post(/keystone/api/users/userID) test again, the file `keystone/admin/client/utils/List.js` didn't get executed at all. Instead, `keystone/admin/server/api/item/update.js` was executed. That file was throwing a 403 error because of an invalid CSRF token. Here is the code from that file. I stopped it at the debugger statement:

```
(function (exports, require, module, __filename, __dirname) { module.exports = function (req, res) {
    var keystone = req.keystone;
    if (!keystone.security.csrf.validate(req)) {
        debugger;
        return res.apiError(403, 'invalid csrf');
    }
    req.list.model.findById(req.params.id, function (err, item) {
        if (err) return res.status(500).json({ err: 'database error', detail: err });
        if (!item) return res.status(404).json({ err: 'not found', id: req.params.id });
        req.list.validateInput(item, req.body, function (err) {
            if (err) return res.status(400).json(err);
            req.list.updateItem(item, req.body, { files: req.files }, function (err) {
                if (err) return res.status(500).json(err);
                res.json(req.list.getData(item));
            });
        });
    });
};

});

```

I manually ran `req.keystone.security.csrf.getSecret(req)` and it returned the CSRF token string "CunSnobyfWscEg==". Next I manually ran `req.keystone.security.csrf.validate(req)`, which returned false. I ran `req.keystone.security.csrf.requestToken(req)` and it returned an empty string.

During this entire time, I'm logged in and can access the Admin UI at `/keystone`, so this clearly seems to be a CSRF issue. 

Any thoughts?
### Expected behavior

Successful update of the list item. 
### Actual behavior

403 error returned due to invalid CSRF.

uhhhh2:
This is a fix for issue #3420 that I mentioned in that issue's discussion.

### DIFF
--- a/lib/security/csrf.js
+++ b/lib/security/csrf.js
@@ -48,6 +48,8 @@ exports.requestToken = function (req) {
 		return req.headers[exports.XSRF_HEADER_KEY];
 	} else if (req.headers && req.headers[exports.CSRF_HEADER_KEY]) {
 		return req.headers[exports.CSRF_HEADER_KEY];
+	} else if (req.cookies && req.cookies[exports.XSRF_COOKIE_KEY]) {
+		return req.cookies[exports.XSRF_COOKIE_KEY];
 	}
 	return '';
 };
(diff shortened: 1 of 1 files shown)


JedWatson:
thanks @uhhhh2, and @christroutner for your help with #3420 and verifying this!

PR #3761, opened 2016-12-06
File link instead of name only

BraunreutherA:
## Description of changes

A quick fix to have the filename as a link. This doesn't solve the missing format function but at least we can click the filename and download or watch the file in a new tab.

## Testing

### DIFF
--- a/fields/types/file/FileField.js
+++ b/fields/types/file/FileField.js
@@ -68,10 +68,13 @@ module.exports = Field.create({
 	hasExisting () {
 		return this.props.value && !!this.props.value.filename;
 	},
-	getFilename () {
+	getFilename (href) {
 		return this.state.userSelectedFile
 			? this.state.userSelectedFile.name
-			: this.props.value.filename;
+			: href
+				? <a href={this.props.value.url} target="_blank">{this.props.value.filename}</a>
+				: this.props.value.filename;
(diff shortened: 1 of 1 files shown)


fields/types/file/FileField.js
```diff
@@ -68,10 +68,13 @@ module.exports = Field.create({
 	hasExisting () {
 		return this.props.value && !!this.props.value.filename;
 	},
-	getFilename () {
+	getFilename (href) {
```
JedWatson: It looks like there's another part of this missing - `href` isn't being passed in from the render method...

BraunreutherA: Right sorry commited this now.

fields/types/file/FileField.js
```diff
@@ -68,10 +68,13 @@ module.exports = Field.create({
...
 		return this.state.userSelectedFile
 			? this.state.userSelectedFile.name
-			: this.props.value.filename;
+			: href
+				? <a href={this.props.value.url} target="_blank">{this.props.value.filename}</a>
+				: this.props.value.filename;
+			// : this.props.value.filename;
```
AurelioDeRosa: Apart from this line to be deleted, this PR seems good to me.

AurelioDeRosa:
Any update on this?

PR #3282, opened 2016-08-05
Nested lists

JedWatson:
### WIP

This PR adds the long-awaited feature of nested schema support for Keystone 🎉 

See keystonejs/keystone-test-project#17 for example.

To work with this, you'll need to switch over to the new API by removing both instances of `/legacy` from this file: https://github.com/keystonejs/keystone/blob/master/admin/client/utils/List.js

Yet to be done:
- [ ] Recursive input validation
- [x] Recursive `.updateItem` calls (required for field types with async updates)
- [x] Recursive `.getData()` calls (required for complex field types)
- [ ] UI is pretty basic

... and probably more.

PR #3775, opened 2016-12-09
[Nested Lists] Adds className and visual offset for list field for better UX.

jeremyzahner:
## Description of changes

Adds className and visual offset for list field for better UX.

## Related issues (if any)

* #3282 

## Testing

### DIFF
--- a/fields/types/list/ListField.js
+++ b/fields/types/list/ListField.js
@@ -118,7 +118,7 @@ module.exports = Field.create({
 	renderUI () {
 		const { label, value } = this.props;
 		return (
-			<div>
+			<div className="FormFieldList">
 				<h3 data-things="whatever">{label}</h3>
 				{this.shouldRenderField() ? (
 					this.renderItems()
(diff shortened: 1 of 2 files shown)


wmertens:
You don't seem to be following the naming scheme for the class. I'm going to defer to @jossmac @mxstbr who know better what is supposed to happen here.

jeremyzahner:
@wmertens Thanks.

fields/types/list/ListField.js
```diff
@@ -118,7 +118,7 @@ module.exports = Field.create({
 	renderUI () {
 		const { label, value } = this.props;
 		return (
-			<div>
+			<div className="FormFieldList">
 				<h3 data-things="whatever">{label}</h3>
```
jeremyzahner: @JedWatson @wmertens Is "whatever" the correct data-things binding?

wmertens: I do believe that was just for the sake of example, to get something going…

jeremyzahner: Then we probably should rename it to something less "exemplary"? =)

PR #3776, opened 2016-12-09
ListType: reset list when no value is submitted, cast objects with nu…

naptowncode:
## Description of changes

1. Reset list when no value is submitted. This allows user to empty out a list.
2. Cast objects with numeric keys to array. This works around an unchangeable setting in `multer` that causes submitted arrays of 22 items or more to be parsed as objects instead of arrays.

## Testing

…meric keys to array

### DIFF
--- a/fields/types/list/ListType.js
+++ b/fields/types/list/ListType.js
@@ -185,17 +185,19 @@ list.prototype.updateItem = function (item, data, files, callback) {
 
 	var field = this;
 	var values = this.getValueFromData(data);
-	// Don't update the value when it is undefined
-	if (values === undefined) {
-		return utils.defer(callback);
-	}
-	// Reset the value when null or an empty string is provided
-	if (values === null || values === '') {
+	// Reset the value when no value or null or an empty string is provided
+	if (values === undefined || values === null || values === '') {
 		values = [];
 	}
-	// Wrap non-array values in an array
 	if (!Array.isArray(values)) {
-		values = [values];
+		if (typeof values === 'object' && values[0] !== undefined) {
+			// Cast an object with numeric keys to an array
+			values.length = Object.getOwnPropertyNames(values).length;
+			values = Array.prototype.slice.call(values);
+		} else {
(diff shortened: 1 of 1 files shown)


fields/types/list/ListType.js
```diff
@@ -185,17 +185,19 @@ list.prototype.updateItem = function (item, data, files, callback) {
...
 	var field = this;
 	var values = this.getValueFromData(data);
-	// Don't update the value when it is undefined
-	if (values === undefined) {
-		return utils.defer(callback);
-	}
-	// Reset the value when null or an empty string is provided
-	if (values === null || values === '') {
+	// Reset the value when no value or null or an empty string is provided
+	if (values === undefined || values === null || values === '') {
```
wmertens: This is a definite change in semantics, I think it would be better to leave the `undefined` case as-is…

fields/types/list/ListType.js
```diff
@@ -185,17 +185,19 @@ list.prototype.updateItem = function (item, data, files, callback) {
...
 		values = [];
 	}
-	// Wrap non-array values in an array
 	if (!Array.isArray(values)) {
-		values = [values];
+		if (typeof values === 'object' && values[0] !== undefined) {
```
wmertens: I think you should use `hasOwnProperty` in this instance, although for this particular use case, I suppose array values should always be at least `null`? Not sure.

fields/types/list/ListType.js
naptowncode: Sorry, I was cheating here to fix two bugs at once.

With the `undefined` case as it is there is no way to empty out the list. The last item will never be deleted because an empty list won't be submitted at all. TextArray, DateArray, and NumberArray all treat undefined the way I've done here.

fields/types/list/ListType.js
naptowncode: Good catch, `hasOwnProperty` is probably more correct. Thanks!

PR #3793, opened 2016-12-14
Fix error building FieldTypes on Linux

### LINKED ISSUE #3754: Error building FieldTypes
I'm currently receiving this error when using Keystone:

> Cannot find module '../../fields/types/resource/resourceColumn' from 'XXXXXXXX/node_modules/keystone/admin/client'

1. Access the admin panel

### Expected behavior

To see the admin panel

### Actual behavior

The `fields.js` file remains pending and the panel is never shown.

AurelioDeRosa:
## Description of changes

Set the first letter of the field to uppercase

## Related issues (if any)

Fixes #3754

## Testing

### DIFF
--- a/admin/server/app/createStaticRouter.js
+++ b/admin/server/app/createStaticRouter.js
@@ -11,6 +11,7 @@ var express = require('express');
 var less = require('less-middleware');
 var path = require('path');
 var str = require('string-to-stream');
+var _ = require('lodash');
 
 function buildFieldTypesStream (fieldTypes) {
 	var src = '';
@@ -19,7 +20,7 @@ function buildFieldTypesStream (fieldTypes) {
 		src += 'exports.' + i + 's = {\n';
 		types.forEach(function (type) {
 			if (typeof fieldTypes[type] !== 'string') return;
-			src += type + ': require("../../fields/types/' + type + '/' + fieldTypes[type] + i + '"),\n';
+			src += type + ': require("../../fields/types/' + type + '/' + _.upperFirst(fieldTypes[type]) + i + '"),\n';
 		});
 		// Append ID and Unrecognised column types
 		if (i === 'Column') {


jstockwin:
Would it be worth instead using the `keyToLabel` function from `keystone-utils` here? Or are the two operations not quite the same? Seems to me that we should use the utils functions where possible, as they should be pretty good at handling all possible cases correctly. 

https://github.com/keystonejs/keystone-utils/blob/master/lib/index.js#L636

AurelioDeRosa:
As far as I can tell, only the first letter needs to be uppercased, but you surely know the app better than I do.

jstockwin:
Haha, don't be too sure of that. You're probably right.

jstockwin:
We have to end up with camel-cased file names right, e.g. "CloudinaryImage", so I guess it depends what the input is. I'm sure your way works fine, but I think using the utils function should too. 

I guess it doesn't really matter that much anyway though, so maybe just leave it.

AurelioDeRosa:
Cool then :)

JedWatson:
We should actually be using the `properName` property on the node implementation for the Field Type. cc/ @bassjacob

AurelioDeRosa:
Let me know how to update the PR to have it landed.

jstockwin:
Does `fieldTypes[type].properName` work?

pavethiran:
I don't think `fieldTypes[type].properName` would work.

My `fieldTypes` looks like this:
```JavaScript
{
  boolean: 'Boolean',
  number: 'Number',
  text: 'Text',
  select: 'Select',
  file: 'File',
  color: 'Color',
  grid: 'Grid',
  datetime: 'Datetime',
  relationship: 'Relationship',
  textarea: 'Textarea',
  name: 'Name',
  email: 'Email',
  password: 'Password'
}
```
Grid is a Custom-Fieldtype, so it can be ignored.

PR #3800, opened 2016-12-15
HTMLField: Interpret markup when showing value

### LINKED ISSUE #2926: [v0.4] HTML field with noedit: true and wysiwyg: true displays source code instead of rendered HTML
### Expected behavior
A field that has the following options should just render HTML into the admin UI:
```
{
    type: Types.Html,
    noedit: true,
    wysiwyg: true
}
```
### Actual behavior
A grey box is shown that contains the HTML source stored in the field.

Create a model with a field that has the following options:
```
{
    type: Types.Html,
    noedit: true,
    wysiwyg: true,
    default: '<a href="#">Test</a>'
}
```
Create a new item of that model and open it's edit page. Click on "Test" inside the field's contents. Nothing happens because it's not a link.

I know, this borders on a feature request, but please bear with me. It just dawned on me that this could enable a very basic form of admin UI customization, especially if used in combination with a save hook. It is probably very easily implemented as well. And it is indeed the behaviour I would expect, because you know... what you see is what you get!

AurelioDeRosa:
## Description of changes

Interpret the markup passed to the field

## Related issues (if any)

Fixes #2926

## Testing


fields/types/html/HtmlField.js
```diff
@@ -204,10 +204,12 @@ module.exports = Field.create({
 	},
 
 	renderValue () {
+		const htmlValue = {
+			__html: this.props.value
```
jstockwin: @AurelioDeRosa The [build is failing](https://travis-ci.org/keystonejs/keystone/jobs/184122696#L508) due to a linting error here - you're missing the trailing comma.

You can check for lint errors by running `npm run lint`, and probably fix it by running `npm run lint-fix`. I thought this was included in `npm run test-all`, but I guess not.

AurelioDeRosa:
Issue fixed.

AurelioDeRosa:
Does this PR look good?

jstockwin:
Looks fine to me, but @JedWatson or @mxstbr will need to look at it before it gets merged.

bassjacob:
Do we want to consider XSS attacks as part of this, given there's a stated intention to add roles to KS which would allow non-admin users to access the same ui as admin users.

JedWatson:
I think the right way to address this is to allow developers to opt-in to this feature. There is all sorts of stuff in HTML that could really mess with the Admin UI, so it's (a) a dangerous feature from a few different perspectives, and (b) definitely not something that should be on by default.

I wonder if there's a safer way to handle this by dropping in an iFrame or something to contain the HTML value...

AurelioDeRosa:
We can enable this via the options passed to Keystone if that's what you had in mind.

webteckie:
There are packages out there (eg, https://www.npmjs.com/package/sanitize-html) to sanitize html.

JedWatson:
Given that this is only going to happen when the `noedit: true` option is set, I think it makes sense to add another option for `HTML` fields `previewHTMLValue` which would invoke this behaviour.

I also think we need to add a `sanitize` feature as @webteckie suggests if we're going to merge this - @AurelioDeRosa if you're happy to add this into the PR then I'm happy to sign off on it.

Spec:

- new `previewHTMLValue` option for `HTML` fields that enables this feature
- new `sanitize` option for `HTML` fields that adds a `pre('save')` hook to the model for the HTML field, that passes the value to be saved through the `sanitize-html` package before committing the new value to the database.

If the `sanitize` option is an object, it should be passed to `sanitize-html` as the options argument so that the sanitisation can be customised. Otherwise if it's a simple truthy value, we should just use the default options.

With all that in place, I think this is going to be a really good addition to Keystone - feel free to ping me if you want any more guidance on how to implement it!

PR #3806, opened 2016-12-16
Fixes Relationship is deleted when updating and the field is noedit: …

### LINKED ISSUE #3802: Relationship is deleted when updating and the field is noedit: true

1. Create a model with a Relationship field having `noedit: true`
2. Create a document of such model. Be sure to set the relation of the field having `noedit: true`
3. Update the document

### Expected behavior

The relation field doesn't change

### Actual behavior

The relation field is unset

As far as I can tell, the reason is that the form doesn't submit the `noedit: true` field (and this is correct). So, when calling the `updateItem()` function, the following code (file `RelationshipType.js`) is executed:

```js
} else if (item.get(this.path)) {
    item.set(this.path, null);
}
```

This unset the value. I think the `if` should also check that the field is `!noedit`.

AurelioDeRosa:
## Description of changes

This fix has worked for my use case but I think that it's the correct solution. Feel free to comment if you disagree and I'll update the PR.

## Related issues (if any)

Fixes #3802

## Testing

### DIFF
--- a/fields/types/relationship/RelationshipType.js
+++ b/fields/types/relationship/RelationshipType.js
@@ -236,7 +236,7 @@ relationship.prototype.updateItem = function (item, data, callback) {
 			if (data[this.path] !== item.get(this.path)) {
 				item.set(this.path, data[this.path]);
 			}
-		} else if (item.get(this.path)) {
+		} else if (item.get(this.path) && !this.noedit) {
 			item.set(this.path, null);
 		}
 	}


JedWatson:
Unless I'm mistaken, `noedit` is being handled higher up and `relationship.prototype.updateItem` should never be called by default. If that's not the case, that's the bug to fix here.

We retain the ability to update `noedit` fields through `List.updateItem` by explicitly specifying them in the `fields` option, which this would break.

Example use-case: a field is, in most instances including the Admin UI, not editable, but there's a specific form or process written in the application through which that field _does_ get set. For example, a set-once field in a signup form that then can't be changed through the Admin UI.

Noviny:
Closing this in favour of #4035 which fixed a slightly broader problem. If you're still finding this occurring, chasing up on Jed's comment is probably a good idea.

AurelioDeRosa:
Sounds good

PR #3812, opened 2016-12-19
WYSIWYG Editor doesn't refresh for dependsOn

### LINKED ISSUE #3551: WYSIWYG Editor doesn't refresh for dependsOn. (4.0.0-beta.3)

AurelioDeRosa:
## Description of changes

Show WYSIWYG Editor when it depends on a dependsOn field. Ensure no other WYSIWYG Editors are corrupted under this condition.

## Related issues (if any)

Fixes #3551
Related #3554

## Testing

### DIFF
--- a/fields/types/html/HtmlField.js
+++ b/fields/types/html/HtmlField.js
@@ -54,7 +54,9 @@ module.exports = Field.create({
 
 		this._currentValue = this.props.value;
 		tinymce.init(opts);
-		this.setState({ wysiwygActive: true });
+		if (evalDependsOn(this.props.dependsOn, this.props.values)) {
+			this.setState({ wysiwygActive: true });
+		}
 	},
 
 	removeWysiwyg (state) {


webteckie:
@AurelioDeRosa this looks good.  Can we add an e2e test for this?  Were you able to run the e2e test locally?

JedWatson:
Good fix, @AurelioDeRosa - thanks!

@webteckie let's follow up adding a test separately, I'd like to land this.

PR #3828, opened 2016-12-29
Allow HTTP/2 with letsencrypt

wmertens:
For some reason, `letsencrypt` forces the `NPNProtocols` option, which inhibits `spdy` from offering HTTP/2.


wmertens:
Good point. I'll push the button so it works, and I'll open an issue.

PR #3831, opened 2016-12-31
Allow pre:admin configuration.

gemscng:
## Description of changes
Allow pre:admin configuration. This allow more flexibility to customise admin UI flow.
For example we want to limit the api request body size of admin UI, we will need to create our own express and pass to keystone. Although this work but not the best way to do it. pre:admin middleware configuration will allow more precise setup.

### DIFF
--- a/server/createApp.js
+++ b/server/createApp.js
@@ -83,6 +83,9 @@ module.exports = function createApp (keystone, express) {
 	// unless the headless option is set (which disables the Admin UI),
 	// bind the Admin UI's Dynamic Router
 	if (!keystone.get('headless')) {
+		if (typeof keystone.get('pre:admin') === 'function') {
+			keystone.get('pre:admin')(app);
+		}
 		app.use(function (req, res, next) {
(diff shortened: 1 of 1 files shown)


bassjacob:
Hi @gemscng, thanks for the contribution.

In the specific case of body size, keystone does allow you to control that via: `keystone.set('file limit', '1mb');` which is called [here](https://github.com/keystonejs/keystone/blob/master/server/bindBodyParser.js#L7)

Other than that, I'm not sure that there's a large difference between passing in an express instance vs overriding one. 

@jed do you have an opinion here? It does strike me as a possible issue that keystone's bodyparser applies at the root middleware, which means you can't have a smaller limit for the keystone ui and a larger one for the general app.

gemscng:
@bassjacob you are correct. Haven't aware that there are already method to do this.

However, we still have another use case. 
For example, we want have add our own route to admin UI with our own Front End Component.

JedWatson:
This is consistent with the rest of the hooks in `createApp` so I'm happy to merge.

Thanks @gemscng!

PR #3836, opened 2017-01-03
Add filters to url params

Noviny:
Filter options were not being passed through to the url in Keystone 4.

This PR makes sure the url is updated, and that initial parameters on page load are used to make the initial query, making sharing links in the admin ui easier, and refreshing pages keep filter state.


admin/client/App/screens/List/index.js
```diff
@@ -104,6 +105,14 @@ const ListView = React.createClass({
 				case 'sort':
 					this.props.dispatch(setActiveSort(query[key]));
 					break;
+				case 'filters':
+					try {
+						const filters = JSON.parse(query[key]);
+						setActiveFilters(filters);
+					} catch (e) {
+						console.warn('Invalid filter provided');
```
bassjacob: is there no way to surface this in the ui? maybe using a flash message or a popup or something?

admin/client/App/screens/List/actions/active.js
```diff
@@ -73,8 +73,18 @@ export function clearAllFilters () {
 	};
 }
 
+// This is being used on first page load to set all filters from params
+export function setActiveFilters (filters) {
```
mxstbr: Why do this in a thunk and not a saga?

admin/client/App/screens/List/index.js
mxstbr: I'm not sure we really need to surface this in the UI, but I wouldn't mind showing a short flash message.

admin/client/App/sagas/index.js
```diff
@@ -8,8 +8,9 @@ import { loadItems } from '../screens/List/actions';
 /**
  * Load the items
```
mxstbr: I don't understand why this function was renamed at all, it currently does two things:

1. Load the items
2. Update the query params

Both of which imo aren't expressed with `filterTheItems`. This should really be called `handleListChange` or something like that?

admin/client/App/screens/List/actions/active.js
Noviny: It's doing the same work as other reads from params, which are all being done in sagas, so I was following the established pattern.

admin/client/App/screens/List/index.js
Noviny: Thinking was that the use-case was obscure enough that it did not need to be surfaced. Other example was that the existing redux action for an invalid action was to warn in the console and apply no filter.

admin/client/App/sagas/index.js
Noviny: Looking at this, I'd probably argue dividing it, leaving the old `loadTheItems` for `actions. INITIAL_LIST_LOAD`, and a second one for filter actions, which does what the new function does. Don't have a strong name for it though.

admin/client/utils/queryParams.js
```diff
@@ -45,3 +45,11 @@ export function stringifyColumns (columns, defaultColumnPaths) {
 	if (columnString === defaultColumnPaths) columnString = undefined;
 	return columnString;
 }
+
+export function parametizeFilters (filterArray) {
```
Noviny: Yes, which implies we should move the function that reads these back to be loaded in to state should also be testable, since currently we need the two to work in harmony more than we need them to appear a specific way.

admin/client/App/sagas/index.js
mxstbr: That sounds reasonable to me, what about `filtersToParams` or something like that?

admin/client/utils/queryParams.js
bassjacob: Sounds like a plan. Would you like a hand? Or are you comfortable with the changes that would need to be made there?

admin/client/App/sagas/index.js
Noviny: I *think* that the filter methods actually want to use the `updateParams` method, as that performs a `loadItems` as its last step. No need to run both. in a new saga action.

Noviny:
@bassjacob Tests for setActiveFilters still incoming, however review of changes would be appreciated. :)

JedWatson:
Looks great @Noviny 

There's a bug though - the history integration isn't working properly; if you add a filter, navigate, then click back it won't have kept the filter value.

If you can fix that, this is good to merge :+1:

JedWatson:
@mxstbr can you confirm you're OK with this now & update your review please?

bassjacob:
Approving this. There is a body of work that fixes the history bug being spiked by @gwyneplaine at the moment that will fix the underlying issue and reorganise the query param parsing.

PR #3842, opened 2017-01-04
write keystone bundles to disk

bassjacob:
presented for comment

this will expose a keystone config option that will allow a developer to
tell keystone to write (admin.js, fields.js, signin.js, packages.js) to
a file in the static directory

this means that keystone won't need to recompile these files between
server boots. this brings the boot time down from 20s on every launch to
around 3-5s if the files are already compiled and available. this
additional mode is compatible with the previous strategy of storing each
bundle in a variable in the closure which it will still do in order to
enable the watchify workflow.

if the user does not enable this mode, it only uses the previous
strategy.

- [*] eslint passed
- [*] npm test passed

I still need to get e2e to run on my machine.

### DIFF
--- a/admin/server/middleware/browserify.js
+++ b/admin/server/middleware/browserify.js
@@ -26,19 +26,37 @@ function logError (file, err) {
 	console.log(ts() + chalk.red('error building ' + chalk.underline(file) + ':') + '\n' + err.message);
 }
 
-module.exports = function (file, name) {
+module.exports = function (opts) {
+	var stream = opts.stream;
+	var expose = opts.expose;
+	var file = opts.file;
+	var hash = opts.hash;
+	var writeToDisk = opts.writeToDisk;
+
 	var b;
 	var building = false;
 	var queue = [];
-	var ready;
 	var src;
-	var logName = typeof file === 'string' ? file.replace(/^\.\//, '') : name;
+	var etag;
+
+	var logName = file.replace(/^\.\//, '');
 	var fileName = logName;
-	if (fileName.substr(-3) !== '.js') fileName += '.js';
+	var outputFilename = path.resolve(path.join(__dirname, '../../bundles/js', hash + '-' + fileName));
+
+	function updateBundle (newSrc) {
+		src = newSrc;
+		etag = crypto.createHash('md5').update(src).digest('hex').slice(0, 6);
+	}
+
 	function writeBundle (buff) {
-		if (devWriteBundles) {
-			fs.outputFile(path.resolve(path.join(__dirname, '../../bundles/js', fileName)), buff, 'utf8');
+		if (devWriteBundles || writeToDisk) {
+			fs.outputFile(outputFilename, buff, 'utf8', function (err) {
+				if (err) {
+					return logError(fileName, err);
+				}
+			});
 		}
+
 		if (devWriteDisc) {
 			var discFile = fileName.replace('.js', '.html');
 			require('disc').bundle(buff, function (err, html) {
@@ -51,6 +69,7 @@ module.exports = function (file, name) {
 			});
 		}
 	}
+
 	function build () {
 		if (building) return;
 		building = true;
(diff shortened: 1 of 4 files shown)


bassjacob:
This is not production ready, I'm just putting it up for comment and some bikeshedding. I've validated that it works with the test-project and it looks good on my machine.

@JedWatson this follows the strategy we were discussing a couple of weeks ago. I think this is more effective than reducing the time it takes to browserify each bundle.

bassjacob:
Ok, outside of the work in this PR so far:

* **Gzip Compression** the compression by express is pretty slow. not sure if that's addressable here, or if it just requires some clever caching and maybe use of nginx/cdn.

* **Bundle Size** the bundles are oversized.
  * is it worth looking into cdn'ing our dependencies on the FE? there's really no reason to be bundling react by default
  * we can't exclude react from out packages bundle and cdn it at the moment because of how we're using it in a few packages

* **keystone.min.css/less** why aren't we caching the keystone.min.css? it takes around 15s on my brand new machine to bundle it but it doesn't change in between server restarts unless keystone changes. less lets you ship a cacheFile.json somewhere on your machine to to help cache between server restarts.

* **postinstall/bundle on deploy** given that packages.js and keystone.min.css really don't change between keystone versions, should they be built by a postInstall script and then looked for on boot and built if they don't exist?

admin/server/middleware/browserify.js
```diff
@@ -1,3 +1,4 @@
+var keystone = require('../../../index');
```
bassjacob: I'd like to refactor this file and add unit tests to it. now that it's being used a bit more complicatedly I think it's worth protecting.

admin/server/middleware/browserify.js
```diff
@@ -35,7 +36,16 @@ module.exports = function (file, name) {
 	var logName = typeof file === 'string' ? file.replace(/^\.\//, '') : name;
 	var fileName = logName;
 	if (fileName.substr(-3) !== '.js') fileName += '.js';
+
+	var writeToDisk = keystone.get('write keystone files to disk');
```
bassjacob: is this a bit lengthy for a config var?

admin/server/middleware/browserify.js
```diff
@@ -35,7 +36,16 @@ module.exports = function (file, name) {
...
+	var staticPath = keystone.get('static') || path.join(keystone.get('module root'), 'public');
```
bassjacob: why isn't there a default for the static folder? can we get away with using one? or should we default to storing some static files inside the keystone repo (gitignored) if the user doesn't declare a static folder?

admin/server/middleware/browserify.js
```diff
@@ -35,7 +36,16 @@ module.exports = function (file, name) {
...
+	var outputPath = path.resolve(path.join(staticPath, 'keystone', fileName));
+
 	function writeBundle (buff) {
+		if (writeToDisk) {
+			fs.outputFile(outputPath, buff, 'utf8');
```
bassjacob: doesn't need to use a callback since the value of the buffer is attached to the closure variable `src` already by this point, but this still makes me uncomfortable. I'll add a cb just to check the error value since it happens during production, not just dev.

admin/server/middleware/browserify.js
```diff
@@ -67,17 +77,24 @@ module.exports = function (file, name) {
 		if (devWriteDisc) {
 			opts.fullPaths = true;
 		}
-		if (name) {
+		if (/packages\.js/.test(outputPath)) {
```
bassjacob: this is one of the complexities I'd like to refactor out. but maybe packages is complex enough that it can just live by itself rather than reusing this file?

admin/server/middleware/browserify.js
```diff
@@ -100,12 +117,18 @@ module.exports = function (file, name) {
 		});
 	}
 	function serve (req, res) {
-		if (!ready) {
-			build();
-			queue.push([req, res]);
-			return;
-		}
-		send(req, res);
+		fs.readFile(outputPath, (err, data) => {
```
bassjacob: can use this regardless of the user enabling the new "write to file" workflow since if they haven't the file shouldn't exist. not optimal performance (one file read that will fail with another predicate check) but hardly going to be the largest issue in this workflow.

admin/server/middleware/browserify.js
JedWatson: go with `cache admin bundles`, defaults to true

JedWatson:
Great work @bassjacob! This makes Keystone *much* faster after first boot 😀

Issue #3844, opened 2017-01-05
TinyMCE Error: remove is not a function


pavethiran:
### Steps to reproduce the behavior
Use Types.HTML in a List.

### Expected behavior
No Error on change

### Actual behavior
Error on every change on the page

JedWatson:
As per my comment on #3845 I can't replicate this

thejameskyle:
Closing since this has sat for awhile without reproduction.

PR #3845, opened 2017-01-05
bugfix(types.html): add check if remove function exists

pavethiran:
## Description of changes
Added check for remove function

## Related issues (if any)
#3844 

## Testing

### DIFF
--- a/fields/types/html/HtmlField.js
+++ b/fields/types/html/HtmlField.js
@@ -19,7 +19,7 @@ function getId () {
 function removeTinyMCEInstance (editor) {
 	var oldLength = tinymce.editors.length;
 	tinymce.remove(editor);
-	if (oldLength === tinymce.editors.length) {
+	if (tinymce.editors.remove != undefined && oldLength === tinymce.editors.length) {
 		tinymce.editors.remove(editor);
 	}
 }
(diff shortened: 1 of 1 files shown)


jstockwin:
@pavethiran the build is failing due to a linting error. You've got a `!=` instead of a `!==`. You should be able to check the linting by running `eslint .`. Would you mind fixing that? Thanks!

bassjacob:
Looking at the tinymce docs `remove` should always exist, no? What are the circumstances where `remove` is no longer a function/how can we reproduce this?

pavethiran:
@bassjacob To reproduce, just make a List containing a "Types.Html"-field.

This is the setting from my List.
```javascript
embed: {
  type: Types.Html,
  wysiwyg: true,
  required: false,
  initial: false,
  dependsOn: {
    'fullscreen.type': 'embed'
  }
},
```

JedWatson:
@pavelvlasov I can't replicate this using latest keystone linked to the [`keystone-test-project`](https://github.com/keystonejs/keystone-test-project)

Can you provide some more info? I added a `dependsOn` setting to the second `HTML` field (`content.extended`) in the `Post` model, and adding / removing editors worked just fine.

Tested on Safari Version 10.0.2 (macOS Sierra) with the same version of TinyMCE included by default in Keystone's Admin

bassjacob:
@pavethiran I think github completed the wrong username in Jed's message. Are you able to provide some more info for the context of this PR, we aren't able to reproduce the error. A project with the error occurring would be the best example if possible.

PR #3856, opened 2017-01-06
Manually launch sauce connect

jstockwin:
## Description of changes

Hopefully, this should start SauceConnect manually, rather than relying on travis. It enables useful things like being able to retry the connection if it fails! Let's see if travis passes. cc:// @webteckie @bassjacob 

If Travis starts playing nicer with Sauce Connect then maybe we'll switch back to the old way.

## Testing

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,3 +83,5 @@ addons:
     - mongodb-3.2-precise
     packages:
     - mongodb-org-server
+  jwt:
+    secure: EDmMTTPbjAwagS9jR5G9yMa7cyO3gdNuQGQ53NEcBKbM7CfUq+mtlyF1XGaLo8n9VseRgtw2RbxqztUgER3irlWpQu82t6DIJylYJpGy2oiAFimN2xp94p3HGatuCF4BMkRVBUvbtvmfFpZf4HKT5nup1MGvKVrUyZuQPObm9rQ=
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ node_js:
 - '6'
 env:
   global:
+    - SAUCE_USERNAME=jstockwin1
     - KEYSTONE_PREBUILD_ADMIN=true
   matrix:
     - JOB=lint-and-test-cov
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,43 +9,31 @@ env:
 matrix:
   include:
     - node_js: '6'
-      addons:
-        sauce_connect: true
       script:
         - npm run test-e2e-saucelabs-group
       env:
         - JOB=e2e GROUP=group006Fields
     - node_js: '6'
-      addons:
-        sauce_connect: true
       script:
         - npm run test-e2e-saucelabs-group
       env:
         - JOB=e2e GROUP=group005Item
     - node_js: '6'
-      addons:
-        sauce_connect: true
       script:
         - npm run test-e2e-saucelabs-group
       env:
         - JOB=e2e GROUP=group004List
     - node_js: '6'
-      addons:
-        sauce_connect: true
       script:
         - npm run test-e2e-saucelabs-group
       env:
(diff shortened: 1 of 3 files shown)


jstockwin:
To address: E2E normally doesn't run if the job is a PR due to security issues. Have I broken this? If so, it'll be an easy fix, but it needs doing!

jstockwin:
It seems to be working. Also [this line](https://github.com/keystonejs/keystone/pull/3856/files#diff-1a4e05a3b3fe0ce6e9820094a87abf88R174) might be an absolute godsend, even if I do say so myself. 
cc @mxstbr and @JedWatson to 🎉

jstockwin:
Regarding PR Security, we could also implement this instead: https://docs.travis-ci.com/user/jwt

This would actually be super helpful, as E2E will run on PRs from external forks. 

Thoughs @webteckie? @JedWatson Would you be happy with this?

jstockwin:
Keen to push this through if people are happy. There are loads of greenkeeper PRs that are read, purely because sauce connect failed. I think half of it is that travis is using an outdated version of sauceconnect....

jstockwin:
Updated title to [DO NOT MERGE] until PR security issue is resolved. I think the JWT thing would work nicely and has the added benefit of running e2e on PRs from external clones. 

@JedWatson Would you mind trying it out: run the following locally, with the correct access key etc. I think you'll need the travis gem installed to have this work properly though. I've got in installed, but don't know the access key.
```
travis encrypt --add addons.jwt SAUCE_ACCESS_KEY=your-access-key
```

jstockwin:
Okay, I've done the JWT stuff so this should be good to go. Once it's merged (or maybe a while after), we should remove the `SAUCE_ACCEESS_KEY` environment variable from online travis settings.

.travis.yml
```diff
@@ -95,3 +83,5 @@ addons:
     - mongodb-3.2-precise
     packages:
     - mongodb-org-server
+  jwt:
+    secure: EDmMTTPbjAwagS9jR5G9yMa7cyO3gdNuQGQ53NEcBKbM7CfUq+mtlyF1XGaLo8n9VseRgtw2RbxqztUgER3irlWpQu82t6DIJylYJpGy2oiAFimN2xp94p3HGatuCF4BMkRVBUvbtvmfFpZf4HKT5nup1MGvKVrUyZuQPObm9rQ=
```
bassjacob: Should this be a travis envvar rather than a secret checked into source control?

jstockwin: @bassjacob This is encrypted, see https://docs.travis-ci.com/user/jwt. 

It's the result of running `travis encrypt --add addons.jwt SAUCE_ACCESS_KEY=your-access-key` with the correct key.

jstockwin: The idea being that this way is also secure for PRs from external forks, and so we can run e2e on these, which we can't currently do.

bassjacob: Ah, got it, thanks :smile: that's an interesting POV from travis of security vs usefulness. cool :tada:

jstockwin:
So, I've done some reading and the encrypted variable in `travis.yml` will override the one set in the web interface, therefore there are no security concerns about people gaining access to our saucelabs account from malicious external PRs.

bassjacob:
So the only concerns now are if a PR can log out a private env_var or something.

jstockwin:
If the PR logs `SAUCE_ACCESS_KEY` they'll get a JWT code which is only valid for 90 minutes. Then they'll be locked out again, big deal. That's the whole reason JWT exists.

jstockwin:
All other env_vars will remain private in external PRs. I've not told travis to stop doing this, the only thing I've done is made `SAUCE_ACCESS_KEY` available in a secure way with the help of JWT.

jstockwin:
Superceeded by #3858

PR #3891, opened 2017-01-14
Disable test functionality that does not seem to work well with latest browsers

webteckie:
## Description of changes

Disable test functionality that does not seem to work well with latest browsers.  This should be ran with the next version of KNE.

## Testing


bassjacob:
This works for me on linux. :+1:

PR #3896, opened 2017-01-16
formate time

xiaoTuiMao:
sometime we want to formate time

### DIFF
--- a/fields/types/date/DateColumn.js
+++ b/fields/types/date/DateColumn.js
@@ -14,7 +14,7 @@ var DateColumn = React.createClass({
 		const value = this.props.data.fields[this.props.col.path];
 		if (!value) return null;
 
-		const format = (this.props.col.type === 'datetime') ? 'MMMM Do YYYY, h:mm:ss a' : 'MMMM Do YYYY';
+		const format = this.props.col.field.formatString || ((this.props.col.type === 'datetime') ? 'MMMM Do YYYY, h:mm:ss a' : 'MMMM Do YYYY');
 		return moment(value).format(format);
 	},
 	render () {


xiaoTuiMao:
hello

jstockwin:
Hmm, so given `this.props.col.field.formatString` has a default of `const DEFAULT_FORMAT_STRING = 'Do MMM YYYY';` (set [here](https://github.com/xiaoTuiMao/keystone/blob/546f956e7284ac975120bb5f6f1ae2d92d8fce80/fields/types/date/DateField.js#L17)), you're never going to reach the `'MMMM Do YYYY'` bit of this code. Could this be neater, and easier to read as the following?
```
const format = (this.props.col.type === 'datetime') ? 'MMMM Do YYYY, h:mm:ss a' : this.props.col.field.formatString;
```

Note, however, that this will actually be changing the default displayed format from `MMMM Do YYYY,` to `Do MMM YYYY`, unless the `DEFAULT_FORMAT_STRING` is updated. 

@JedWatson / @bassjacob Is there a reason for this discrepancy? Is there a reason we're not currently formatting the date here?

xiaoTuiMao:
some time 
```js
model.add({
      expireDate: { type: types.Datetime, label: '* expireTime', format: 'YYYY-MM-DD HH:mm:ss', initial: true }, 
})
```
i want to formate time `YYYY-MM-DD HH-mm-ss` in view when i has set in model

by the way can you speak chinese

jstockwin:
No I don't speak Chinese, I'm afraid. 
So what you actually want to do is format the `datetime` field, which I don't think this PR achieves?

PR #3899, opened 2017-01-16
Update package.json

yedaodao:
Move keystone-email to deps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,8 @@
     "string-to-stream": "1.1.0",
     "vkey": "1.0.1",
     "watchify": "3.7.0",
-    "xhr": "2.3.1"
+    "xhr": "2.3.1",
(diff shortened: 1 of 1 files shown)


bassjacob:
Hi @yedaodao,

Thanks for making this PR :smile: This is actually by design though. We don't include keystone-email as a dependency since not every installation needs to send emails. The call site for the email functionality will [check if keystone-email is required when the application boots and throw an error if it isn't installed](https://github.com/keystonejs/keystone/blob/master/lib/email.js#L74).

It isn't the cleanest pattern but it works quite well towards slimming down the core keystone install size. We're experimenting with alternative plugin strategies and definitely open to discussing different ways of achieving this.

I'm going to close this PR with that in mind, but please feel free to reopen it or to continue the discussion if you think this should still be merged.

PR #3906, opened 2017-01-17
Replacing DB connection error event listener; allows recovery after DB connection failure

molomby:
## Description of changes

Replaces the `.once()` listener for the Mongoose connection 'error' event with a sensible, permanent listener. This provides logging of failures and causes the thread to fail properly if the DB connection drops and cannot be reestablished automatically. 

The previous listener created an issue whereby a failed DB connection could put Keystone into a state where request without DB interaction would work as normal but request requiring DB interaction would hang until express timed them out. This state would occur without any error being logged and, because the app hadn't actually crashed, thread monitoring tools (ie. forever, pm2, etc.) wouldn't automatically restart it.

## Related issues (if any)

Can't find any git issues but seen real-world issues from this. Has been reported in the Keystone Slack.

## Testing

`npm run test-all` ran successfully.

### DIFF
--- a/lib/core/openDatabaseConnection.js
+++ b/lib/core/openDatabaseConnection.js
@@ -45,7 +45,7 @@ module.exports = function openDatabaseConnection (callback) {
 
 	keystone.mongoose.connection.on('error', function (err) {
 
-		// The DB connection has been established previously and this a ValidationError cased by restrictions Mongoose is enforcing on the field value
+		// The DB connection has been established previously and this a ValidationError caused by restrictions Mongoose is enforcing on the field value
 		// We can ignore these here; they'll also be picked up by the 'error' event listener on the model; see /lib/list/register.js
 		if (mongoConnectionOpen && err && err.name === 'ValidationError') return;
 
--- a/lib/core/openDatabaseConnection.js
+++ b/lib/core/openDatabaseConnection.js
@@ -43,20 +43,24 @@ module.exports = function openDatabaseConnection (callback) {
 		keystone.mongoose.connect(keystone.get('mongo'), keystone.get('mongo options'));
 	}
 
-	keystone.mongoose.connection.once('error', function (err) {
+	keystone.mongoose.connection.on('error', function (err) {
 
-		if (keystone.get('logger')) {
-			console.log('------------------------------------------------');
-			console.log('Mongo Error:\n');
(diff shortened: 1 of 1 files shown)


lib/core/openDatabaseConnection.js
```diff
@@ -43,21 +43,25 @@ module.exports = function openDatabaseConnection (callback) {
...
 
-	keystone.mongoose.connection.once('error', function (err) {
+	keystone.mongoose.connection.on('error', function (err) {
 
-		if (keystone.get('logger')) {
-			console.log('------------------------------------------------');
-			console.log('Mongo Error:\n');
-			console.log(err);
-		}
+		// The DB connection has been established previously and this a ValidationError cased by restrictions Mongoose is enforcing on the field value
```
bassjacob: caused?

jstockwin:
@molomby This looks awesome and is definitely an issue I've experienced - would be good to have it fixed! I've actually not experienced it recently, but I think this is because I set the `keepAlive` mongo option which helps with the disconnection. 100% agree that keystone should properly error if it can't connect, though. I've had sites go down and haven't noticed until the client asks why they can't access their site, because it never actually errors so forever doesn't catch it. 

Just to clarify something:
> This provides logging of failures and causes the thread to fail properly if the DB connection drops and **cannot be reestablished automatically.**

There is nothing in your code that attempts to re-establish the connection automatically. Is this something mongoose does by itself, and so your code only gets reached after it's tried to reconnect?

JedWatson:
@jstockwin this is also the motivation behind the health checks feature we just added - the default implementation includes a query on the `User` model (if specified) which ensures you can successfully retrieve data from the database. Between that new feature and this change, it should make proper monitoring much easier to set up and reduce the chance for apps to go into an unnoticed error state.

Not sure about your question though, I'll leave that for @molomby to answer...

Anyway this is good to merge 👍  nice work John!

molomby:
@jstockwin
> Is this something mongoose does by itself, and so your code only gets reached after it's tried to reconnect?

Yeah, the Mongo driver does this on it's own. If the DB drops for < 30 seconds (by default) it'll recover without raising an exception. See the `reconnectTries` and `reconnectInterval` settings here.. 
https://mongodb.github.io/node-mongodb-native/2.1/api/Server.html

PR #3907, opened 2017-01-17
Feature/healthchecks

bassjacob:
## Description of changes
Adds an optional healthcheck endpoint at `/keystone/server-health` using the new [keystone-healthchecks](https://github.com/keystonejs/keystone-healthchecks) library.

It adds a healthcheck endpoint, and will add healthchecks in the following way:

if `keystone.get('healthchecks') === true` it will use a default keystone healthcheck that asserts that the mongodb database is able to be queried.

otherwise, it will try and use `keystone.get('healthchecks')` as an object, and pass it straight to the healthchecks module.

## Testing

- [x] npm run test

### DIFF
--- a/admin/server/app/createDynamicRouter.js
+++ b/admin/server/app/createDynamicRouter.js
@@ -1,9 +1,9 @@
 var bodyParser = require('body-parser');
 var express = require('express');
 var multer = require('multer');
+var safeRequire = require('../../../lib/safeRequire');
 
 module.exports = function createDynamicRouter (keystone) {
-
 	// ensure keystone nav has been initialised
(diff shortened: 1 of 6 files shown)


jstockwin:
@bassjacob This sounds cool. Can you show a screenshot of what it looks like?

bassjacob:
@jstockwin output is just JSON, so you'll get something similar to this:


```json
status: 200

{
    "healthchecks": [
        {
            "data": "Can query the database for list Users",
            "healthy": true,
            "name": "Can Query List: Users"
        }
    ],
    "healthy": true
}
```

JedWatson:
`keystone-healthchecks` is published so we're good to merge 🎉 

Nice work @bassjacob :)

PR #3909, opened 2017-01-17
Display file originalname inside FileField when it is present

gfrancischelli:
## Description of changes
Display file `originalname` inside `FileField` when it is present

screenshot


bassjacob:
Should we expose the field in the UI as part of this? @JedWatson

JedWatson:
Thanks for the PR @gfrancischelli!

Can you provide some insight into your reasoning for this change? Original filenames can often be duplicate or contain misinformation (e.g. screenshot titles on Mac, `untitled1.png`, etc) whereas the stored filename is pretty important data to display. If the original name is important to the user, wouldn't it be more helpful to use the `originalname` filename naming method for uploaded files?

Not saying we won't support this, just want to hear more about your use case / thought process.

If we do add this feature I think it's safer to implement it as an option on the File field, which lets the developer decide which value to display.

gfrancischelli:
Thanks for the thougthful response @JedWatson (: 
Lots of things i haven't even considered. 

I had some pdfs for download but needed flexibility in displaying the file name. I thought maybe it could be useful for people out there.

Should we call it a file Title (?) ~~naming things it's too hard~~ :dizzy_face: 

### Changes
- Add the boolean option `displayname` in the File field
```javascript
Post.add({
  file: {
    type: Types.file,
    storage: storage,
    displayname: true,
  }
});

```
- Add a virtual method `displayfilename`  that replaces spaces with dashes and append file extension (taken from filename).   i.e `Alt File Name`  ->  `Alt-File-Name.jpg`

- Add `displayname` in Admin UI (not pretty though)

admin ui

### Use Cases
```pug
// Pretty, optional alternative filenames
a(
  href=#{file.url},
  download=#{file.displayfilename}
)= file.displayname || file.filename

// Better alt attr
img(src=#{image.url}, alt=#{image.displayname})
````

jstockwin:
@gfrancischelli The [build is failing](https://travis-ci.org/keystonejs/keystone/jobs/195640989#L580) due to linting errors. You can check these locally by running `eslint .` in your keystone directory. 

I'll leave the discussion about the direction of this PR to you and Jed, but it's probably worth fixing the linting - it won't be merged until it's fixed. Thanks for spending the time to do this!

PR #3932, opened 2017-01-24
Minor improvement to the handling of text indexes

molomby:
Addresses the (hopefully?) unusual case where `searchUsesTextIndex` is enabled for a list but the collection has an existing text index that *isn't* described by the list. In this scenario, ensureTextIndex() will log the issue to stderr and the text search will use the existing index. This might produce unexpected search behaviour.. or might be exactly what the app builder intended; we don't know.

## Description of changes

Removes the requirement for the Mongoose list to define (either manually or automatically) a text index for a text index to be utilised. Allows existing text indexes on the collection to be used.

## Testing

> OK. 1790  total assertions passed. (17m 3s)

### DIFF
--- a/lib/list/ensureTextIndex.js
+++ b/lib/list/ensureTextIndex.js
@@ -89,7 +89,9 @@ function ensureTextIndex (callback) {
 			}
 
 			// It's a text index but not one of ours; nothing we can do
-			debug('Existing text index \'' + existingIndexName + '\' in \'' + list.key + '\' not recognised; no action taken');
+			console.error('list.ensureTextIndex() failed to update the existing text index \'' + existingIndexName + '\' for the \'' + list.key + '\' list.');
+			console.error('The existing index wasn\'t automatically created by ensureTextIndex() so will not be replaced.');
+			console.error('This may lead to unexpected behaviour when performing text searches on the this list.');
 			return;
 		}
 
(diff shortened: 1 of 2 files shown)


molomby:
Thinkmill people - This is being brought across from the Keystone branch in use by `blueshyft-core`

lib/list/ensureTextIndex.js
```diff
@@ -89,7 +89,9 @@ function ensureTextIndex (callback) {
 			}
 
 			// It's a text index but not one of ours; nothing we can do
-			debug('Existing text index \'' + existingIndexName + '\' in \'' + list.key + '\' not recognised; no action taken');
+			console.error('list.ensureTextIndex() failed to update the existing text index \'' + existingIndexName + '\' for the \'' + list.key + '\' list.');
```
bassjacob: Will this print on every boot? should we keep it in a debug stream?

molomby: It should print very rarely but when it does happen it's pretty important to flag. Besides, it's not actually exiting or anything drastic like that..

JedWatson: These comments would be better IMO as one line with `\n` line break characters, keeps them grouped in log aggregation environments as a single piece of information, not three. I'll merge & clean that up.

Issue #3945, opened 2017-01-26
Tests broken out of the box on Windows 10


OskarKlintrot:
### Steps to reproduce the behavior
On Windows 10 PS:
```
> git clone url/to/keystone
> cd keystone
```
Add `"linebreak-style": 0` to `.eslintrc`
```
> npm install
> npm run test-all
```

### Expected behavior
Tests passing

### Actual behavior
Tests not passing.

___________________
[output from PowerShell](https://github.com/keystonejs/keystone/files/732050/console-log.txt)
[npm-debug.log](https://github.com/keystonejs/keystone/files/732047/npm-debug.log.txt)

bassjacob:
I believe the `npm run test-all` command runs `npm run test-e2e-bg` which is intended for linux. @webteckie what's the correct way to run this on windows? should we be using [cross-env](https://www.npmjs.com/package/cross-env) rather than export here?

bassjacob:
@webteckie the error output complain about the export shell command in the test-e2e-bg command from the package.json

jstockwin:
@bassjacob If you look at the PowerShell output, the error is to do with the Mocha tests not the e2e tests :)

bassjacob:
@jstockwin, I don't think so. All the unit tests pass (the printout during the unit test run is console.error output from a model validator).

Line 53 of the output from powershell shows the actual failure:

```
> keystone@4.0.0-beta.5 test-e2e-bg C:\_repos\keystone
> export KNE_SELENIUM_START_PROCESS=false && node test/e2e/server.js --env default

'export' is not recognized as an internal or external command,
operable program or batch file.

npm ERR! Windows_NT 10.0.14393
npm ERR! argv "C:\\Program Files\\nodejs\\node.exe" "C:\\Program Files\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "run" "test-e2e-bg"
npm ERR! node v6.2.2
npm ERR! npm  v3.9.5
npm ERR! code ELIFECYCLE
npm ERR! keystone@4.0.0-beta.5 test-e2e-bg: `export KNE_SELENIUM_START_PROCESS=false && node test/e2e/server.js --env default`
npm ERR! Exit status 1
```

jstockwin:
Ah yeah okay my bad. `export` works on my (Windows 10) machine... strange. Maybe it's PowerShell. 

Running `npm test-e2e` should work, as it won't export anything. As a fix, cross-env looks good to me.

[...]

PR #3946, opened 2017-01-26
Added option to use custom signin and signout functions

OskarKlintrot:
## Description of changes
Added option to use custom signin and signout functions by setting `'signin'` and `'signout'` to functions in `keystone.init({})`

## Testing
Tests dosen't seem to be working on windows 10 with a clean install (see #3945). At least all tests that runs is passing.


OskarKlintrot:
What do you guys think of this? Since this behavior already exists for `admin` IMHO it would make sense to have the same behavior for `signin` and `signout` as well.

PR #3947, opened 2017-01-26
Use cross-env to export on multiple platforms

jstockwin:
## Description of changes

cc @bassjacob 

## Related issues (if any)
#3945

## Testing


bassjacob:
we should also add it here: https://github.com/keystonejs/keystone/pull/3947/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R158

jstockwin:
My bad. I'm on my way to a lecture now, but feel free to change it :)

bassjacob:
np :smile: I'll add a commit to this pr

package.json
```diff
@@ -162,7 +163,7 @@
     "test-unit": "node test/pretest.js && mocha",
     "test-admin": "mocha --opts test/mocha-admin.opts",
     "test-e2e": "node test/e2e/server.js --env default",
-    "test-e2e-bg": "export KNE_SELENIUM_START_PROCESS=false && node test/e2e/server.js --env default",
+    "test-e2e-bg": "cross-env KNE_SELENIUM_START_PROCESS=false && node test/e2e/server.js --env default",
```
webteckie: fyi...we don't really need to use an env var export here as there's a `--selenium-in-background` switch that can be specified.

jstockwin: Hmmm @webteckie I might be missing something, but does setting the environment variable like this actually do anything..? It looks to me like KNE overwrites it? 

https://github.com/keystonejs/keystone-nightwatch-e2e/blob/master/index.js#L101

webteckie:
Correct...should use the switch, instead.

PR #3949, opened 2017-01-27
App state refactor: Followup to https://github.com/keystonejs/keystone/pull/3836 

gwyneplaine:
## Description of changes
Refactor of existing List scene app state and data flows. The aim of this was to eliminate recurring bugs to do with query parameters in varying use cases. Overall, the highlight points are as follows:
* Manual manipulation of List queries (i.e. adding filters via the ui) are now decoupled from automatic processes happening from URL changes, such as deep-linking from a given URL, or hitting back button in the browser.
* Before this change, query computation was happening inconsistently, sometimes after loading items, sometimes after setting the active list. This has been changed so that the following actions will always occur in the following order: 
  * SELECT_LIST
  * SET_ACTIVE_LIST 
  * QUERY_HAS_CHANGED
  * @@router/LOCATION_CHANGE
  * LOAD_ITEMS

A more detailed log of changes are as follows: 
### List/index.js
* the `parseQueryParams` method within the component has been removed completely.
* dispatch to `initialListLoad` has been removed. This is no longer necessary, as we do not need to distinguish an initial load from a successive one. 

### List/actions/active.js
* `setActiveFilters` action creator removed
* `setActiveSort` action creator now returns a 'SELECT_ACTIVE_FILTER' action. 
* `setActiveColumns` action creator now returns a 'SELECT_ACTIVE_COLUMNS' action. 
* `setFilter` action creator now returns a 'SELECT_FILTER' action.

### List/reducers/active.js
* `cachedQuery` object added to initial state. 
* case `QUERY_HAS_CHANGED` added to reducer. 
  * replaces `search`, `sort`, `currentPage`, `filters`, and `columns` with the `parsedQuery` object in the action. 
* case `REPLACE_CACHED_QUERY` added to reducer.
  * replaces the existing cachedQuery object with the cachedQuery object in the action. 

### sagas/index.js
* setActiveFilter saga added.
  * Moved from previous thunk based paradigm to a saga based one
* setActiveColumns saga added.
  * Moved from previous thunk based paradigm to a saga based one
* setActiveSort saga added.
  * Moved from previous thunk based paradigm to a saga based one
* `rootSaga` has been overhauled. 
  * Removed code that was implementing conditional saga calls based on INITIAL_LIST_LOAD.
  * We now fork three new sagas `setActiveFilter`, `setActiveColumns`, `setActiveSort`
  * takeLatest for updateparams now takes 'QUERY_HAS_CHANGED` as a triggering action, as opposed to `SET_ACTIVE_LIST`.

### sagas/updateParams.js
* renamed to `queryParamsSagas.js`
* `updateParams` now puts  a `REPLACE_CACHED_QUERY` action to the store, containing the new constructed query as the cachedQuery. 
* Added `evalQueryParams` saga.
  * This saga is tasked with evaluating whether or not the query parameters have changed by comparing the passed in query with the cachedQuery in state. 
  * If the query has changed, it dispatches `QUERY_HAS_CHANGED` to the store
  * This also activates a saga flow once the state has updated, and in turn then triggers the `updateParams` saga
  * If the query has not changed it dispatches `QUERY_HAS_NOT_CHANGED` to the store, and calls the loadItems action creator. 
* Added parseQueryParams method. 
  * parseQueryParams is a normal function that takes the inputted query and passes each key and the currentList to the associated parser. 
  * the returned object is a new query object with all of the passed in query params parsed, interpreted and transformed correctly. 

### parsers/
Added a parsers folder, this folder contains logic previously within the thunk Action creators (setActiveColumns for example). The goal of this abstraction, is to decouple the act of dispatching actions from the interpretation and extraction of associated data from the List constructor. 

This folder contains the following functions: 
* sortParser:
  * Takes in a path, and a currentList object
  * returns an expanded sort object.
* columnsParser
  * Takes in a columns array and a currentList object. 
** returns an array of expanded columns objects. 
* filtersParser
  * Takes in an array of filters or a string query.
  * returns an array of expanded filter objects. 
* filterParser
  * Takes in a filter object
  * returns an expanded filter object.
* createFilterObject (moved from within the List/actions/active file. 
  * This is no longer called within the List/actions/active file, and is called in here instead. 
  * It's unique use in one location has not changed, and therefore has not warranted it being moved to a more generic utils file. 

## Testing


JedWatson:
:+1: Looks good @gwyneplaine, thanks!

PR #3950, opened 2017-01-27
Passing `req` to signin hooks to make it possible to add custom validation

OskarKlintrot:
## Description of changes
Passing `req` to signin hooks. This makes it possible to ie validate the users against a third part, for example Microsoft Active Directory or AWS Cognito, by taking the credentials from `req.body`.

## Related issues (if any)
Closes #3944 

## Testing

- [x] Please confirm `npm run test-all` ran successfully.

PR #3965, opened 2017-02-03
Adding req to singin hook

OskarKlintrot:
## Description of changes
I missed this place in PR #3950, sorry. 

## Testing

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -126,7 +126,7 @@ var doSignin = function (lookup, req, res, onSuccess, onFail) {
 };
 
 exports.signin = function (lookup, req, res, onSuccess, onFail) {
-	keystone.callHook({}, 'pre:signin', function (err) {
+	keystone.callHook({}, 'pre:signin', req, function (err) {
 		if (err) {
 			return onFail(err);
 		}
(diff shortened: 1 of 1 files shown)


OskarKlintrot:
I also noticed an unrelated issue here (I think). This hook is located *before* the signin process whilst the other hook is located *inside* the signin process. This creates an inconsistent behavior. Is this by design or should I open an issue?

creynders:
Ow... there's 2 `session` files now? This seems pretty bad, it's not DRY and makes things error prone. I think it'll need it a refactor.

OskarKlintrot:
If I understand it correct ([and the way I am using this `session.js`](https://github.com/Certaincy/intranet/blob/master/keystone/src/auth/api/signin.ts#L24)) is that one of them is for the keystone admin and the other one is to be used when you are opening up the api. Not very DRY, I agree.

JedWatson:
Agreed with the DRY issue... It happened this way to avoid a breaking change in the API while updating the way the Admin UI signing works. Not ideal - we'll address that after 4.0 is released.

Thanks @OskarKlintrot!

PR #3973, opened 2017-02-07
Geopoint validation should succeed if an array with undefined values

vitalbone:
## Description of changes
The geopoint field type input validation currently  returns `true` if a value is either `undefined`, `null`, or an empty string. It should also accept an array with undefined values.

## Related issues (if any)
N/A

## Testing


bassjacob:
hey @vitalbone, good catch.

Can you add a test case here: https://github.com/keystonejs/keystone/blob/master/fields/types/geopoint/test/type.js#L51 so we don't regress in the future?

bassjacob:
Is there any point to ensuring that the array of undefined values is 2 in length?

PR #3959, opened 2017-01-30
Exception with the JSON download & Fix

export-mike:
Hi!

Firstly can I say going through this codebase is a joy, I've known of keystone for awhile and tinkered with it in my spare time - its great to see how its evolved. 

I found that this was throwing an exception when you click to download data as json.

you can reproduce the bug at http://demo.keystonejs.com/keystone/posts download > json > download

the server hangs with no response... 

this fixes the issue, the flow is now similar with the CSV download. Which is where I pinched the 'list-to-array' require from. Its now fixed for this scenario.

A few other points I have noted for discussion:
- Do we also want it to be a json file download or json response?
- we should probably also have an exception handler to return 500 from express for anything else?
- I notice the user can enter "password" into the query string and reveal the hash, is this expected behaviour? I assume its safe under a users logged in session
- I've noticed the require for 'baby-parse' is within the handler? do we want to hang the server on request? https://github.com/keystonejs/keystone/blob/master/admin/server/api/list/download.js#L9

If you're after any more help on issues or more info on this bug, I'd be more than happy to help! 

Thanks 

Mike

## Description of changes

I've added the missing check for "is a string" if so then convert to an array which done by the third-party dependency 'list-to-array'.

## Testing

a note: I got this selenium error before I made changes on the branch, this has actually prevented me from getting the selenium tests to pass. 

```Timed out while waiting for element <.primary-navbar [data-section-label="octicon-sign-out"]> to be present for 60000 milliseconds.  - expected "visible" but got: "not found"
    at Object.after (/Users/MikeJames/projects/keystone-play/keystone/test/e2e/adminUI/tests/group006Fields/commonFieldTestUtils.js:13:22)
```

## Testing

- [ ] Please confirm `npm run test-all` ran successfully.

PR #3984, opened 2017-02-09
Improved Exception Handling on Download.js

export-mike:
### Updated
https://github.com/keystonejs/keystone/pull/3983 (closed) @bassjacob 

I noticed this wasn't behaving as expected previously when using callbacks, the server would hang. 

This should help any unknown exceptions that may occur to be caught and handled, allowing a response to the client.

## Description of changes
In a previous PR https://github.com/keystonejs/keystone/pull/3959 I noticed the download request for JSON was failing which has now been fixed however the server actually hangs if theres an exception, as there was no try/catch in the callback code, I've now converted this to the mongoose promise way of doing things.

## Related issues (if any)
https://github.com/keystonejs/keystone/pull/3959 (merged)

### DIFF
--- a/admin/server/api/list/download.js
+++ b/admin/server/api/list/download.js
@@ -5,7 +5,7 @@ TODO: Needs Review and Spec
 var moment = require('moment');
 var assign = require('object-assign');
 
-module.exports = function (req, res) {
+module.exports = function (req, res, next) {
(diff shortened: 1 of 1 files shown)


bassjacob:
I'll wait for CI to catch up but LGTM

PR #4003, opened 2017-02-15
Moved the check for keystone readiness for e2e to kne.

webteckie:
## Description of changes
This change moved the check for keystone readiness for e2e testing to the kne framework.

## Related issues (if any)
N/A

## Testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,6 @@
     "react-engine": "4.2.1",
     "rimraf": "2.5.4",
     "sinon": "1.17.7",
-    "superagent": "3.4.1",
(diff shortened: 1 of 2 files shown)


bassjacob:
Did this move previously into KNE? The KNE version wasn't bumped as a part of this, so I'm not sure where this code is now being run?

webteckie:
@bassjacob greenkeeper will do the bump here once the kne package is published.

bassjacob:
Are we safe to remove this code before that bump?

webteckie:
Well, the bump should happen first then this.

Issue #4030, opened 2017-02-23
RelationshipType.updateItem() always uses new data when the relationship is `many`


Noviny:
### Steps to reproduce the behavior
Have a relationship with many, that has a value. For our example, cats.
Call `List.updateItem(itemToUpdate, { name: 'Bob' }, callback)`

### Expected behavior
As we have not specified cats, Bob's 3 cats are still linked

### Actual behavior
Bob has no cats anymore. The field is reset to `[]`.

Picking this issue up myself. Plan is to change `relationship.prototype.updateItem` to treat no `data[this.path]` as a skip to updating for both single and many relationships

Noviny:
Just looked in to this a little more, and this will remove for either single or multiple relationships. This seems intentional, however interacts poorly with the above pattern.

PR #4035, opened 2017-02-24
Update relationship field to use standard error handling

Noviny:
Previously, Relationship fields were cleared by calls to updateItem if
not explicitly passed in. Now this and other undefined values will
leave them untouched.

This is consistent with how other fields work, however is a breaking
change to updateItem() for relationships.

See #4030

### DIFF
--- a/fields/types/relationship/RelationshipType.js
+++ b/fields/types/relationship/RelationshipType.js
@@ -219,10 +219,15 @@ relationship.prototype.updateItem = function (item, data, callback) {
 	if (item.populated(this.path)) {
 		throw new Error('fieldTypes.relationship.updateItem() Error - You cannot update populated relationships.');
 	}
+
+	var value = this.getValueFromData(data);
+	if (value === undefined) {
+		return process.nextTick(callback);
+	}
 	if (this.many) {
 		var arr = item.get(this.path);
 		var _old = arr.map(function (i) { return String(i); });
-		var _new = data[this.path];
+		var _new = value;
 		if (!utils.isArray(_new)) {
 			_new = String(_new || '').split(',');
 		}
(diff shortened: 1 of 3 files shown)


fields/types/relationship/RelationshipType.js
```diff
@@ -232,9 +237,9 @@ relationship.prototype.updateItem = function (item, data, callback) {
 			item.set(this.path, _new);
 		}
 	} else {
-		if (data[this.path]) {
-			if (data[this.path] !== item.get(this.path)) {
-				item.set(this.path, data[this.path]);
+		if (value) {
```
bassjacob: perhaps `if (value && value !== item.get(this.path))`?

fields/types/relationship/RelationshipType.js
```diff
@@ -219,10 +219,15 @@ relationship.prototype.updateItem = function (item, data, callback) {
 	if (item.populated(this.path)) {
 		throw new Error('fieldTypes.relationship.updateItem() Error - You cannot update populated relationships.');
 	}
+
+	var value = this.getValueFromData(data);
+	if (value === undefined) {
+		return process.nextTick(callback);
```
bassjacob: why nextTick rather than setImmediate?

Noviny: nextTick is the pattern for this check in keystone.

fields/types/relationship/RelationshipType.js
Noviny: We don't want the 'else if' to run if there is a value though the if statements should likely be tidied up to make that clear

JedWatson:
Good change @Noviny.

I'm going to add this to the upgrade guide because while it's a bug fix, it could be considered breaking.

PR #4059, opened 2017-03-03
Update money field for update in numeraljs

Noviny:
In numeraljs 2.0.0, language was renamed to locale, and language
function and folder has been deprecated.

This fixes this, and adds test of the format method.


Noviny:
Testing this exposed that numeraljs as a singleton interacts poorly with two different money fields, only one of which has a currency. Once the currency is set, it remains set.

This is probably an indication we should move away from numeraljs.

Issue #4060, opened 2017-03-03
Numeral is not unformatted after a format is called


Noviny:
### Steps to reproduce the behavior
Call `.format()` on a `money` field with a locale set, such as `en-gb`, then call `numeral.format()` without defining a locale anywhere else within keystone.

### Expected behavior
The new call will be agnostic of the first call occurring, and have no locale.

### Actual behavior
The locale is not unset, and all future calls to `numeral.format()` will use the defined locale, until the locale is reset.

### Possible solutions
Move away from numeral, probably to using `toLocaleString()`.
Refactor code so our locale is reset to the default numeral locale ('en') after we have formatted anything. If we take this path, we need to also set a default for all money fields.

Additionally, this holds an impact on how the `Number` field operates if there is a `Money` field, and a `Money` field that sets `currency` is going to cause numbers to be formatted using the locale unless it is reset afterwards.

PR #4061, opened 2017-03-03
Revert "Update money field for update in numeraljs"

bassjacob:
Reverts keystonejs/keystone#4059


Noviny:
#4060 revealed that fixing this issue has possible negative consequences for the `Number` field. This change is being reverted so we can solve for this.

Issue #4071, opened 2017-03-07
Edit buttons return 404


CameronSima:
After changing my blog's url from /blog to /main/blog, the edit buttons which appear beside posts when signed in link to /blog/keystone/posts/58bdec38a03fa26f2c3c54ef, which returns a 404 error, instead of the expected /keystone/posts/58bdec38a03fa26f2c3c54ef, which is the correct url for the page to edit the post.

Noviny:
This is a bug with [data-ks-editable](https://github.com/keystonejs/keystone/blob/c4789388ef09d09f39585f68b2a6fc898b7da040/admin/public/js/content/editor.js), line 42. It's using a relative path, which will take in the current location.

Made a PR for this on #4072. Not sure what a workaround would be while you wait for that to be merged beyond applying this fix to your own keystone version.

CameronSima:
@Noviny Thank you, works like a charm

PR #4072, opened 2017-03-07
Fix data ks editable with field paths

Noviny:
See explanation on #4071


bassjacob:
just confirming that this changes a relative url path to an absolute path from the root. is data.path drawn against whatever the keystone url is?

Noviny:
data.path is always `keystone/modelName` irrespective of location.

bassjacob:
LGTM

Noviny:
@bassjacob Just went to look in to this locally, and all tests passed. Can we retrigger this manually?

PR #4078, opened 2017-03-09
Use same query object to perform count and query

### LINKED ISSUE #4056: Keystone Beta 4.0.0-beta.5 - Issue with Pagination
Found a issue when updating to 4.0.0-beta.5. While using pagination, I get back only the results in the "where" clause, but the total count returned, is not specific to the conditions of the where clause, but shows the total number of documents in the collection.

Example Query:

```
var q = Contact.paginate({
  page: req.query.page || 1,
  perPage: req.query.limit || 10,
  maxPages: 10
}).where('site', site);
```

Keep in mind, that "site" is a relationship to contact. A contact always has a "site".

In Keystone 0.3.21, the query returns the correct "total" count. So say I have 20 contacts in site "a", I'll get back 10 contacts, with a total of 20 to paginate thru.

But in Keystone 4.0.0-beta.5, with the same query, I get back 10 contacts, but the total displays an incorrect number, say 60 contacts, the total number of documents in the collection.

Noviny:
This fixes a bug where people were modifying the query after creation,
but this was not being applied to countQuery so the count would be
wrong.

Fixes #4056

### DIFF
--- a/lib/list/paginate.js
+++ b/lib/list/paginate.js
@@ -20,7 +20,6 @@ function paginate (options, callback) {
 	options = options || {};
 
 	var query = model.find(options.filters, options.optionalExpression);
-	var countQuery = model.find(options.filters);
 
 	query._original_exec = query.exec;
 	query._original_sort = query.sort;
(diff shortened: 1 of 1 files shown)


Noviny:
Talked about this with @JedWatson, who highlighted that this solution does not work with promise calls on the query object, and is not an adequare solution. Closing this as a PR. See #4126 for more details.

Issue #4082, opened 2017-03-09
Datetime field type adds 5 hours to data


christroutner:
### Expected behavior
When converting a JavaScript Datetime object to an ISOString and then uploading to a KeystoneJS app, I expect the get the same time string back from the server. 

### Actual behavior
The Datetime string returned by the server has 5 hours appended to the value.

### Steps to reproduce the behavior
1. I create a model with a Datetime field like so: `reportGenerated: { type: Types.Datetime, default: Date.now }`

2. Below is an example of an object with JavaScript Date objects:

![Image 1](http://connextcms.com/images/documentation/keystone-date-issue/img1-dateobj-before-isostring.jpg)

3. I use an API, similar to [the API Gist by Jed](https://gist.github.com/JedWatson/9741171#file-routes-index-js-L24) or the [File and Image API Guide](https://github.com/keystonejs/keystone/blob/master/docs/guides/API-File-Image-Uploads.md). That API will reject JavaScript Date objects, but will accept an ISO string. Here is an example of the same object with it's Date objects converted to ISO strings in preparation of uploading to the KeystoneJS server:

![Image 2](http://connextcms.com/images/documentation/keystone-date-issue/img2-isostring-before-sending-to-server.jpg)

4. Send data to the server succeeds. However, when retrieving the same data from the server, the ISO strings have 5 hours added to them:

![Image 3](http://connextcms.com/images/documentation/keystone-date-issue/img3-isostring-after-getting-from-server.jpg)

tomhardman0:
It looks like it's normalising your original GMT-0800 to UTC, so the ISO string you end up with looks like it's 8 hours ahead when really it's just UTC.

intmainvoid:
@christroutner - Was/Is your client in a separate timezone to your server when you tried this?

What happens if you turn the utc option on on your reportGenerated field? (ie. change

```
reportGenerated: { type: Types.Datetime, default: Date.now }
```
to
```
reportGenerated: { type: Types.Datetime, default: Date.now, utc: true }
```

I've been working on a fix to a couple of related problems (See PR: https://github.com/keystonejs/keystone/issues/4112). Maybe it will fix your issue also?

thejameskyle:
Does #4112 fix this issue?

creynders:
@thejameskyle I don't think so, since it only fixes the Date field, not DateTime fields. TBH I think these kinds of bugs keep on popping up until the fields are re-written properly from scratch.

creynders:
[...]

Issue #1702, opened 2015-09-30
Saving causes published date to rollback 1 day


jimulle:
Every time I save in the Admin UI, the published date for the item I just saved rolls back one day (regardless of editing the published date field)

Saving with date '2015-04-22' inputs '2015-04-21' in the database (and subsequently displays that new date when the page reloads).

foobar8675:
seems to work fine for me. i'm on version version 0.3.14.

mxstbr:
I'll close this since it's not reproducable, if this issue still exists for you feel free to comment again!

kraigh:
I have this exact issue on 0.3.19. Any ideas what could be causing it? Tested on Firefox and Chrome on OSX.

wmertens:
Could it be that the browser and the server are in different time zones?

kraigh:
Yes, the server is UTC and my browser is CST. How can I fix this?

wmertens:
Then this is a bug in Keystone. Dates should be handled in UTC.

wmertens:
It looks like it should be sufficient to use the `moment.utc(…)` functions in https://github.com/keystonejs/keystone/blob/master/fields/components/DateInput.js

wmertens:
Basically, dates without time should be formatted and parsed in UTC, so they don't jump depending on time zone. Not sure how far that impacts things, and it will probably break existing deployed code that assumes that the date value it gets from the server is in the browser time zone.

Let's fix this and make sure people realize it's a breaking change.

Alternatives:
* store date-without-time values as the string `YYYY-MM-DD`. This doesn't break lexicographic sort, easily converts to a Date value, takes less room in JSON, and makes it explicit that there is no time zone attached to it. May break even more code assuming to get Date objects, may cause lots of back-and-forth conversions
* store dates as noon on the +12 timezone, so that no matter what time zone it gets converted to, it still is the same date. This may cause the least breakage?
 * (My math may be off here, I mean the moment in time that is in the same day for all time zones. It may not exist, be off by a second, so that may also be a problem)

Anybody experienced at reliably storing dates?

arthurcamara1:
Same happening here.

wcalebgray:
+1 same issue happening here

jstockwin:
@wmertens This is also an issue for Datetime, and your second alternative won't fix that, so not that one. I think I agree with your initial suggestion of storing in UTC. A breaking change to fix a bug is okay...

[...]

PR #4112, opened 2017-03-16
Implemented timezone fix for fields that use the 'Date' type

intmainvoid:
## Description of changes

This change fixes two bugs on keystone 4.0.0-beta.5:

- If using a field with type “Date” (ie. Types.Date) in the model without the UTC option turned on AND the keystone users are in different timezone(s) to where the server is running, dates may be recorded inaccurately. Specifically, when the user clicks on the save button for a form that contains a date, the date saved (and subsequently displayed) may be the date for one day before the date intended to be saved.

- If using a field with type “Date” (ie. Types.Date) in the model with the UTC option turned on, the incorrect date may be displayed upon selecting a new date from the date picker. Specifically, whenever the user selects a new date, the date for the day immediately before that date would be displayed. Clicking on save at that point would cause that incorrect date to be saved. This happens irrespective of the configured timezone(s).

## Related issues (if any)

https://github.com/keystonejs/keystone/issues/1702
https://github.com/keystonejs/keystone/issues/4082

## Notes

For the fix to take effect, the UTC option for the DateType field must be set.

ie. change:
```
publishedDate: { type: Types.Date },
```
to
```
publishedDate: { type: Types.Date, utc: true },
```

### DIFF
--- a/fields/types/date/DateField.js
+++ b/fields/types/date/DateField.js
@@ -43,20 +43,22 @@ module.exports = Field.create({
 			value: value,
 		});
 	},
-	moment (value) {
-		var m = moment(value);
-		if (this.props.isUTC) m.utc();
-		return m;
+	toMoment (value) {
+		if (this.props.isUTC) {
+			return moment.utc(value);
+		} else {
+			return moment(value);
+		}
 	},
 	isValid (value) {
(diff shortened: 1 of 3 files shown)


JedWatson:
Nice! Thanks @intmainvoid

PR #4130, opened 2017-03-21
Add warning for paginate the first time it is run

Noviny:


JedWatson:
You're on the right track - the functionality in `lib/list/paginate.js` should be centralized though.

So the call is something like `showDeprecationWarning('list.paginate', 'List.paginate is no longer supported. It will will be deprecated in keystone 5')`

PR #4140, opened 2017-03-24
Use safeRequire instead of manual try-catch requires

### LINKED ISSUE #3908: Replace inline try/catch with safeRequire
In current state, the following is copied throughout the codebase:

```javascript
try {
  return require('some unbundled lib');
catch (e) {
  if (e.code === 'MODULE_NOT_FOUND') {
    // do something
  }
  throw e;
}
```

we can replace this in these places with the new safeRequire function from #3907 for consistency.

thejameskyle:
Resolves #3908


lib/safeRequire.js
```diff
@@ -3,7 +3,9 @@ module.exports = function safeRequire (library, feature) {
 		return require(library);
 	} catch (error) {
 		if (error.code === 'MODULE_NOT_FOUND') {
-			throw new Error('\nTo use ' + feature + ' install ' + library);
+			console.error('\nTo use ' + feature + ' install ' + library);
+			process.exit(1);
```
thejameskyle: I switched to using `process.exit(1)` because most of these cases were already doing that and it seems to make more sense.

lib/safeRequire.js
```diff
@@ -3,7 +3,9 @@ module.exports = function safeRequire (library, feature) {
...
+			return;
```
bassjacob: why return after a `process.exit(1)` ?

thejameskyle: Became a habit because of Flow

bassjacob: ah, the always return even null. a null return after an IO call seems even weirder from a type system perspective :stuck_out_tongue:

export-mike:
LGTM 👍  @Noviny @blargity @JedWatson

PR #4157, opened 2017-03-30
Expose password validation 

export-mike:
extract password validation so users can access the validation within their own code for things like password reset

## Description of changes
I've extracted validation as a static method added to the PasswordType, allowing consistent password validation in their own application logic.

In userland you can now access the same logic,
```JavaScript
const User = keystone.list('User');
const min = User.fields.password.options.min;
const max = User.fields.password.options.max || 72;
const complexity = User.fields.password.options.complexity;
const { result, detail } = Types.Password.validate(password, confirmPassword, { min, max, complexity });
```

Whilst this isn't perfect with the amount of boilerplate required. It does allow a user to have access to the same logic.

## Testing

- [Y] Please confirm `npm run test-all` ran successfully.


export-mike:
I'm closing this.

vitalbone:
@Noviny @JedWatson 

This is now passing all it's tests. The broken Travis build is related to a dependency not liking the use of `const` in strict mode 😕  

screen shot 2017-09-21 at 12 19 54 pm

Good to go? or more work required? Not sure due to the lack of context on where this issue was left...

Noviny:
@vitalbone This dependency is in the tests themselves?

If the tests are failing because they themselves are blocking strict mode then that's a bit ¯\_(ツ)_/¯

If that's the only problem and has your LGTM, feel free to hit the merge button.

Issue #4055, opened 2017-03-02
Link preview is a span instead of a link


ellipticaldoor:
I'm using keystone 4.0.0-beta.5

When I add a file everything works just fine, but in the file field inside the admin I get a link that doesn't work because is a span. When I change the tag from **span** to **a** it works.

``` javascript

const publicPath = 'offers/'
const storage = new Storage({
	adapter: Storage.Adapters.FS,
	fs: {
		path: join(__dirname, '../../../static/' + publicPath),
		publicPath,
	},
	schema: { url: true },
})

// And in my schema
image: { type: Types.File, storage, initial: true, required: true },

```

Also, how I can preview the image in the admin?

![screenshot](https://cloud.githubusercontent.com/assets/3449006/23508147/4313c306-ff48-11e6-8792-66e179a8843a.jpg)

giusepperj:
+1

r1cebank:
+1

w0rldart:
+1

gggllm:
+1

w0rldart:
I think the project is abandoned :cry:

ellipticaldoor:
😢

Noviny:
Rejoice! #4174 has been merged.

(new beta release likely on Monday, which will have the fix in it)

proteye:
+1

PR #4174, opened 2017-04-03
Make file field a link when it has a href

sktt:
If you have specified `schema: { url: true }` in your storage adapter, admin ui will now make the span into a link. If not, then it will just be a span like before. 

Cheers

### DIFF
--- a/fields/types/file/FileField.js
+++ b/fields/types/file/FileField.js
@@ -126,7 +126,7 @@ module.exports = Field.create({
 		return (
 			<div>
 				{(this.hasFile() && !this.state.removeExisting) ? (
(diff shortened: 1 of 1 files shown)


sktt:
This fixes https://github.com/keystonejs/keystone/issues/4055

Noviny:
@sktt Nice simple fix, thanks for this!

sktt:
wooop nice to see u in action, @Noviny, I thought all members where long gone : D

Noviny:
Nup, we're still here =D

We're going to get a new beta out on Monday as well. Watch this space!

PR #4175, opened 2017-04-03
Cloudinary admin url update: fixes #2724

### LINKED ISSUE #2724: Cloudinary secure option not working?
I'm running a site for a charity, the page [here](https://www.students4students.org.uk/committee) displays mixed content warnings, even though I do `keystone.set('cloudinary secure', true);` [here](https://github.com/Students4Students/KeystoneJS-Website/blob/master/keystone.js#L58), and use the handlebars CloudinaryUrl helper [here](https://github.com/Students4Students/KeystoneJS-Website/blob/master/templates/views/committee.hbs#L13) as set out in the keystone [docs](http://keystonejs.com/docs/configuration/#services-cloudinary).

Apologies for linking directly to a site, but it seemed the quickest way of demonstrating the issue. I've restarted the app many times, so that's not the problem. Am I missing something here..?

Also, sometimes these pages load without any content for me, but I've been unable to get anyone else to reproduce it... If the page loads without any content for you, I'd be interested to know so I can look into it further. It's probably just some newbie mistake in the route...

Noviny:
Changes:
Cloudinary fields pass down `secure` property with their `getOptions`
call.

When a url is created by the admin UI, the secure property is added. This is handled at each location an image is referenced.

### DIFF
--- a/fields/types/cloudinaryimage/CloudinaryImageType.js
+++ b/fields/types/cloudinaryimage/CloudinaryImageType.js
@@ -277,6 +277,21 @@ cloudinaryimage.prototype.getData = function (item) {
 	return typeof value === 'object' ? value : {};
 };
 
+cloudinaryimage.prototype._originalGetOptions = cloudinaryimage.prototype.getOptions;
+
+cloudinaryimage.prototype.getOptions = function () {
+	this._originalGetOptions();
+	// We are performing the check here, so that if cloudinary secure is added
+	// to keystone after the model is registered, it will still be respected.
+	// Setting secure overrides default `cloudinary secure`
+	if ('secure' in this.options) {
+		this.__options.secure = this.options.secure;
+	} else if (keystone.get('cloudinary secure')) {
+		this.__options.secure = keystone.get('cloudinary secure');
+	}
+	return this.__options;
+};
+
 /**
  * Detects whether the field has been modified
  */
--- a/fields/components/columns/CloudinaryImageSummary.js
+++ b/fields/components/columns/CloudinaryImageSummary.js
@@ -58,7 +58,8 @@ var CloudinaryImageSummary = React.createClass({
 	},
 	renderImageThumbnail () {
 		if (!this.props.image) return;
-		const url = this.props.image.url.replace(/image\/upload/, `image/upload/c_thumb,g_face,h_${IMAGE_SIZE},w_${IMAGE_SIZE}`);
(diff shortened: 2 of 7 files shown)


sktt:
Does `secure` imply serve over https? If so why would you ever not want to use https?

Noviny:
@sktt secure does mean over https, and it was a legacy implementation we are working on moving away from.

MillerApps:
Any word on this being merged?

fields/types/cloudinaryimage/CloudinaryImageType.js
```diff
@@ -277,6 +277,21 @@ cloudinaryimage.prototype.getData = function (item) {
...
+	// to keystone after the model is registered, it will still be respected.
+	// Setting secure overrides default `cloudinary secure`
+	if ('secure' in this.options) {
+		this.__options.secure = this.options.secure;
+	} else if (keystone.get('cloudinary secure')) {
+		this.__options.secure = keystone.get('cloudinary secure');
+	}
+	return this.__options;
+};
+
```
cvn: You've got me curious. What's the use case for setting `secure`/`cloudinary secure` after the models have been registered?

fields/types/cloudinaryimage/CloudinaryImageType.js
```diff
@@ -277,6 +277,21 @@ cloudinaryimage.prototype.getData = function (item) {
 	return typeof value === 'object' ? value : {};
 };
 
+cloudinaryimage.prototype._originalGetOptions = cloudinaryimage.prototype.getOptions;
+
+cloudinaryimage.prototype.getOptions = function () {
+	this._originalGetOptions();
+	// We are performing the check here, so that if cloudinary secure is added
+	// to keystone after the model is registered, it will still be respected.
+	// Setting secure overrides default `cloudinary secure`
```
cvn: You've got me curious. What's the use case for setting `secure`/`cloudinary secure` after the models have been registered?

Noviny: It's mostly about the ordering of your `keystone.js` file. It is possible and valid to import your models and then after that line set `cloudinary secure`, however rather obtusely without checking again, you can end up not applying the option it looks like you set.

cvn: I see. Thanks for the explanation.

PR #4177, opened 2017-04-04
fixes #4158 by not calling join when there is no value

### LINKED ISSUE #4158: Strange behavior in creating page where there is a TextArray field
### Steps to reproduce the actual/current behavior

1. create a model like this:
```Javascript
const Test = new keystone.List('Test');

Test.add({
	tags: { type: Types.TextArray, required: false, initial: true },
	remark: { type: String, initial: true },
});
```
2.visit admin ui and open create page(ie. [http://localhost:3000/keystone/tests]), click 'Create Test'
3.Then click 'Cancel' button but the popup window does not close.And all other fields seems to be disabled until you add some tags.
4.If some tags are added first, then the page works normally.
### Environment

| Software         | Version
| ---------------- | -------
| Keystone         | 4.0.0-beta.5
| Node             | 6.10.1

Noviny:


fields/mixins/ArrayField.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
 	},
 
 	componentWillReceiveProps: function (nextProps) {
+		if (!nextProps.value) return;
```
intmainvoid: If nextProps.value doesn't exist (or is generally falsey) should we clear the component for the user? ie. `this.setState({ values: null });` or even `this.setState({ values: [] });` ?

sourav-prescriber-360: Was this issue resolved in keystone 4.0.0 . I am facing the issue if there is no value passed

PR #4198, opened 2017-04-08
Image previews like for CloudinaryImage

outluch:
## Description of changes
Added image preview in File field, like CloudinaryImage field

## Testing

eslint passed, data tests failed cause of timeouts etc.
Please maybe not approve pullrequest, but make such changes yourself in your dev env?

### DIFF
--- a/fields/types/file/FileField.js
+++ b/fields/types/file/FileField.js
@@ -12,6 +12,7 @@ import {
 	FormInput,
 	FormNote,
 } from '../../../admin/client/App/elemental';
+import ImageThumbnail from '../../components/ImageThumbnail';
 import FileChangeMessage from '../../components/FileChangeMessage';
 import HiddenFileInput from '../../components/HiddenFileInput';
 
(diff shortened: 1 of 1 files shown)


fields/types/file/FileField.js
```diff
@@ -120,6 +128,20 @@ module.exports = Field.create({
...
+		const { value } = this.props;
+
+		return (
+			<ImageThumbnail
+				component="a"
+				href={this.getImageSource()}
+				target="__blank"
+				style={{ float: 'left', marginRight: '1em' }}
+			>
+				<img src={this.getImageSource()} style={{ height: 90 }} />
```
wmertens: it would be cleaner to do `const src = this.getImageSource()` and then use that in both places

fields/types/file/FileField.js
```diff
@@ -201,12 +223,19 @@ module.exports = Field.create({
 			</div>
 		);
 
+		const imageContainer = (
+			<div style={this.isImage() ? { marginBottom: '1em' } : null}>
+				{this.isImage() && this.renderImagePreview()}
```
wmertens: Same with `isImage`

fields/types/file/FileField.js
```diff
@@ -65,6 +66,13 @@ module.exports = Field.create({
 	hasFile () {
 		return this.hasExisting() || !!this.state.userSelectedFile;
 	},
+	isImage() {
+		const href = this.props.value ? this.props.value.url : undefined;
+		return href && href.match(/\.(jpeg|jpg|gif|png)$/i) != null;
```
wmertens: I think you can add a `svg` in there as well…

PR #4285, opened 2017-05-12
Fixes some broken links

mradambeck:
## Description of changes
I noticed on http://keystonejs.netlify.com/documentation/configuration/server-options/ that many external links were being read as relative paths. I went through and added http/https to the problematic links.

## Testing
::thumbs up::


JedWatson:
Fantastic, thanks @mradambeck!

PR #4301, opened 2017-05-17
Ensure embedly type does not update with null values when no changes have been made

gwyneplaine:
## Description of changes
Added a conditional to the updateItem method in EmbedlyType.js 
Since all the data in this field is entirely dependent on the specified from path, the updateItem method should not be setting values when the fromPath has not changed. 

This is a symptomatic fix to a potentially more involved issue surrounding how the admin UI renders values for this field type. (In short, we may be passing through null values due to the fact that the values are not being captured wholly within the FormInput components used to render out extracted Embedly data).  

## Testing
Yes


Noviny:
I'm happy with this solution. It's similar enough to how other fields are maintaining this.

Issue #3696, opened 2016-11-14
Admin UI: Switching page to referenced model (temporary) breaks app


fabianbru:
### Steps to reproduce the behavior

Keystone version: 4.0.0-beta.3

1. Open detail view of one item of model referenced somewhere else (e.g.: "ExampleAuthor" of "Author")
2. Navigate to list view with referenced model used as defaultColumns (e.g.: "Post" model with "Author" model as defaultColumn)
3. Click on one of the referenced items (e.g.: "OtherExampleAuthor" of "Author")

### Expected behavior
Opens detail view of model.

### Actual behavior
Opens detail view of item from list view with the name field from the last viewed referenced model (in the example above: Clicking on "OtherExampleAuthor" in list view opens a "Post" model with an name field of "ExampleAuthor" -> all is mixed up) and outputs JS error in console:
```
fields.js:9421 Uncaught TypeError: values.split is not a function
    at t.loadValue (https://XYZ/keystone/js/fields.js:9421:52)
    at t.componentDidMount (https://XYZ/keystone/js/fields.js:9362:8)
    at r.notifyAll (https://XYZ/keystone/js/packages.js:2759:33)
    at r.close (https://XYZ/keystone/js/packages.js:3605:157)
    at r.closeAll (https://XYZ/keystone/js/packages.js:3795:49)
    at r.perform (https://XYZ/keystone/js/packages.js:3792:144)
    at o.perform (https://XYZ/keystone/js/packages.js:3792:61)
    at o.perform (https://XYZ/keystone/js/packages.js:3686:318)
    at Object.O [as flushBatchedUpdates] (https://XYZ/keystone/js/packages.js:3688:3)
    at r.closeAll (https://XYZ/keystone/js/packages.js:3795:49)
```
Also all field values are gone/empty.

In this state if I click on another menu item in top navigation, then nothing happens except a JS error in console:
```
fields.js:3360 Uncaught TypeError: Failed to execute 'removeChild' on 'Node': parameter 1 is not of type 'Node'.
    at t.componentWillUnmount (https://XYZ/keystone/js/fields.js:3360:17)
    at p.unmountComponent (https://XYZ/keystone/js/packages.js:3087:66)
    at Object.unmountComponent (https://XYZ/keystone/js/packages.js:3608:210)
    at p.unmountComponent (https://XYZ/keystone/js/packages.js:3088:29)
    at Object.unmountComponent (https://XYZ/keystone/js/packages.js:3608:210)
[...]

PR #4367, opened 2017-06-19
Fix issue with Admin UI: Switching page to referenced model

unsigned6:
Fix issue with Admin UI: Switching page to referenced model (temporary) breaks app.

Issue was happening because to `loadValue` was passing object, and because of this was failing in ternary operator :
```
values = Array.isArray(values) ? values : values.split(',');
```

changed this ti check if really string or array and only then use split if this param is type of string. In other case set empty array.

Tried to use this instead of npm package: for me it resolves issue described in https://github.com/keystonejs/keystone/issues/3696

But for me occurs another one when use repository from my github : occurs warnings in browser console of admin page
```
[Violation] Forced reflow while executing JavaScript took 35ms
```
I think this because in my repository is missing bundles which are there when install from npm and are placed in `keystone/admin/js` folder .
Could anybody please prepare me this bundles?

### DIFF
--- a/fields/types/relationship/RelationshipField.js
+++ b/fields/types/relationship/RelationshipField.js
@@ -98,7 +98,15 @@ module.exports = Field.create({
 				value: null,
 			});
 		};
-		values = Array.isArray(values) ? values : values.split(',');
+
+        if (Array.isArray(values)) {
+            values = values;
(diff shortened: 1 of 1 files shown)


unsigned6:
was doing this from version 4.0.0-beta.4 release

Issue #3676, opened 2016-11-07
Adding links to custom pages in admin UI


sharlet-wilson:
How to add links to my own routes and templates (which are not keystone lists) in the admin UI for 4.0 beta version (as was possible in the previous versions by adding an object in keystone.set("nav", {}))?

jstockwin:
@sharlet-wilson What happens if you do `keystone.set("nav", {}))` now? Are there any errors, or does it just not work?

sharlet-wilson:
It's been quite a while since I have used it, but as far as I remember, it simply doesn't work. The url does change but we need to reload the page to actually go to the link.

frapan:
I've got the same problem with 4.0 beta version: the custom link appears in the search bar but you need to reload to see the custom page.
Actually there are two different problems, as stated also in issue #1645:

1. you need to wrap the object inside an array as a workaround, otherwise it won't work
2. the custom page is shown only on manual reload of the page
  
I'm working on a solution on both problems and I'm going to make a pull request to fix this bug.

siffash:
@frapan any news on this issue? I'm using 4.0.0-beta.5, and still you need to reload the page to see the custom page.

frapan:
I've been very busy recently.
I think I might work on it in August, possibly next week.

Francesco

Il mer 25 lug 2018, 18:39 Anton <notifications@github.com> ha scritto:

frapan:
Hi @siffash, my PR about this issue has been merged in the 4.0.0-rc.0 release (see https://github.com/keystonejs/keystone/releases/tag/v4.0.0-rc.0).
You need to upgrade to at least that release.

siffash:
@frapan thank you, the upgrade fixed it. However external links are also added to the admin homepage, to the list of lists/models; it says "0 items", and when you click on it nothing happens except changing the url path.

Issue #1645, opened 2015-09-03
Add nav-links for custom pages


ajayvishwanath:
I have gotten around to understand some of the basic ways to configure keystone. For example, creating models, adding routes, middleware and changing which columns appear in Admin UI. They all work pretty smooth. :)

However, what has been bugging me for the past few days is that I am able to create my custom routes and templates but unable to provide any "links" to those routes, either in the homepage or the header navigation. The header navigation allows you to add only keystone lists.

How do I add a non-keystone list in my navigation panel (say an About Us page) and at the same time have the Admin UI enabled for the keystone lists?

creynders:
Hi @ajayvishwanath you can add any links you want to the navigation by adding an object instead of a string:

```js
keystone.set( "nav", {
  "content": [
    "posts",
    "post-categories",
    "enquiries",
    {
      label: "reports",
      key: "reports",
      path: "/api/reports"
    },
  ]
});

ajayvishwanath:
Hey, thanks for the reply!

I tried and it works pretty well. Thanks.

r3wt:
this no longer works? whats the new way to work around this issue?

ifiok:
@r3wt Did you find a workaround for this?

alexandrecardinal:
@r3wt @ifiok I just found a workaround
`keystone.set('nav', {
	venues: 'venues',
	customers: 'customers',
	users: 'users',
	'reservations': [
		{
			label: 'Reservations',
			key: 'reservations',
			path: '/reservations'
		}
	]
});`

In this case, this will give me a menu item called 'Reservations' leading me to '/reservations'. You HAVE to wrap your menu item with an array... this is probably a bug that should be fixed at some point.

alexandrecardinal:
@r3wt I really doubt you can do that without modifying the keystonejs module. I did some research and it seems doable in v4.0.0 but not the in v0.3.x

PR #4387, opened 2017-07-19
Fix custom nav links

frapan:
## Description of changes
There are 3 bugs that get fixed with this pull request:

1. There was an error in the unit test about the accepted languages which made "npm test" fail (see file /test/unit/lib/middleware/language.js)
2. In Keystone v0.3 it was possibile to add an object (or an array of objects) to the nav to make external links. In v4.0 it was only possibile to add an array of objects, forcing you to wrap the object inside an array. With this pull request you can add single objects again.
3. If you add an object (inside an array) with an external link to the nav, v4.0 shows the link but it is erroneously wrapped inside a Link component and therefore the router cannot handle it. This behaviour has been changed so that external links are represented simply by "a" tags, not by Link components.

## Related issues
The following issues are related to this pull request:

- #3676: this issue is about bug number 3 and could be closed
- #1645: this issue is about bugs number 2 and 3 and could be closed with a note: afaik it's not possibile to use an external link starting with /keystone/ beacuse it would be intercepted by the router.

## Testing


frapan:
How do I know what's wrong with deploy/netlify? The deploy preview failed but I don't think it's because of my changes.
If I click on the Details link, I get a 404 (I signed up to netlify with github account).
Can anyone help me?

frapan:
I've just realized that the netlify deploy went wrong beacuse of the commit about the nav section as simple objects.
Should I assume that this behaviour is not wanted anymore?
Should I remove that commit?
Or the netlify deploy should be fixed to let nav sections be simple objects?

frapan:
@JedWatson can you please answer me? Or could you please invite me to the slack channel in order to discuss about theese questions?

frapan:
@mxstbr can you help me with this pull request? I have no access to the netlify logs (see my comments above) and I don't know what the problem is.
I tried to reset the initNav.js as it was before my pull request but it keeps on giving me the same netlify failure.
How can I work it out? If you invite me to the Keystone slack, do you think I could find an answer?
Thanks.

test/unit/lib/middleware/language.js
```diff
@@ -69,7 +69,7 @@ describe('language', function () {
 		});
 		var expected = 'zh-CN';
 		var req = mockRequest({
-			acceptLanguage: 'zh-CN;1,en-US;q=0.8'
+			acceptLanguage: 'zh-CN;q=1,en-US;q=0.8'
```
blargity: Hey @frapan, sorry for the delay in getting back to you. The only thing I don't understand in your PR is the change to this file. Could you let me know what this is doing for us? I'd be happy to merge the other changes, just want to make sure I understand what's happening here.

frapan: Thank you @blargity  for your reply and I'm sorry as well for being late.
I had an error during tests because of that line, and I couldn't go on with the tests.
In fact, the syntax of acceptLanguage expects the weights of the languages to be written with ";q=", not simply ";".
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language as a reference.

Noviny:
@frapan Thankyou for some more patience!

I was wondering if you would mind documenting this feature, likely on [this](https://github.com/keystonejs/keystone/blob/master/docs/documentation/Configuration/AdminUI-Options.md) page. I like my hidden features in video games, but less so in my node modules. 😛

frapan:
@Noviny I agree with you. I've just added the documentation to the page you suggested.

PR #4390, opened 2017-07-22
Feature/location google maps

mrpatiwi:
## Description of changes

Added a Google Map view for `Location` field.

* It requires `google api key` for displaying it.
* Built with https://github.com/tomchentw/react-google-maps.
* Supports lazy and asynchronous loading of the vendor google maps scripts.
* Left click marks a spot and sets the lat/lng attributes.
* Right click on any place removes the marked spot.

preview

## Related issues (if any)
I think not.

## Testing

This test is not passing on my machine. Maybe I have not the correct setup.

```txt
1) language must allow Accept-Language selection:

      AssertionError: "en-US" must be equivalent to "zh-CN"
      + expected - actual

      -en-US
      +zh-CN

      at Context.<anonymous> (test/unit/lib/middleware/language.js:79:34)
```

### DIFF
--- a/fields/types/location/LocationField.js
+++ b/fields/types/location/LocationField.js
@@ -1,5 +1,7 @@
 import _ from 'lodash';
 import React from 'react';
+import { withGoogleMap, GoogleMap, Marker } from 'react-google-maps';
+import withScriptjs from 'react-google-maps/lib/async/withScriptjs';
 import Field from '../Field';
 import CollapsedFieldLabel from '../../components/CollapsedFieldLabel';
 import NestedFormField from '../../components/NestedFormField';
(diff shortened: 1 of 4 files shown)


Noviny:
This looks sweet. So happy to see you documented the usage as well.

A couple of small things. What happens when no google maps api key is provided? I think we want to section off the map itself much more firmly, so I would move all logic around map state into `AsyncGoogleMaps`, including everything going in to the `renderMap` function.

Then line 378 can become:

```
{browserApiKey ? <AsyncGoogleMaps browserApiKey={browserApiKey} /> : null}
```

Does this update make sense to you, to ensure maps are hidden when they should be?

tomcardoso:
👍
FWIW, would love to see this merged in.

radames:
it works like a charm, just need to use the old react-google-maps version 6.3.0, the newest version seems not working

PR #4397, opened 2017-07-28
Fix: model with required field that defaults to 0 can't be saved

jyrkive:
## Description of changes

This pull request fixes a bug that makes it impossible to create a document if the associated model has a **numeric** field that's **required** and has a **default** value of **zero**.

This minimal model reproduces the problem:

```js
var keystone = require("keystone")

var DocumentList = new keystone.List("Document", {
  schema: {collection: "documents"}
})

DocumentList.add({
  num: {type: Number, required: true, default: 0}
})

DocumentList.register()
```

Attempting to create such a document in Keystone gives an error: "Num is required".

The problem, as you might guess, is that zero is a falsy value in JavaScript. The affected check in KeystoneJS code is supposed to check if the getter returns a value, but instead it checks if it returns a **truthy** value. The fix is a simple one-liner.

Obviously the same problem also occurs with boolean fields which default to false.

## Testing

One test fails. It looks unrelated to my changes, though.

> 1) language must allow Accept-Language selection:
>
>     AssertionError: "en-US" must be equivalent to "zh-CN"
>     \+ expected - actual
>
>     -en-US
>     \+zh-CN
>
>     at Context.\<anonymous\> (test\unit\lib\middleware\language.js:79:34)

### DIFF
--- a/fields/types/boolean/BooleanType.js
+++ b/fields/types/boolean/BooleanType.js
@@ -36,7 +36,7 @@ boolean.prototype.validateInput = function (data, callback) {
 
 boolean.prototype.validateRequiredInput = function (item, data, callback) {
 	var value = this.getValueFromData(data);
-	var result = (value && value !== 'false') || item.get(this.path) ? true : false;
+	var result = (value && value !== 'false') || typeof item.get(this.path) === 'boolean' ? true : false;
 	utils.defer(callback, result);
 };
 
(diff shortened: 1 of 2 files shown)


Noviny:
@jyrkive Thanks for noticing this problem, and providing a solution. Unfortunately I don't think this is the solution we want, as we want to be more specific. Could you rewrite the check as. `typeof item.get(this.path) === 'number'`, and the same for boolean. We might close this PR, add the change, and credit you in our release notes.

Cheers,
Ben

jyrkive:
Hello, Ben. Good to see that you're considering to merge the PR. :)

I rewrote the checks to use typeof comparisons.

PR #4407, opened 2017-08-05
it should be a <a/>

gggllm:
it should be a <a/>

## Testing


gggllm:
actually I don't know how to use react so I am not sure if it is a problem with type check

PR #4408, opened 2017-08-05
it should be a <a>

gggllm:
## Testing

### DIFF
--- a/fields/types/file/FileField.js
+++ b/fields/types/file/FileField.js
@@ -126,7 +126,7 @@ module.exports = Field.create({
 		return (
 			<div>
 				{(this.hasFile() && !this.state.removeExisting) ? (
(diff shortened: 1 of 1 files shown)


Noviny:
@gggllm Thanks for the contribution! I don't think this does what you think it will do though.

The component prop will be passed down to [FileChangeMessage](97a036d5b52d2239c866aa9c46c05e27d96d2159) which will pass it down to [FormInput](https://github.com/keystonejs/keystone/blob/97a036d5b52d2239c866aa9c46c05e27d96d2159/admin/client/App/elemental/FormInput/index.js) but neither of these will render a component they have been given.

Noviny:
Closing this

gggllm:
@Noviny  Actually it works exactly as I expected, it has been a long time since I have submitted the PR, so I cannot remember the detail.

Issue #3979, opened 2017-02-08
blank page login, qs bug from mapped custom domain Uncaught SyntaxError: Unexpected string


jjhesk:
resurrected issue from #3642

### Steps to reproduce the behavior
just login

### Expected behavior
can see the login screen
### Actual behavior

looks like i have the same issue with this 4.0.0-beta5 deploy @jstockwin @joeinnes @conrad-vanl on heroku. the issue doesnt appear on local deploy from testing server Ubuntu-16.4.10 tho. currently is not working.. http://bms.heskeyo.com/admiral/signin

Uncaught SyntaxError: Unexpected string
..

s)"undefined"!=typeof r&&n(r)}}var t=this,n=this.store.dispatch
Uncaught Error: Cannot find module 'qs'

bassjacob:
I'm not sure what would be causing qs not to be requireable, it's a direct dependency of keystone. Can you run `npm cache clear; rm -rf node_modules; npm i` and see if that fixes the issue?

jjhesk:
@bassjacob according to the local test project and the project runs fine without any errors. the only issue happens only on heroku deployment. note that the deployment will clear all cache for each compile deploys. I will check with ur notes to do so and see if there are any chance to get things clear out..

jjhesk:
@bassjacob 
here is what i get when i ran `npm ls qs`
```
galleria-loopback-a73@2.3.5 /app
+-- body-parser@1.15.2
| `-- qs@6.2.0 
+-- loopback@3.1.0
| +-- loopback-datasource-juggler@3.1.0
| | `-- qs@3.1.0 
| `-- strong-remoting@3.1.1 (git+https://github.com/strongloop/strong-remoting.git#35fd6690b73838fde6925c0f99186afaf5f4102e)
|   `-- qs@6.3.0 
+-- loopback-component-storage@3.0.0-alpha.1 (git+https://github.com/strongloop/loopback-component-storage.git#c73dbc43b6f20113708353f5b678d63176f8603d)
| `-- pkgcloud@1.3.0
|   +-- qs@1.2.2 
|   `-- request@2.40.0
|     `-- qs@1.0.2 
+-- loopback-datasource-juggler@2.53.0
| `-- qs@3.1.0 
+-- request@2.76.0
| `-- qs@6.3.0 
`-- stripe@4.14.0 (git+https://github.com/stripe/stripe-node.git#d2660aff03851f2c259b9419323c1552ff9c484d)
[...]

PR #4432, opened 2017-09-08
Fix attempt for issue #3979

t100n:
Set semicolons=true in the build command that is passed to uglifyjs.
Added 'aphrodite/no-important' to the required packages in admin/client/packages.js.

Issue #3979 :
admin/public/js/packages.js generates a Uncaught SyntaxError: Unexpected string when minified, and fails to load the admin interface.
After some digging I found that uglifyjs compresses the typeof checks into the following format "undefined"==typeof x which triggers a syntax error after passing through another minification process when deployed.


Noviny:
@t100n So glad someone got to the bottom of this rabbit hole. We haven't encountered the issue, so I can't validate that this fixes it, but I take that it fixed the issue for you?

Also, why was adding aphrodite relevant to resolve this?

t100n:
@Noviny indeed this was the solution for my problem.

I was forced to add 'aphrodite/no-important' because it was triggering a missing module while loading the admin UI, during my testing of the new packages.js file. So it was just to make it work.

Noviny:
@t100n Thanks for providing this fix.

Zedonboy:
@t100n where will i find all these "semicolons = true".... I should tamper with the npm package(keystone)

PR #4444, opened 2017-09-25
fix:module-name in import statement in EditForm.js

petertdinh:
## Description of changes
Fix the `ConfirmationDialog` import statement in Editform. It works fine but the extra `./` is extraneous and redundant.


JedWatson:
Thanks @petertdinh

Issue #1241, opened 2015-03-23
Show revisions in the Admin UI


JedWatson:
We now have a list history feature (see #1240)

The current revision, and a way to browse revision history, should be revealed in the Admin UI somewhere.

Should be pretty easy to do with a new API and some React magic.

tivac:
I thought about this for a bit. 

Something kinda like this seem reasonable? It's not exactly sexy, but it is pretty functional IMO.
![revisions](https://cloud.githubusercontent.com/assets/49545/6796676/9f534074-d1b4-11e4-9886-49e6894ab3dc.png)
_(please ignore my terrible spacing issues)_

albertogasparin:
What would be really cool is a [time slider like on Etherpad](https://etherpad.mozilla.org/ep/pad/view/wZUhcvCCWr/latest).
We don't need the word by word changes, but with React is quite easy to change the content as you click through the saved revisions.

medv:
Slider or not, what @tivac sketched up there is still good for quick glance.

mxstbr:
We're closing all feature requests to keep the issue tracker unpolluted. I've moved this to ProductPains: https://productpains.com/post/keystonejs/show-revisions-in-the-admin-ui!

PR #4398, opened 2017-07-31
[WIP] History UI

bbqsrc:
## Description of changes

Introduces a History button for models who have `history: true`. This allows a user to track historical revisions and revert an instance if necessary.

This was an immediate need of a client and I am publishing this code here for anybody who may find it useful.

screen shot 2017-07-18 at 19 39 22

## Related issues (if any)

#2212, #1241, etc. 

## Testing

- [ ] Please confirm `npm run test-all` ran successfully.

PR #4445, opened 2017-09-26
feature/history-ui-content-rollback

petertdinh:
## Description of changes
Add History UI in the form of a new `Screen` component as `Revisions`. The UI will navigate the admin user to `Revisions` if `noedit` is false and `history` is true. Going to `Revisions` will show a list all past versions of that particular item. If a `ListItem` is clicked, it will show and compare the differences between the current version and the rollback version and give the user the ability to rollback to that particular version. 

image

image

Refactored `history.js` to have full property names. This will make things a bit clearer to devs going over the revisions collections. I made it fully backwards compatible with the old property names. I also added a 'isModified()' check to only create a revision if the item had changes.

Most of my design choices have been borrowed from Contentful's history UI.

## TODO
-Mobile styling for the new Revision view.
-Write e2e tests for the new Revision view.
-Update the `linter` to include babel static class properties

## Testing

- [] Please confirm `npm run test-all` ran successfully.

Running into some linting errors in files that I didn't make changes to.

## Addresses
#1241


epinapala:
@petertdinh can you please resolve the merge conflicts?

stennie:
Related PR: #4398 ([WIP] History UI)

PR #4447, opened 2017-09-27
Fix eslint errors when running e2e tests

vitalbone:
## Description of changes
Commenting out and removing redundant imports that are making the `test-all` test fail due to the eslint rules.

The offending lines have been commented out (rather than removed entirely) as the functionality they facilitate may be reintroduced later on.

## Testing
Tests are still not running successfully due to another issue related to https://github.com/keystonejs/keystone/pull/3342 but at least this PR addresses the 5 failing tests that were preceding these errors.

### DIFF
--- a/www/components/Navbar/Item.js
+++ b/www/components/Navbar/Item.js
@@ -1,6 +1,6 @@
 import React, { PropTypes } from 'react';
 import Link from 'gatsby-link';
-import theme from '../../theme';
+// import theme from '../../theme';
 
 export default function Item ({ depth, title, url }) {
 	return (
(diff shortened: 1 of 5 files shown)


www/components/Navbar/Item.js
```diff
@@ -1,6 +1,6 @@
 import React, { PropTypes } from 'react';
 import Link from 'gatsby-link';
-import theme from '../../theme';
+// import theme from '../../theme';
```
Noviny: What's the reason to comment these out instead of removing them?

vitalbone: Sorry, I merged and then the UI changed to reflect your comment. Fair call. Have created another PR to clean this up - https://github.com/keystonejs/keystone/pull/4448

Issue #4245, opened 2017-04-29
4.0.0-beta.5: Storage.Adapters.FS and deleting/overwritting files


Allmektiger:
Hello.
I want to make avatar loading for users in local filesystem, but have a little headache with it...

`var myStorage = new keystone.Storage({
    adapter: keystone.Storage.Adapters.FS,
    fs: {
        path: 'public/images/users',
        publicPath: '/images/users',
        whenExists: 'overwrite'
    },
});`
...
`avatar: { type: Types.File, storage: myStorage },`

Firstly, loading in admin UI works fine, but after removing with "Remove File" link and saving file doesn't deleting from disk. Is this expected behaviour or issue?
I know that deleting avatars from admin panel is not necessary, but this may be actually for other content with file uploading/removing from admin UI.

Secondly, i need in possibility to reload avatars. Easiest way that i see is "**whenExists: 'overwrite'**" option. But it require the same naming of files for users. For example by email: **em@a.il.png**. Unfortunately, in current documentation i see:
> The generateFilename() method no longer receives a reference to the item being updated, we may introduce this in a future update.

Are there any alternative ways to naming file with one name for one user?

Thirdly, are there any callbacks for uploading process? For example, scaling and crop big image after saving. I know about cloudinary, but i really need in local storaging.

Fourthly, i want to implement type validation, like **allowedTypes** in old **LocalFile**. Is it possible?

I was really saddened by deprecating old LocalFile field type. Is there an easy way to restore this type in my local repository until new full versions of File type appear, without downgrading to keystonejs 0.3.17?

CruelMoney:
I'm having some problems updating existing files as well. 
I try to overwrite the existing file when updating the file, but it seems that the "whenExists" attribute does not have any effect. 

My API update endpoint :
```
FileData.model.findById(req.params.id).exec(function(err, item) {
    if (err) return res.apiError('database error', err);
    if (!item) return res.apiError('not found');

    var data = (req.method == 'POST') ? req.body : req.query;

    item.getUpdateHandler(req).process(data, function(err) {

      if (err) return res.apiError('create error', err);
       
      item.url =  ('/uploads/files/'+item.file.filename)
      res.apiResponse(item);

    });
  });
```

And the model configuration:
```
var myStorage = new keystone.Storage({
    adapter: keystone.Storage.Adapters.FS,
    fs: {
        path: keystone.expandPath('../public/uploads/files'), // required; path where the files should be stored
[...]

PR #4466, opened 2017-10-17
Fixed overwrite in Adapters.FS

Aldlevine:
## Description of changes

This change solves an issue that would prevent keystone.Storage.Adapters.FS from overwriting a file when `whenExists == 'overwrite'`. For example, previously the below configuration would return an `EEXIST` error when the file path being written already exists.

```js
const storage = new keystone.Storage({
  adapter: keystone.Storage.Adapters.FS,
  fs: {
    path: path,
    publicPath: path,
    whenExists: 'overwrite',
    generateFilename: keystone.Storage.contentHashFilename,
  }
});
```

This issue is resolved by enabling the `clobber` option in [`fs-extra.move`](https://github.com/jprichardson/node-fs-extra/tree/1.0.0#movesrc-dest-options-callback).

_Note: If updating to later versions of fs-extra the `clobber` option must be changed to `overwrite`._

## Related issues (if any)

#4245 (See comment 2)

## Testing

_Same tests pass/fail as do in current master branch._

### DIFF
--- a/lib/storage/adapters/fs/index.js
+++ b/lib/storage/adapters/fs/index.js
@@ -97,7 +97,9 @@ FSAdapter.prototype.uploadFile = function (file, callback) {
 		filename = sanitize(filename);
 		debug('Uploading file with filename: %s', filename);
 		var uploadPath = path.resolve(options.path, filename);
-		fs.move(file.path, uploadPath, function (err) {
+		var fsOptions = {};
+		fsOptions.clobber = options.whenExists === 'overwrite';
+		fs.move(file.path, uploadPath, fsOptions, function (err) {
 			if (err) return callback(err);
 
 			// TODO: Chmod the file.


JedWatson:
This looks good @Aldlevine, thanks!

Pyav123:
Hello,
   Anybody still facing this issue? This solution is not working for me. 
There two things in this. 

First, when I installed keystone(keystone@4.0.0-beta.5 ) I didn't get the fix made https://github.com/keystonejs/keystone/commit/f262c6762f2b98badf1669f1678d752e4a802dfc  for node_modules\keystone\lib\storage\adapters\fs\index.js .

So, Second I tried to copy and paste the adapter fix code (node_modules\keystone\lib\storage\adapters\fs\index.js )  into my node modules but still it isn't working.  Can anyone suggest me on this ? I want to upload and overwrite files such as html,css,js etc in CMS.

Issue #4480, opened 2017-10-24
Bind Ctrl + s event to save functionality on 'Edit Post' screen


zacharysang:
### Expected behavior

When the user is editing a post of theirs, they should be able to press ctrl + s to save the content

### Actual/Current behavior

When the user is editing a post of theirs, ctrl + s simply performs the default browser behavior and prompts to save the html version of the page.

### Steps to reproduce the actual/current behavior

* Select a post from the admin page and choose to edit the post
* press the ctrl and then the 's' keys

### Environment

| Software         | Version
| ---------------- | -------
| Keystone         |  master branch of the repo (npm linked)
| Node             | 6.11.2

zacharysang:
^ I am happy to work on this

PR #4490, opened 2017-11-01
#4480 - Added listening for ctrl+s in EditForm

zacharysang:
## Description of changes
Added a listener for keypress event on the EditForm component. This listener checks that ctrl+s keys are pressed and if so, triggers the updateItem (save) function.

## Related issues (if any)
#4480 

## Testing


stennie:
@zacharysang  Apologies for the delayed feedback.

Can you confirm which browser(s) you tested this with?

This appears to be work with Firefox 61.0 (macOS) but doesn't override the default key event when I tested in Chrome 67.0.3396.87 (macOS).

Thanks,
Stennie

PR #3865, opened 2017-01-09
add preview image to file field

gemscng:
## Description of changes
Add preview to CMS for file field

PR #4509, opened 2017-11-13
(feat) implements image thumbnails for the File type

cadriel:
Implements an image thumbnail for type `File`. Depends on a new schema property called `showThumb`. This improves upon the existing PR's in place that have stalled.

CSS has been implemented to ensure larger format images don't break layout.

Intent is to keep this simple for now until more complex solutions re: image middleware are sorted.

Master currently has e2e and unit tests failing from a prior merge. Can these be resolved so I can rebase?

@JedWatson

### DIFF
--- a/fields/types/Type.js
+++ b/fields/types/Type.js
@@ -29,6 +29,7 @@ var DEFAULT_OPTION_KEYS = [
 	'collapse',
 	'dependsOn',
 	'autoCleanup',
+	'thumb',
 ];
 
 /**
--- a/fields/types/file/FileField.js
+++ b/fields/types/file/FileField.js
@@ -32,11 +32,11 @@ module.exports = Field.create({
 		label: PropTypes.string,
 		note: PropTypes.string,
 		path: PropTypes.string.isRequired,
+		thumb: PropTypes.bool,
 		value: PropTypes.shape({
 			filename: PropTypes.string,
 			// TODO: these are present but not used in the UI,
 			//       should we start using them?
-			// showThumbnail: PropTypes.bool
 			// filetype: PropTypes.string,
 			// originalname: PropTypes.string,
 			// path: PropTypes.string,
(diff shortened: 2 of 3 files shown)


cadriel:
@wmertens

wmertens:
:+1: looks very clean, maybe @jossmac can merge?

cadriel:
Upon merge, probably best to close the following;
https://github.com/keystonejs/keystone/pull/4198
https://github.com/keystonejs/keystone/pull/3865

There might be other related issues and PR's too.

cadriel:
Looking at this further, it might make more sense to have the showThumb property on the field itself, rather than defining it on the adapter.

That way you can have multiple image fields in one model - and only show the thumb where wanted. I'll take a quick gander and see if I can implement that..

cadriel:
@wmertens @jossmac @JedWatson 

Might want to have another quick look. Instead of defining the showThumb property on setup of the Storage adapter, you now set it per field.

This makes more sense because a) it doesn't save it to the database now and b) you can set it per field.

```
Model.add({
  image: { type: Types.File, thumb: true, storage: myStorage },
  ...
```

cadriel:
@wmertens @JedWatson @jossmac 

Is this, or any other PR going to be looked at soon? It's quite disconcerting to have no movement on any of our PR's or Issues over this length of time - especially given the investment thus far into the project.

jossmac:
@cadriel looks good mate

jossmac:
Will need @Noviny or @JedWatson to release

htor:
allright! so glad that this made it in. is this feature documented somewhere?

cadriel:
No docs;

But - as per example above, its easy to apply;

```
Model.add({
  image: { type: Types.File, thumb: true, storage: myStorage },
  ...
```

This isn't currently smart enough to use a thumbnail image, or set dimensions. It'll just do its best to present the original file in keystone.

htor:
@cadriel 
which i think is what many people need right know. just being able to look at the images you upload into your cms is absolutely essential! after reading issues after issues about this not being possible (any longer) i got the feeling it was not being a priority. i got it working just fine.

Issue #4332, opened 2017-05-26
You can't fully delete relationships


ellipticaldoor:
I'm on the last commit on master but also this bug is present o the latest beta.

When you have a relationship you can't delete, it just ignore's it.

On a one to may relationship you can delete all of them except the last one.

Noviny:
Can you provide a replication case? We have not encountered this bug and use relationships extensively.

JacobPVoronin:
Same issue. @Noviny here's the case:

Regular `Relationship` field in KeystoneJS (`categories` in this example):

    Post.add({
        author: { type: Types.Relationship, ref: 'User' },
        categories: { type: Types.Relationship, ref: 'PostCategory', many: true }
    });

Initially it's empty, but when I add some value to it, save it and then clear it (so that field is empty again), saving this Model won't update `categories`.

`console.log(this.categories)` in `Post.schema.pre('save', ...` shows that `this.categories` has previous value, while **it should be empty**.

Any ideas on this behavior?

ryannhg:
Hey there,

Installing Keystone `v4.0.0-beta.5` fixed this bug for me.

It looks like the `v4.0.0-beta.8` version has this issue for both:

- Single Relationship Fields
- Many Relationship Fields

As a temporary workaround, I would stick to an earlier beta version like `v4.0.0-beta.5`

Good luck!
Ryan

Twansparant:
Is there a workaround for v4.0.0-beta.8 without editing the **keystone/fields/types/relationship/RelationshipType.js** core file?

stennie:
@Noviny Looks like this bug was introduced with commit 5df6846: when all relationships are removed there will be no references in the data object.

Should this commit be reverted or is there some other context for the change?

Noviny:
The context for the change is that the relationship field being absent from a submission (undefined) should not automatically delete it.

This was causing a bug when calling update where the field was absent it would get deleted unintentionally.

It should definitely be valid to pass no data for a field when saving and not erase it.

PR #4530, opened 2017-12-06
(fix) relationship saves

cadriel:
(fix) Clearing relationships was not applying to the db
(fix) Relationships were re-saving on every save in admin

I'm not sure of the original intent here, because there was a test failing that indicated the code expected the old value to return as type ObjectId. In my testing, a single object was returned that had a string in it - so I've updated the test and code to reflect that.

In any case, relationships of both single and many types now clear properly. If you don't change a relationship, it doesn't get re-saved.

No tests were broken during the course of this change.

### DIFF
--- a/fields/types/relationship/RelationshipType.js
+++ b/fields/types/relationship/RelationshipType.js
@@ -212,43 +212,45 @@ relationship.prototype.inputIsValid = function (data, required, item) {
 /**
  * Updates the value for this field in the item from a data object.
  * Only updates the value if it has changed.
- * Treats an empty string as a null value.
  * If data object does not contain the path field, then leave the field untouched.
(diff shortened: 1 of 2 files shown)


dam1:
I guess it's to prevent saving an object and losing the relationship when all the field are not loaded.

should we add a "clear-field" value in the select of the admin ? that would solve both problems

cadriel:
How do you mean? In what situation would the related fields not be loaded for a model?

stennie:
Thanks for the PR (and apologies for the delay in feedback).

A fix for this issue was included in Keystone `4.0.0-rc.0` (by way of #4332).

If you are still seeing any related problems in the latest release, please create a new GitHub issue with details.

Thanks,
Stennie

Issue #3160, opened 2016-07-14
Field type 'Datetime' - format customization missing


chiss2906:
There is no working way to specify the format of Datetime. Also the placeholder is hardcoded in DatetimeField.js `<FormInput name={this.props.paths.time} value={this.state.timeValue} placeholder="HH:MM:SS am/pm" onChange={this.timeChanged} autoComplete="off" />`.

jstockwin:
http://keystonejs.com/docs/database/#fieldtypes-datetime
You can format your Datetime field when adding it to your model:
`year: { type: Types.Datetime, format: 'YYYY' }` should work, for example.

mxstbr:
Formatting works perfectly fine, but should definitely affect the placeholder!

wmertens:
The input format of the DateTime field is hardcoded, no? https://github.com/keystonejs/keystone/blob/master/fields/types/datetime/DatetimeField.js#L14-L15

jonathan-l-rivera:
Shouldn't `this.parseFormatString = this.formatString`, or just validate against the formatString? Otherwise you'd also need to set parseFormat. DateTime validation should function the same as Date, seems to be handled differently.

https://github.com/keystonejs/keystone/blob/master/fields/types/datetime/DatetimeType.js#L20
https://github.com/keystonejs/keystone/blob/master/fields/types/date/DateType.js#L76

It's also hardcoded incorrectly in v3.x. This is a critical issue for me, and an easy fix.
https://github.com/keystonejs/keystone/blob/v0.3.x/fields/types/datetime/DatetimeType.js#L86

asliwinski:
@jstockwin 
`year: { type: Types.Datetime, format: 'YYYY' }`  - doesn't work, renders 2 fields, date and time

Noviny:
@asliwinski I question using DateTime for a year value. I would likely store this as a number field, as this is the information you are most interested in representing.

The problem with storing it as a DateTime is that you make it timezone-specific, which a simple year value does not want.

If you want to use a richer field, Date is also better for this.

If anyone wants to make a PR so format affects the placeholder, that would be grand.

hari419:
when i use Types.Datetime , it showing below error
Error: Invalid value for schema path `date.type` in `unavailability`.

gsteinert:
I'm working on a PR for the first part of this issue (lack of control over the format in the admin UI)

It looks like the issue stems from the format property being used to read from and save to the database but being ignored completely by the admin UI field.

[...]

PR #4539, opened 2017-12-31
Datetime formats

gsteinert:
## Description of changes

A few changes to DatetimeType to address #3160 

### Major Changes

`format` option removed. In its place `dateFormat`, `timeFormat `and `tzFormat `created. This allows the formats used in DatetimeType to be shared with DatetimeField. Now the display format specified in the three new options is used for validation and the admin UI fields.

`parseFormat `option behaviour changed.

Previously, a format specified in `parseFormat `would be the only acceptable format for validation purposes. With an incorrectly specified `parseFormat `option, you could end up in a situation where a date chosen from the datepicker wouldn't validate.

Now, any format specified in `parseFormat `is added to the list of acceptable formats. This list already includes the default list and the format specified in `dateFormat`, `timeFormat `and `tzFormat`. `parseFormat `now only needs to be set if you want to accept additional formats not already in the list - in many cases it can be omitted, even with a custom display format.

### Other changes

The format specified in `dateFormat`, `timeFormat `and `tzFormat `is used to format the admin UI fields.

The placeholder in the time field in the admin UI now reflects `timeFormat`.

Documentation updated

## Testing

No new failures introduced. 7 pre-existing failures in unrelated areas remain.

### DIFF
--- a/fields/types/datetime/DatetimeField.js
+++ b/fields/types/datetime/DatetimeField.js
@@ -20,11 +20,6 @@ module.exports = Field.create({
 
 	focusTargetRef: 'dateInput',
 
-	// default input formats
-	//dateInputFormat: 'YYYY-MM-DD',
-	//timeInputFormat: 'h:mm:ss a',
-	tzOffsetInputFormat: 'Z',
-
 	// parse formats (duplicated from lib/fieldTypes/datetime.js)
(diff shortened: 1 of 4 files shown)


gautamsi:
why add breaking changes where you could have the format option alongside other options. you can check if both options are provided and throw error. 

not removing format option would have been better for people already using it.

PR #4545, opened 2018-01-07
Fix for known security vulnerability in marked 0.3.6 dependency

vitalbone:
## Description of changes
Github found a known security vulnerability in one of our dependencies, which was fixed with a patch upgrade, so this moves keystone's dependency to the recommended version.

## Testing
- [x] `npm run test-all` ran successfully


vitalbone:
This seems to be failing in CI on an unrelated issue to do with expected password strength.

PR #4546, opened 2018-01-08
Publishing workflow

Gaurav490:
The changes have been commited for filtering the files while importing from keystone.importer. You can pass a object with either of two properties, {match : "*.js"} or {exclude:'*.js'}. These will be include or exclude the files accordingly.


sheldon-sminq:
In the publishing schemaPlugin the 'User' model is directly referenced in the relationship, can this be changed to 

```
approvedBy: { type: Types.Relationship, ref: 'User', index: true, noedit: true }

approvedBy: { type: Types.Relationship, ref: keystone.get('user model'), index: true, noedit: true }
```

Gaurav490:
@sheldon-sminq  Yes, I will change and commit the changes.

sheldon-sminq:
Another instance where the user api is referenced.

```
List.prototype.getPermissions = function (callback) {
        var userPerms = {};
        var userApi = keystone.
        let url = Keystone.adminPath + '/api/users/' + Keystone.user.id;

        xhr({
                url: url,
                responseType: 'json',
        }, (err, resp, data) => {
                if (err) return callback(err);
                // Pass the data as result or error, depending on the statusCode
                if (resp.statusCode === 200) {
                        userPerms.isContributor = data.fields.isContributor || false;
                        userPerms.isEditor = data.fields.isEditor || false;
                        userPerms.isAuthor = data.fields.isAuthor || false;

                        callback(userPerms);
                } else {
                        callback(data);
                }
        });

        // Get User permissions
        /* Keystone.User.model.findById(Keystone.user.id).exec(function (err, user) {
                if (user) {
                        if (user.isContributor) {
                                userPerms.isContributor = user.isContributor || false;
                                userPerms.isEditor = user.isEditor || false;
                                userPerms.isAuthor = user.isAuthor || false;

                                callback(userPerms);
                        }

                }
        });*/


};

```
This will fail if the user model is a different name, can the model query be used instead of the api call?

Gaurav490:
@sheldon-sminq  Hi Sheldon, I will check and will let you know.

PR #4559, opened 2018-01-29
Admin UI not showing error messages when deleting Issue number 4403

Gaurav490:
## Testing


laurenskling:
You are hacking into the response being empty while in `200`. This is because a failed `remove` doesn't actually give an error status from the API. 

I think we should fix this first.

laurenskling:
Also, you are using `alert` just like the remove does in other cases. I think we should look into how to transfer all these errors into the flash messages. These alerts look terrible.

laurenskling:
https://github.com/keystonejs/keystone/pull/4832 this should actually return an error from the API

Issue #4494, opened 2017-11-01
Session maxAge in cookie signin options is not respected


hreimer:
### Expected behavior

Following [this closed issue #1895](https://github.com/keystonejs/keystone/issues/1895) I was expecting to set the maxAge property of the session cookie, after which the user is automatically signed out, and because its 2017 (and nowhere properly documented, [in the keystonejs npm package in lib/session.js:55 we will find the current syntax](https://github.com/keystonejs/keystone/blob/41c6e1657bf04aa269dd3c1ee3477185b71318bb/lib/session.js#L55)) and the syntax from 2015 doesn't work anymore it is now done for v 4.0.0 like this in the main keystone.js start file:

`'cookie signin options' : {
		'maxAge': 5 * 60 * 1000 //5min
	},`

to set the session timeout to 5min. 

### Actual/Current behavior

Yes, the keystone.uid cookie is set accordingly to the desired maxAge, however there is also a "this.sid" cookie, that get's initialized whenever keystone starts up and has an unlimited (or none) expiry date, and will only get deleted upon browser close since it's a session cookie. If the keystone.uid cookie is deleted, this.sid is still valid and active and therefore the proposed maxAge configuration (which itself should convert the keystone.uid Session cookie to a cookie with expiry date) is useless.

(To be noted, there is no difference in the actual behaviour when I use the built-in express session store or the recommended external "session store" via mongo / connect-mongo)

### Steps to reproduce the actual/current behavior

Just add 

`'cookie signin options' : {
		'maxAge': 5 * 60 * 1000 //5min
	},`

in your keystone.js file, log in and open the developer console in Chrome, Application>Storage>Cookies><yourhost:port> to observe it and try to interact with keystone after the cookie has expired and is deleted, while this.sid is still present.

### Proposal for a possible solution:

in the same lib/session.js file in the keystonejs package, simply add one line:

`			req.session.cookie.maxAge = cookieOpts.maxAge;
`
just above [https://github.com/keystonejs/keystone/blob/41c6e1657bf04aa269dd3c1ee3477185b71318bb/lib/session.js#L60](https://github.com/keystonejs/keystone/blob/41c6e1657bf04aa269dd3c1ee3477185b71318bb/lib/session.js#L60)  to reset that initial Session cookie every time a user signs in and the maxAge property is set.

I am not experienced in PRs and if this is a valid solution or are there other approaches, maybe somebody with more experience in keystonejs could comment and approve/disapprove?  

### Environment

[...]

PR #4565, opened 2018-02-05
Added session max age cookie to match keystone.uid cookie expiry

millsy:
## Description of changes
Add the session max age cookie (defaults to name this.sid) to match the expiry of the keystone.uid cookie. This correctly then expires when the keystone.uid cookie expires forcing the user to need to re-authenticate.

## Related issues (if any)

As described in Issue #4494 

## Testing

npm test run-all doesn't work - fails on **npm test && npm run test-e2e-bg** but this occurred before the change was made, so it's likely my env causing build issues.

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -57,6 +57,7 @@ function signinWithUser (user, req, res, onSuccess) {
 				httpOnly: true,
 				maxAge: 10 * 24 * 60 * 60 * 1000,
 			});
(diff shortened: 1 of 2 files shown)


stennie:
Thanks @millsy! I cherrypicked and merged your first commit after testing. Updates to dependencies should be done separately, so I skipped the Mongoose bump.

Regards,
Stennie

PR #4601, opened 2018-03-19
Adding in the missing DayPicker-WeekdaysRow

jsundquist:
## Description of changes
Updating the admin.less file as it is currently missing the DayPicker-WeekdaysRow. Since this class is missing, the days are not aligning correctly with the calendar below

## Testing

### DIFF
--- a/admin/public/styles/keystone.less
+++ b/admin/public/styles/keystone.less
@@ -116,7 +116,7 @@
 }
 
 .DayPicker-WeekdaysRow {
-	display: tahle-row;
+	display: table-row;
 }
 
 .DayPicker-Weekday {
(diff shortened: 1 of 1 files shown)


admin/public/styles/keystone.less
```diff
@@ -113,9 +113,10 @@
 
 .DayPicker-Weekdays {
 	display: table-header-group;
-	// > div {
-	//	 display: table-row;
-	// }
+}
+
+.DayPicker-WeekdaysRow {
+	display: tahle-row;
```
tmvanetten: Should this be `display: table-row;`

jsundquist: Yes it should have been. Thank you for catching that.

stennie:
Thanks @jsundquist! Good catch :)

PR #4696, opened 2018-07-06
Allow decimals in NumberArray

### LINKED ISSUE #4564: Can't type float number (1.23, 3.44 ) with NumberArray field
I use NumberArray Type, I want to save data like [1.33, 4.55, 4.2]， but On the admin UI, I cant type dot,  Then the data saved is [133, 455, 42].

ttsirkia:
Fixes #4564 by allowing a dot in the NumberArray

## Description of changes

Changed the faulty regular expression

## Related issues (if any)

#4564 

## Testing

### DIFF
--- a/fields/types/numberarray/NumberArrayField.js
+++ b/fields/types/numberarray/NumberArrayField.js
@@ -11,7 +11,7 @@ module.exports = Field.create({
 	mixins: [ArrayFieldMixin],
 
 	cleanInput (input) {
-		return input.replace(/[^\d\.]/g, '');
+		return input.replace(/[^-\d\.]/g, '');
 	},
 
 });
--- a/fields/types/numberarray/NumberArrayField.js
+++ b/fields/types/numberarray/NumberArrayField.js
@@ -11,7 +11,7 @@ module.exports = Field.create({
 	mixins: [ArrayFieldMixin],
 
 	cleanInput (input) {
-		return input.replace(/[^\d]/g, '');
+		return input.replace(/[^\d\.]/g, '');
 	},
 
 });


fields/types/numberarray/NumberArrayField.js
```diff
@@ -11,7 +11,7 @@ module.exports = Field.create({
 	mixins: [ArrayFieldMixin],
 
 	cleanInput (input) {
-		return input.replace(/[^\d]/g, '');
+		return input.replace(/[^\d\.]/g, '');
```
stennie: Thanks for taking a look at this issue! Unfortunately the fix isn't as simple as allowing `.` in the input, as that allows for invalid Number values like `3.4.5` or `..`.

I suggest comparing with the validation for the [`Number` field type](https://github.com/keystonejs/keystone/blob/master/fields/types/number/NumberField.js), which validates appropriately.

- Some allowable values to test: 12.34, -12.34, 0
- Some invalid values: A, 12.34.56

Once you have the code change working, please add or update the end-to-end (e2e) functional test coverage. As per the pull request template, please also list the specific browser versions that were tested.

For e2e testing Keystone 4 uses [Nightwatch.js](http://nightwatchjs.org/). See the [`test/e2e`](https://github.com/keystonejs/keystone/tree/master/test/e2e) directory in the Keystone repo for information. The e2e tests run automatically on every PR or commit, but you can run them much faster locally before pushing a PR.

Regards,
Stennie

ttsirkia:
In my opinion, `cleanData` is not used for validation and that is done elsewhere. That only tries to prevent some "stupid" input mistakes and therefore it doesn't have to be complete? However, minus sign is still missing.

stennie:
@ttsirkia Compare with the behaviour of the Number field, which doesn't allow invalid numbers to be input. Although there is validation elsewhere, the NumberArray field type should behave consistently instead of allowing inputs which are quietly corrected to Numbers when saved.

There is also some missing test coverage for NumberArray since the reported bug of decimals not being accepted did not cause any test failures. Adding/updating tests will confirm the code change works as expected (and continues to do so in future).

Thanks,
Stennie

ttsirkia:
It seems that the logic for `NumberField` does not work either because the `valueChanged` doesn't fire when the field is empty and any characters can be added. I'll close this PR and get back to this when I have a proper environment to run and enhance the tests. Probably the logic for the both has to be changed somehow.

The tests for `NumberField` should also have negative and decimal values and when such tests are added for `NumberArray` can they be used for the other, too.

ttsirkia:
Or actually, the `valueChanged` in `NumberField` works only properly if you start filling the field with numbers. If you start with some random letters, any input is accepted until there is a valid value and only after that the wrong input is blocked.

PR #4703, opened 2018-07-07
Replace / with . to delimiter the file extension

ocalvet:
## Description of changes
Updated the filename to read index.js in the Step 3 of the documentation

## Testing

Its a documentation change, no need for test cases


stennie:
Thanks @ocalvet ! However, please note that future PRs should be against the `master` branch rather than `docs` (which is 172 commits behind master). I'll cherry-pick and merge to master.

Regards,
Stennie

Issue #4694, opened 2018-07-03
Security hole in Keystone app?


intrnet:
Is it normal when you can post any file, requesting from any domain, without authorization? Posted files are stored in /tmp folder on server. I tested Keystone app without any changes, fresh installs. You can try on your Keystone app http://jsfiddle.net/N4Jxk/1089/, just change your domain in form action.

What can I do to secure my Keystone app??

ttsirkia:
Related to #4428

stennie:
@intrnet Thanks for reporting! As noted by @ttsirkia, this is related to the `multer` integration and issues #3597 and #4428.

I'm not familiar with the history, but we are going to address this in Keystone 4.0.0-rc.2. I'm going to close as a duplicate of #4428.

Regards,
Stennie

intrnet:
This is a good piece of news. So I can't wait new version, when it happens?

Issue #4428, opened 2017-09-02
Multer middleware stores all POSTed files to the local disk


zzxc:
While testing Keystone for a production website, we noticed that multer is used to parse uploaded files in POST data on all requests, saving the file(s) to /tmp. Depending on where /tmp is located on the system, this could result in a denial of service. This happens on all requests, even if file data isn't expected by the application.

Preferably, configuring the multer middleware (or, better yet, busboy) would be left to the application.

Steps to reproduce:
1. Use curl to upload a file from a remote system: curl -F 'data=@/path/to/local/file' https://keystone-site/any-path
2. On the server hosting the Keystone install, look for the uploaded file in /tmp: ls -alt /tmp | head

I tested with Keystone v0.3.22.

zzxc:
Looking at this more closely, I see that this is an old issue, #3597. We don't use any of the local file or image types, so my temporary solution is to simply comment out the initialization of multer.

What work is remaining to fix this properly? Would a stopgap solution (eg. only calling multer on pre-authenticated admin API requests) be acceptable?

Noviny:
@zzxc Our last look into multer yielded the results of #3175, which isn't ideal. Essentially, the rearchitecting necessary to resolve this is beyond our current scope. If you find a stopgap method that helps you and want to contribute it back, that would be great.

stennie:
Closed via PR #4704

Issue #3597, opened 2016-10-13
Update to new Multer version


sorryididntmeantto:
Keystone depends on multer and uses a really old version, vulnerable to a very simple DoS which brings the server down on a malformed multipart body. Also, even if updated, the way it is used is strongly discouraged and would allow a more complex DoS by filling up the disk with unhandled files.

The details about the first vulnerability are on expressjs/multer#224. The note about the second vulnerability is on that project's [README](https://github.com/expressjs/multer#any).

This vulnerability has already been exploited on one of my servers, probably accidentally by some badly written crawler, but bringing any Keystone instance down is rather simple until this gets fixed.
### Steps to reproduce the behavior

Send an incomplete or malformed multipart body as specified in expressjs/multer#224.
### Expected behavior

The server returns a 400 error because of bad input.
### Actual behavior

The server crashes.

VinayaSathyanarayana:
Would recommend that we also do a `nsp check` to check for other issues

JedWatson:
@sorryididntmeantto Not sure if your username and otherwise new account is a reference to just this issue... I would honestly have preferred a more responsible disclosure strategy but here we are. Hope you don't mind that I've updated the issue title to lower visibility until we can get a fix released; let me know if you think this should be dealt with differently.

@VinayaSathyanarayana agreed

Regarding the work to update multer, there were several blocking tasks that needed to be completed first, because (as you mention) the way it is used is not ideal, but Keystone has historically made a number of fundamental assumptions about handling uploaded files based on that way from before this vulnerability was discovered which means it's been very difficult to update from the old version (to put it simply, with the v0.3.x architecture we simply didn't have way to know which fields to expect in the middleware configuration, which is a problem more specific to generic solutions and less of a concern for applications implementing multer directly)

Anyway, the new architecture doesn't have these problems and in light of this issue I'm going to push through the remaining work to upgrade to the new version as fast as I can. If anybody wants to lend a hand (especially with testing) please let me know.

For what it's worth, @sorryididntmeantto I'm sorry you've had issues in production due to this - that's never fun, and hopefully it won't be an issue much longer.

stennie:
Closing as a duplicate of #4428, which will be addressed in Keystone `4.0.0-rc.2`.

Issue #3175, opened 2016-07-19
multer middleware is being used globally 


cameronbourke:
I am trying to use [multer](https://github.com/expressjs/multer) with Keystone. However, it is my understanding that multer configuration that I'm using is being overridden by Keystone's configuration.

I think the issue is within [bindBodyParser.js](https://github.com/keystonejs/keystone/blob/master/server/bindBodyParser.js#L13)

``` js
app.use(multer({
    includeEmptyFields: true,
}));
```

The use of multer **should not** be set globally for the express app. 
### Steps to reproduce the behavior

Within `routes/index.js`, I've set a route for posting a file (in this case a csv)

``` js
app.post('/upload', multer({ dest: './uploads'}).single('csv'), routes.views.upload)
```

Then within `routes/views/upload.js`.

``` js
module.exports = function(req, res) {
       var view = new keystone.View(req, res);

    view.on('post', (next) => {
        console.log(req.body); // {}
        console.log(req.file); // undefined
        console.log(req.files); /* { csv:
                  { fieldname: 'csv',
                    originalname: 'PJT001024.csv',
                    name: 'aa906fb2dcefdd1e4d45ed21f520998a.csv',
                    encoding: '7bit',
                    mimetype: 'text/csv',
                    path: '/var/folders/mn/6ttm31s566q4_xtyc7dfsns80000gp/T/d1ad8d73581f217b494ca74c84498c24.csv',
                    extension: 'csv',
                    size: 2652,
                    truncated: false,
                    buffer: null } 
                } */
        res.status(204).end();
    });

    view.render('upload');
}
```
### Expected behaviour

The object at the key `csv` on `req.files` should be appended to the request object with the key `file`, which is the expected behaviour as per the multer docs. So `req.file` should be an object, and `req.files` should be `undefined` instead.

The csv should then be written to the `./uploads` directory, which has been specified using the options object passed to `multer`. Therefore the `path` property on the `csv` object should point to `uploads/<some random id>.csv`.
### Actual behaviour

`req.file` is undefined, and `req.files` is an object. The csv is written to the temp directory on OSX `./var/folders/` instead of `./uploads`. My current workaround is to use [fs-extra](https://www.npmjs.com/package/fs-extra#movesrc-dest-options-callback) to read it from that temp directory and write it to `./uploads` instead.

wmertens:
Would this be solved by making the multer config configurable? Care to do a PR?

blargity:
[...]

PR #4704, opened 2018-07-08
Update to multer 1.3.1 and automatically clean uploaded files

JedWatson:
This has been frustrating me for a long time, but I've worked out a way to update `multer` to the latest version and retain backwards compatibility while still being safe with uploaded files.

This resolves several issue that have been raised about our use of an old multer version, including #4694 #4428 #3597 #3175 and #1563

It also introduces some new features:

- Keystone will automatically delete uploaded files at the end of the request
- You can disable the built-in multer in your app by setting the `handle uploads` Keystone option to `false` (it will still be included by the admin UI)
- You can pass options to multer my setting the `multer options` Keystone option

e.g

```
keystone.init({
  // ...
  'handle uploads': false, // turns it off
});

// or

keystone.init({
  // ...
  'multer options': { /* passed to multer */ },
});
```

**Technically** this is possibly a breaking change, because we now automatically clean up uploaded files. This was never a thing I'd expect to be used but any weird edge-case implementation that expects uploaded files to hang around after the request is completed will break. Because of this, I'd like to merge this for `4.0.0` final (not RC) and make a prominent note about it in the changelog.


intrnet:
I've tried on 4.0.0-rc.1 and 4.0.0-beta.8 version but I met same error, every requests ends without respond. And I guess there is little mistake in **_server/bindBodyParser.js_** `module.exports = function bundBodyParser (keystone, app) {` I think there should be `module.exports = function bindBodyParser (keystone, app) {` but this is not the couse of the problem.

JedWatson:
Thanks for catching that @intrnet 

Just to confirm, do you mean you tested merging this into your own branch based on `4.0.0-rc.1`?

We're looking into the failure to connect now. It all worked for me in local testing, but it looks like there is an issue with this in different environments.

intrnet:
I simply update multer and selected files, according to your changes ([Update to multer 1.3.1 and automatically clean uploaded files](https://github.com/keystonejs/keystone/pull/4704/commits/a3fcb46e791f4ed75c6cb06c82535ce4eecdce61)) in local testing based on linux ubuntu.

stennie:
@JedWatson I'm encountering the same issue with e2e tests (and keystone-demo) as reported in CI failures: `/keystone` times out waiting for response. Tried with Node 6.14.3 and 8.11.3. Also tried with both `yarn` and `npm` since your PR includes `yarn.lock` (I believe Travis is using `npm`).

I'm pulling from the `update-multer` branch associated with this PR, so not sure what is different in the test environments. Is there a specific `yarn`/`npm` incantation you would suggest to ensure a clean rebuild?

lib/uploads.js
```diff
@@ -0,0 +1,41 @@
+var fs = require('fs');
+var multer = require('multer');
+var os = require('os');
+
+function handleUploadedFiles (req, res, next) {
+	if (!req.files || !Array.isArray(req.files)) return;
```
ttsirkia: Should this be `return next();`?

ttsirkia:
I like the approach of cleaning the files when the request ends. However, if someone relies on the files still being on the disk, would it be useful to add a possibilty for a flag in the request or response object which skips the deletion for that request? Or should we assume here that if the file is wanted to store permanently, it must be copied elsewhere?

Issue #4541, opened 2018-01-05
GET parameters are not always strings


ttsirkia:
### Expected behavior

`req.query` contains only strings.

### Actual/Current behavior

There can be lists and objects.

### Steps to reproduce the actual/current behavior

Create a GET request that has the same parameter twice.

### Environment

https://github.com/keystonejs/keystone/blob/master/server/bindBodyParser.js#L11

Is there a reason why `extended` option is set to `true`? With this setting, a developer can never be sure that the parameter is actually a string. If a malicious user creates requests that are extended to rich objects, it is likely that the application will crash because of the `req.query.param` data type. At least I sometimes call methods such as `toLowerCase`, `toUpperCase`, `split` etc. which do not work with arrays or objects.

ttsirkia:
It seems that in Keystone 4 this is needed for the Admin UI but for that is is defined here: https://github.com/keystonejs/keystone/blob/master/admin/server/app/createDynamicRouter.js#L19 Could the `bindBodyParser` read this option from a configuration?

This is kind of a nasty way to crash the process just by adding the same GET parameter twice to the request if the code assumes to get a simple string instead of something else.

PR #4730, opened 2018-07-25
Add new extend query params option

ttsirkia:
## Description of changes

Adds a new optional config option `extend query params` which can be set to `false` to ensure that `req.query` will contain only string values.

Currently, GETting a page /mypage?a[r]=1 produces `{a: {r: 1}}` instead of a plain string which can crash the site if an object was not expected.

## Related issues (if any)

#4541 

## Testing

This does not affect the Admin UI as the default has not been changed and for the Admin UI this option is set elsewhere. There was not any suitable tests to extend to cover this feature.

### DIFF
--- a/docs/documentation/Configuration/Server-Options.md
+++ b/docs/documentation/Configuration/Server-Options.md
@@ -97,6 +97,10 @@ Optional config options that will be passed to the morgan middleware; see [githu
 
 Set this to enable processing of the HTTP request `X-Forwarded-For` header. Extracted IP addresses will be available as the array `req.ips` ([see docs here](http://expressjs.com/en/api.html)).
 
+<h4 data-primitive-type="Boolean"><code>extend query params</code></h4>
+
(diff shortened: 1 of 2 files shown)


ttsirkia:
Unfortunately, this doesn't completely solve the problem because if the same GET parameter appears twice, the result will be an array in `req.query`.

ttsirkia:
An option would be to add a simple middleware function if this option is `false`. It loops all the `req.query` keys and if there is an array, converts it to its first (or last) value. In this way, a developer can always be sure that the value is a string.

To recap the need, I treat the parameters as strings and can sometimes, for example, split the string without checking the type. If someone makes a malformed query, the site can crash. Checking the type of the value in all places where it is used requires quite a lot unnecessary code if a centralized solution is possible.

Any comments how to proceed?

ttsirkia:
I have to close this for now because it seems that Express is in some cases doing its own parsing and the setting here does not have any effect. I'll get back to this after further investigation.

PR #4799, opened 2018-09-12
Update part-3.md

ebrahimamer:
## Testing


ebrahimamer:
I think it's a bug

stennie:
@ebrahimamer Thanks for the documentation feedback.

The intent of the section of the tutorial you changed is to demonstrate adding a function to handle a route. The current example is correct, but incomplete for someone wanting to cut & paste. The actual function expected to be added is the following example (which does seem confusing).

The change you've suggested should instead be defined in a function `routeFunction`. Would you be able to update your PR to add this?

Regards,
Stennie

JedWatson:
Thanks @ebrahimamer 

I've updated the example as @stennie suggested to make it a bit clearer.

PR #4827, opened 2018-10-12
Fixed a horizontal scrollbar appearing on certain website pages

Vultraz:
The use of `100vw` was causing horizontal scrollbars to appear on the home and 404 pages.
Using 100% fixes that. Technically, block elements like <div> don't need a width key since
they automatically expand to take up as much width as possible, but I kept the value at 100%
here to matter match the original intent of the code.

## Testing


Vultraz:
@JedWatson pinging since this seems to have fallen through the cracks again 😬

Issue #1857, opened 2015-11-23
_req_user is undefined on pre/post remove hooks


poksme:
The following middlewares output an undefined `_req_user` :

```js
Article.schema.pre('remove', function reqUserDebug (next) {
	console.log('================ pre remove _req_user ================')
	console.log(this._req_user);
	return next();
});

Article.schema.post('remove', function reqUserDebug (next) {
	console.log('================ post remove _req_user ================')
	console.log(this._req_user);
	return next();
});
```
output:
```
================ pre remove hook _req_user ================
undefined
================ post remove hook _req_user ================
undefined
```

I do not know if it is useful but here is some background about my project regarding this issue:

I am developing a website with some _role limitation_ on the back-office. I am using the `_req_user` on pre save hook to limit modifications. But without the `_req_user` on pre remove hook it is impossible to limit the remove rights.

snowkeeper:
The admin ui adds `_req_user` with `updateHandler` (https://github.com/keystonejs/keystone/blob/master/lib/updateHandler.js#L174-L175).  

@JedWatson are you interested in supporting `_req_user` for remove also?  Would adding a remove method to **updateHandler** be ok?
```javascript
UpdateHandler.prototype.remove = function(callback) {
	
	if ('function' !== typeof callback) {
		callback = function() {};
	}
	
	// Make current user available to pre/post remove events
	this.item._req_user = this.user;
		
	this.item.remove(callback);
	
};
```

Our deletes can be something like
```javascript
req.list.model.findById(req.query['delete']).exec(function (err, item) { 
	if (err || !item) return res.redirect('/keystone/' + req.list.path);
	
	item.getUpdateHandler(req).remove(function (err) {
		if (err) {
			console.log('Error deleting ' + req.list.singular);
			console.log(err);
			req.flash('error', 'Error deleting the ' + req.list.singular + ': ' + err.message);
		} else {
			req.flash('success', req.list.singular + ' deleted successfully.');
		}
		return res.redirect('/keystone/' + req.list.path);
	});
});

```

erg0dic:
One workaround I have been using is to turn on tracking for the target model & use the updatedBy field to determine the user who made the changes.

poksme:
I could not get the `updatedBy` field to work properly but I found a solution, I tried to keep it as clean as possible without modifying any keystone source file.

> There may be some extra code because my project uses _FlowType_ annotations.

### 1 - You will need to install a couple of npm packages
```shell
[...]

Issue #3187, opened 2016-07-21
How to create a plugin where I need request object in plugin?


macroramesh6:
I need to write a mongoose `plugin`.  where I need a Request object to find `hostname`. 

```
module.exports = exports = function(schema, options) {
    schema.add({
        hostname: String
    })

    schema.pre('save', true, function(next, done) {
        this.hostname = req.headers.host; //need request object
        next();
        setTimeout(done, 100);
    });
}
```

wmertens:
This is not currently possible. The only thing that is passed to the schema plugins is the user: https://github.com/keystonejs/keystone/blob/0e5a4a11d13d4ced058b392b119140f42116dc7e/lib/updateHandler.js#L177

To have the hostname, it would similarly need to be passed. However, that is extra work for every call, without many things using it.

I wonder if there could instead be some session context object that could be used instead? This would be a breaking API change, but it would allow pluggable additions, and the object would be created once per session and just referenced on each save.

so instead of 
```js
function UpdateHandler (list, item, req, res, options) {
        …
	this.user = req.user;
        …
}
…
	var saveItem = function () {

		// Make current user available to pre/post save events
		this.item._req_user = this.user;
```
we'd have
```js
function UpdateHandler (list, item, req, res, options) {
        …
	this.context = req.session.context;
        …
}
…
	var saveItem = function () {

		// Make current user etc available to pre/post save events
		this.item.__context = this.context;
```
You could then add the hostname to the context object in one of the other hooks.

@JedWatson thoughts?

webteckie:
@wmertens t would be nice if we could do this for any mongoose hook and not just pre/post save.  And not, that this keeps pushing [business] logic into the mongoose where we should instead be attempting to do this up in keystone.  There has been some discussion about this in slack.  That been said, I like your proposal for pre/post save hooks :=)

wmertens:
@webteckie Well, we can just pass this session context object to all the mongoose hooks, no? Would also simplify authorization, eg. just check the userlevel in the context object vs the current data…

webteckie:
@wmertens probably.   For now I would issue a PR for your idea, if @JedWatson is alright with it.  You don't have to break anything if you leave the _req_user alone.

webteckie:
btw, I would call the item property `_ _ session_context`

wmertens:
No time for a PR right now, sorry, so anybody that feels up to it?

macroramesh6:
[...]

PR #4831, opened 2018-10-12
Add _req_user to removed item so you can use this in the remove pre/p…

laurenskling:
…ost hook

As discussed in https://github.com/keystonejs/keystone/issues/1857 the remove pre/post hooks are "missing" the _req_user. I, amongst others, am using this object to handle some permissions. It's available in the updateHandler (in `create` and `update`) but missing in `remove`.

it's added on the updateHandler here: https://github.com/keystonejs/keystone/blob/d34f45662eb359e2cb18b397f2ffea21f9883141/lib/list/updateItem.js#L157

Remove doesn't use the updateHandler, so this is missing.

## Testing

### DIFF
--- a/admin/server/api/list/delete.js
+++ b/admin/server/api/list/delete.js
@@ -38,6 +38,7 @@ module.exports = function (req, res) {
 			return res.apiError('database error', err);
 		}
 		async.forEachLimit(results, 10, function (item, next) {
+			item._req_user = req.user;
 			item.remove(function (err) {
 				if (err) return next(err);
 				deletedCount++;


autoboxer:
this seems like a worthwhile addition, and would be consistent with the other hooks.  My concern is that a larger change seemed pending in #3187, but discussion died.  I'd like to think this change is independent of #3187, and to prevent a breaking change, both options should live side by side, however, I'd feel more comfortable if @dominikwilkowski or @JedWatson weighed in before continuing.

laurenskling:
Would be great if we could get this PR sorted. I am sure many like me are awaiting this value to settle permissions on deleting records. A simple line of code, endless possibilities :D

JedWatson:
Thanks @laurenskling.

Sorry this went unmerged for so long, will push a new release with the fix this weekend.

laurenskling:
Awesome @JedWatson thanks so much

PR #4885, opened 2019-01-15
Removing deprecated package "letsencrypt-express" and replacing it with "greenlock-express"

akopcz2:
Removing deprecated package "letsencrypt-express" and replacing it with "greenlock-express". this removes the vulnerable dependency trail of letsencrypt-express requiring 
"le-acme-core" trickling down to "request" requiring "hawk" which requires the vulnerable package "cryptile" which exposes the vulnerability CVE-2018-1000620

Upgrading to the new module "greenlock-express" removes this entire dependency trail

## Description of changes
**Package.json**
Removed deprecated letsencrypt-express dependancy and replace it greenlock-express.

**Inside /server/initLetsEncrypt.js**
Replaced  require statement to require "greenlock-express" instead of deprecated letsencrypt-express


JedWatson:
Thanks @akopcz2 

I resolved the conflict here and updated to the current version. Also actually removed the `letsencrypt-express` package from the dependencies because it looks like that wasn't part of the PR.

Issue #4851, opened 2018-11-15
yo keystone fails


bakovkin:
### Expected behavior
The "yo keystone" process should have gone without errors.

### Actual/Current behavior
**First error:**

make: вход в каталог «/home/and/project/test/my-site/node_modules/kerberos/build»
  CXX(target) Release/obj.target/kerberos/lib/kerberos.o
In file included from ../../nan/nan.h:190:0,
                 from ../lib/kerberos.h:9,
                 from ../lib/kerberos.cc:1:
../../nan/nan_maybe_43_inl.h: In function ‘Nan::Maybe<bool> Nan::ForceSet(v8::Local<v8::Object>, v8::Local<v8::Value>, v8::Local<v8::Value>, v8::PropertyAttribute)’:
../../nan/nan_maybe_43_inl.h:88:15: error: ‘class v8::Object’ has no member named ‘ForceSet’
   return obj->ForceSet(GetCurrentContext(), key, value, attribs);

**Next errors:**
In file included from /home/and/.node-gyp/10.13.0/include/node/node.h:63:0,
                 from ../lib/kerberos.h:4,
                 from ../lib/kerberos.cc:1:
/home/and/.node-gyp/10.13.0/include/node/v8.h:10046:15: note: declared here
 Local<Object> Value::ToObject() const {
               ^
kerberos.target.mk:105: ошибка выполнения рецепта для цели «Release/obj.target/kerberos/lib/kerberos.o»

make: *** [Release/obj.target/kerberos/lib/kerberos.o] Ошибка 1
make: выход из каталога «/home/and/project/test/my-site/node_modules/kerberos/build»
gyp ERR! build error 
gyp ERR! stack Error: "make" failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/home/and/.nvm/versions/node/v10.13.0/lib/node_modules/npm/node_modules/node-gyp/lib/build.js:262:23)
gyp ERR! stack     at ChildProcess.emit (events.js:182:13)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:240:12)
gyp ERR! System Linux 4.10.0-38-generic
gyp ERR! command "/home/and/.nvm/versions/node/v10.13.0/bin/node" "/home/and/.nvm/versions/node/v10.13.0/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
[...]

Issue #1564, opened 2015-07-31
Installing Keystone's npm dependencies fails.


yn5:
The problem occurs in OSX node v0.12(.2)  as well as in a node:0.12 Docker container.

Since today installing the npm dependencies of a Keystone project generated with Yeoman generator (or causing the npm install by creating a new project with the Yeoman generator) fails on installing  the nodejieba dependency.

This is the npm output of the freshly generated Keystone project by Yeoman:

```
npm WARN peerDependencies The peer dependency browserify@>= 2.3.0 < 12 included from browserify-shim will no
npm WARN peerDependencies longer be automatically installed to fulfill the peerDependency
npm WARN peerDependencies in npm 3+. Your application will need to depend on it explicitly.

> fsevents@0.3.7 install /Users/yn5/stuff/web/phkstone-temp/node_modules/keystone/node_modules/watchify/node_modules/chokidar/node_modules/fsevents
> node-gyp rebuild

  SOLINK_MODULE(target) Release/.node
  SOLINK_MODULE(target) Release/.node: Finished
  CXX(target) Release/obj.target/fse/fsevents.o
  SOLINK_MODULE(target) Release/fse.node
  SOLINK_MODULE(target) Release/fse.node: Finished

> kerberos@0.0.11 install /Users/yn5/stuff/web/phkstone-temp/node_modules/keystone/node_modules/mongoose/node_modules/mongodb/node_modules/kerberos
> (node-gyp rebuild 2> builderror.log) || (exit 0)

  CXX(target) Release/obj.target/kerberos/lib/kerberos.o
  CXX(target) Release/obj.target/kerberos/lib/worker.o
  CC(target) Release/obj.target/kerberos/lib/kerberosgss.o
  CC(target) Release/obj.target/kerberos/lib/base64.o
  CXX(target) Release/obj.target/kerberos/lib/kerberos_context.o
  SOLINK_MODULE(target) Release/kerberos.node
  SOLINK_MODULE(target) Release/kerberos.node: Finished

> bson@0.2.22 install /Users/yn5/stuff/web/phkstone-temp/node_modules/keystone/node_modules/mongoose/node_modules/mongodb/node_modules/bson
> (node-gyp rebuild 2> builderror.log) || (exit 0)

  CXX(target) Release/obj.target/bson/ext/bson.o
  SOLINK_MODULE(target) Release/bson.node
  SOLINK_MODULE(target) Release/bson.node: Finished
[...]

PR #4888, opened 2019-01-22
Remove kerberos from dependencies

larrybolt:
## Description of changes
Removal of kerberos since it's [not used anywhere in the project](https://github.com/keystonejs/keystone/search?q=kerberos) anyway and only creates problems when installing packages since kerberos requires compiling.

The only thing that happened is that the package version was bumped over time. It also makes it more difficult to deploy keystone in a docker container (based of alpine, because of the compile requirement).

## Related issues (if any)
#4851, #1564, 

## Testing
I didn't test since for some reason even the unit tests gave me issues without having modified anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
     "glamor": "^2.20.25",
     "grappling-hook": "^3.0.0",
     "i": "^0.3.5",
-    "kerberos": "^1.0.0",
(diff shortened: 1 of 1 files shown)


JedWatson:
Thanks @larrybolt

PR #4921, opened 2019-05-18
add support for language cookie options

Hsn723:
## Description of changes
https://github.com/keystonejs/keystone/blob/master/lib/middleware/language.js#L8
The language middleware accepts a `language cookie options` configuration, but that value is not being used when creating the cookie:
https://github.com/keystonejs/keystone/blob/master/lib/middleware/language.js#L14-L17

This PR fixes this by consuming the given option.

## Testing

 - [x] List browser version(s) any admin UI changes were tested in: Chrome 74
 - [x] Please confirm you've added (or verified) test coverage for this change.


JedWatson:
Thanks @Hsn723!

PR #4937, opened 2019-06-26
fix value for noEdit InputForm in UrlField

laurenskling:
PR https://github.com/keystonejs/keystone/pull/4856/ resulted in a bug when UrlField was set to `noedit`

this fixes it again.

### DIFF
--- a/fields/types/url/UrlField.js
+++ b/fields/types/url/UrlField.js
@@ -35,11 +35,9 @@ module.exports = Field.create({
 		const { value } = this.props;
 		return (
 			<div>
-				<FormInput
-					noedit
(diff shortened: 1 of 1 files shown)


laurenskling:
`value` needs to be passed down as `children`.

laurenskling:
The original PR doesn't add any documentation by the way, so the `thumb` flag is not documented right now for UrlField

Tathanen:
+1 for this PR, I just noticed this issue when pulling in the latest Keystone changes in my fork.  Let's get a merge!

JedWatson:
👍 Thanks @laurenskling

